Issue #46, opened 2018-04-23
Add per monitor workspace style


hedning:
# Multi monitor support

There's two main ways to incorporate workspaces and monitors:
1. The usual DE way, where a workspace spans all monitors
2. The usual tiling wm way where a workspace spans one monitor

The reason DEs mostly use the first style is the lack of any workspace layout. It's simply not very feasible to move the content of one monitor to another as the sizes might differ completely.

This can cleary be seen by how Gnome Shell handles adding new monitors. A new blank monitor gets added while windows remain on existing monitors. The new monitor being assigned primary status doesn't change the behavior.

A big problem with this approach is when it's time to remove a monitor and you're forced to move windows automatically, creating a big mess.

Mutter doesn't actually enforce style 1., it simply exposes what's essentially the X workspace standard, where a workspace is a list of windows.

With the new architecture we've become mostly independent of how Gnome Shell handles workspaces and monitors.

## Implementation

- Windows on a workspace would correspond exactly with the windows in a `Space`.

- The `Space` would be attached to any monitor, and the windows would simply be moved to that monitor.

- Every monitor would have a selected `Space` with a corresponding workspace.

- When a window enters a monitor we look up the selected `Space` and insert it.

We can only do this cleanly by using the `span workspaces across all monitors`.

Issue #17, opened 2017-09-21
Multimonitor support


olejorgenb:

hedning:
Commit 51e95bf998e061a83d3a00baca1eab18a77d3957 fixes workspaces for the primary display. Having a monitor on the left or right of the primary doesn't really work practically as you get the stack protruding onto the monitor, this can potentially be fixed by setting some clip values.

Also, I'm not having much luck in setting `/org/gnome/shell/overrides/workspaces-only-on-primary` to false.

PR #49, opened 2018-04-27
Multimonitor support

### LINKED ISSUE #47: Fix coordinates
We now deal with several coordinate systems:

- Screen coordinates, MetaWindows and WindowActors live here
- Monitor relative `Space` coordinates, the window clones live here.

This require some care about what system we're dealing with.

hedning:
## Features

- When changing monitors the active workspace will follow the primary monitor, the remaining monitors will be populated with the top of the workspace mru.

- The workspace mru stack is shared between all monitors, scaled to fit the monitor accessing the stack.

- Non active monitors are covered in a overlay which will activate the corresponding workspace when detecting the mouse pointer. It seems like there's no `pointer-entered-monitor`.

- When changing to another visible workspace the pointer will warp to the center of its monitor.

## Some implementation notes

We create an actor spanning the whole screen, `Spaces.spaceContainer` which holds all spaces. A `Space` is now first contained in a `Space.clip` container clipping away anything that protrudes from the monitor. The `cloneContainer` is then moved within the `clip` when doing previews.

There's now an associated `monitor` per `Space` (`space.monitor`), which can be set by `Space.setMonitor(monitor, animate)`. When doing previews the `space.clip` actor is moved and scaled without setting the associated monitor.

We also keep visible workspaces through a  `monitor` -> `Space` mapping. Currently implemented with  a `Map` accessed through `spaces.monitors`. So used like this `spaces.monitors.get(monitor)`. It might be better to just keep a reference on the monitor objects instead though.

## Some remaining issues:

- [x] The top bar shows the name of the active workspace, not the workspace that's on the primary monitor.
- [ ] The minimap is always placed on the primary monitor.
- [x] When moving windows to another monitor the pointer will warp back to the original monitor for some reason. Disabled moving the focus for now.
- [x] Move and resize the top bar to the focused monitor.
- [ ] The activities overview won't show windows on other monitors. Will delay this.
- [ ] Windows not in the tiling or the scratch layer gets hidden when focusing another monitor.

Fixes #47, #46, #48, #17

### DIFF
--- a/app.js
+++ b/app.js
@@ -2,8 +2,8 @@
   Application functionality, like global new window actions etc.
  */
 
-const Shell = imports.gi.Shell;
-const Tracker = Shell.WindowTracker.get_default();
+var Shell = imports.gi.Shell;
+var Tracker = Shell.WindowTracker.get_default();
 
 function newWindow(metaWindow) {
     metaWindow = metaWindow || global.display.focus_window;
--- a/extension.js
+++ b/extension.js
@@ -1,27 +1,27 @@
-const Extension = imports.misc.extensionUtils.getCurrentExtension();
-const convenience = Extension.imports.convenience;
+var Extension = imports.misc.extensionUtils.getCurrentExtension();
+var convenience = Extension.imports.convenience;
 
-const modules = [
+var modules = [
     Extension.imports.tiling, Extension.imports.scratch,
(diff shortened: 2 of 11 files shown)


hedning:
There's some remaining issues with eg. transient windows etc. but should be a major improvement against what we had before so merging this, and then I'll fix things later.

PR #60, opened 2018-06-25
Hide window previews after workspace was selected

xanf:
I'm using paperwm without coloring each workspace (actually I'm using other extension which changes wallpaper on workspace change). In that case I can see workspace previews in background (especially if target workspace is empty).

This change hides previews when workspace change was done

### DIFF
--- a/navigator.js
+++ b/navigator.js
@@ -438,6 +438,7 @@ function switchWorkspace(to, from, callback) {
     let xDest = 0, yDest = global.screen_height;
 
     let toSpace = Tiling.spaces.spaceOf(to);
+    toSpace.actor.show();
     let selected = toSpace.selectedWindow;
     if (selected)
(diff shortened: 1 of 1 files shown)


hedning:
Thanks, that seems sane.

You do need something like this too though:
```diff
@@ -553,6 +553,7 @@ function switchWorkspace(to, from, callback) {
     let xDest = 0, yDest = global.screen_height;
 
     let toSpace = Tiling.spaces.spaceOf(to);
+    toSpace.actor.show();
     let selected = toSpace.selectedWindow;
     if (selected)
         Tiling.ensureViewport(selected, toSpace, true);
```
to support activating workspaces outside of the navigator, eg. using the dock.

hedning:
Would be nice to have them squashed :)

xanf:
Sure, give me a minute :)

PR #87, opened 2018-11-04
added default background to preferences UI

ganthern:
Hey,
something easy to get my feet wet. The title says all I think.

Cheers and thanks for the awesome extension!


hedning:
Thanks, this looks good :)

PR #89, opened 2018-11-17
Improve Slurp behaviour with freshly created windows

ganthern:
Slurp now pushes the active window into the column to the left
if there is no column to the right.
This allows freshly created windows to be stacked without
having to change focus.

### DIFF
--- a/tiling.js
+++ b/tiling.js
@@ -925,7 +925,7 @@ class Spaces extends Map {
 
         this.forEach(space => space.init());
 
-        // Redo the stack 
+        // Redo the stack
         this.monitorsChanged(); // Necessary for X11
         this.stack = this.mru();
     }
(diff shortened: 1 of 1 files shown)


hedning:
That does make a lot of sense in general :+1: 

Some notes:

- Focus should probably not change, though that's mainly due to `removeWindow` and `addWindow`  doing stuff that's not wanted in this case. We should really have a lower level method which only modifies the columns. 
- When the last column is tiled vertically the behavior is kinda strange, as the focused window is added to the left column. It might make more sense if the behavior only works on single window columns? Or alternatively slurping from the left column.

ganthern:
I agree about the focus change, slurping should not affect which window is focused.

The odd behaviour you mention in your second point makes me think that maybe it would be useful to have an order modifying function that would only ever affect the focused window?
Along the lines of moving the window through the columns to the left and to the right. With numbers being unfocused windows and X being the focused window:
```
|1|2|4|X|      |1|2|4|      |1|2|X|4|      |1|2|4|      |1|X|2|4|
| | | | |      | | | |      | | | | |      | | | |      | | | | |
| | | | |  to  | | | |  to  | | | | |  to  | |3| |  to  | | | | |
| |3| | |      | |3|X|      | |3| | |      | | | |      | | |3| |
| | | | |      | | | |      | | | | |      | |X| |      | | | | |
| | | | |      | | | |      | | | | |      | | | |      | | | | | and vice versa
```

hedning:
> The odd behaviour you mention in your second point makes me think that maybe it would be useful to have an order modifying function that would only ever affect the focused window?

That could be nice, though I think arrow key navigation should behave the way it currently does. Being able to rearrange with the mouse would also be nice.

For this PR I'd say either only slurp single window columns, or do what amounts to a reverse slurp:

```
  | 1 | 2 | 4 | X |    | 1 | 2 | 4 |    | 1 | 2 | 3 | 
  |   |   |   |   |    |   |   |   |    |   |   |   | 
  |   |   |   |   | to |   |   |   | to |   |   | 4 | 
  |   | 3 |   |   |    |   | 3 | X |    |   |   |   | 
  |   |   |   |   |    |   |   |   |    |   |   | X | 
  |   |   |   |   |    |   |   |   |    |   |   |   | 
```

Barfing would then be a bit weird as it wouldn't restore the previous order, so I think it might make the most sense to restrict reverse slurp to single window columns tbh.

ganthern:
I _think_ the weird focus behaviour might be fixed now, but I'm actually not quite sure why this seems to work. The last-column-slurp now only works on single-window columns. Not a big loss, since new windows always should be alone, and new windows are what this PR is supposed to improve.

hedning:
Right, the focus behavior was pretty puzzling, as `ensureViewport` shouldn't switch focus at all. And looking again, neither `removeWindow` nor `addWindow` triggers `Main.activateWindow`.

Looks like the behavior we saw had the correct focus, but the wrong highlighting due to `ensureViewport` on the wrong window. Your solution correctly picks the window that's actually focused. A simpler solution would be to use `display.focus_window`, but that would unfortunately break navigation using the minimap (where focus is only changed after `Super` is let go). The problem arose due to `removeWindow` selecting a new `selectedWindow` if the removed window was selected, which is necessary for sane behavior when closing non-focused windows from while using the minimap.

If you could squash the commits and remove trailing whitespace I'll merge :+1:

hedning:
Thank you :)

PR #101, opened 2019-03-25
Default workspace switching and moving to wrap around

goodwillcoding:


hedning:
Thanks :) That works pretty well (nice seeing the animation code handle simple changes like this without any problems :laughing:). It also has the added benefit of making `super-shift-pageup` go to what's most likely an empty workspace :+1:

hedning:
Lets give it a go, I'm not thrilled about how it works when holding eg. pagedown down, but all in all should be worth it.

PR #107, opened 2019-04-28
Per workspace top bar hiding

hedning:
Integrating with the hide topbar extension gets rather hacky (we have to check the position of the panel to see if it's been moved away, which only works due to the fact that the extension happens to move the panel).

Instead add a «zen-mode» which can be toggled per-workspace to hide the top-bar. Currently turned on by default for testing, but should default to showing the topbar when merged. We should also provide a global option which sets the preferred default.

Working:

- [x] easily toggle the top bar from the workspace menu
- [x] always show the top bar when using the workspace carousel 

Needs: 

- [x] global default preference
- [ ] update the preference gui
- [ ] show the top bar when activating the hot corner (nice to have at least)

The top bar is currently turned off by default for testing, but we'll default to showing the top bar.


hedning:
Haven't updated the preference gui yet.

To hide the top-bar by default use this:
```bash
dconf write /org/gnome/shell/extensions/paperwm/default-show-top-bar false
```

PR #124, opened 2019-06-22
Fix typo in keybindings example

jvns:
I tried using this example and got an error because `i` should be `k`. With this change it works for me.

Thanks for the great tool! I'm really enjoying it so far.


olejorgenb:
Thanks :)

Issue #125, opened 2019-06-23
Weird behavior when inserting (not appending) new workspaces using the overview


olejorgenb:
Drag and drop a window from the overview between two existing workspaces to insert a new workspace.

Mutter only support appending workspaces so gnome-shell will append a new workspace and move windows around to emulate a insertion.

jtaala:
Fixed in #577.

PR #126, opened 2019-06-23
Handle adding/removing workspaces on the fly

hedning:
- handle dynamic-workspaces when used with multiple monitors
- make sure every monitor has a space
- actually hook up spaces correctly when they're created on the fly


hedning:
There's some outstanding issues like #125, and workspace preferences aren't handled perfectly when using dynamic workspaces.

Issue #129, opened 2019-06-24
Configurable cycle-width and cycle-height steps


olejorgenb:
Ref: https://paperwm.zulipchat.com/#narrow/stream/174096-general/topic/width/near/168872494

- Per monitor (or per monitor width/height)?
- Support both ratios and fixed pixel count? (value <= 1.0 interpreted as ratio, fixed pixel count otherwise)

if someone want to change the ratios manually before an option is implemented this is the places: [width](https://github.com/paperwm/PaperWM/blob/4322009d9fd88154e3c34b99b9b6c29a578faf8e/tiling.js#L2410), [height](https://github.com/paperwm/PaperWM/blob/4322009d9fd88154e3c34b99b9b6c29a578faf8e/tiling.js#L2443) (restart the shell for changes to take effect)

PR #130, opened 2019-06-24
Configurable cycle-width/height steps

olejorgenb:
Remaining: support fixed pixel count values also.

See #129


olejorgenb:
Pixel count implemented. Not supporting mixed ratio and pixel count in same steps array for now (no validation)

PR #131, opened 2019-06-24
Sync space settings on workspace changes

olejorgenb:
If the indices isn't synced a disable/enable cycle will jumple the workspace
names/backgrounds

### DIFF
--- a/tiling.js
+++ b/tiling.js
@@ -1366,11 +1366,24 @@ class Spaces extends Map {
             }
         }
 
+        let nextUnusedWorkspaceIndex = nWorkspaces;
         for (let [workspace, space] of this) {
             if (workspaces[space.workspace] !== true) {
                 debug('workspace removed', space.workspace);
                 this.removeSpace(space);
+
+                // Maps in javascript (and thus Spaces) remember insertion order
+                // so the workspaces are sorted by index. The relative ordering
+                // of the removed workspaces will thus be preserved when resurrected.
+                space.settings.set_int('index', nextUnusedWorkspaceIndex);
+                nextUnusedWorkspaceIndex++;
             }
+        }
+
+        // Ensure the live spaces have correct indices
+        for (let [workspace, space] of this) {
+            space.settings.set_int('index', workspace.index());
+            Meta.prefs_change_workspace_name(workspace.index(), space.name);
         }
     };
 


settings.js
```diff
@@ -109,6 +109,15 @@ function getWorkspaceSettingsByUUID(uuid) {
         path: `/org/gnome/shell/extensions/paperwm/workspaces/${uuid}/`});
 }
 
+function disconnectWorkspaceSettings(uuid) {
+    let list = workspaceList.get_strv('list');
+    let i = list.indexOf(uuid);
+    if (i > -1) {
+        list.splice(i, 1);
+        workspaceList.set_strv('list', list)
```
hedning: Won't this result in losing all the workspace settings? Saved preferences are looked up through the list of uuids, so if they're no longer there we simply can't find them again.

olejorgenb: Yes, so it should either be combined with code that picks from existing "disconnected" settings when new workspaces are created. (PS: I see that the disconnectWorkspaceSettings is called for all workspaces - that's not intended ofc.)

Not sure if it's hard to enumerate all workspace settings (in the dconf database). Maybe it's simpler to set the index to -1 and without removing it from the list.

tiling.js
```diff
@@ -1367,6 +1367,15 @@ class Spaces extends Map {
                 this.removeSpace(space);
             }
         }
+
+        // Ensure the space settings use correct indices
+        for (let [workspace, space] of this) {
+            space.settings.set_int('index', workspace.index());
+            // Trim to correct size too?
+            Meta.prefs_change_workspace_name(workspace.index(), space.name);
+            Settings.disconnectWorkspaceSettings(space.settings.get_str('uuid'));
```
olejorgenb: The disconnectWorkspaceSettings call should be in the "removed" loop above

settings.js
hedning: Hmm, took the list of uuids pattern from gnome-terminal, I assumed it was basically required, though dconf-editor obviously manages to look them up, so it would be a bit weird if it was impossible when I think about it.

tiling.js
hedning: Might make sense to set the index to the first free index instead. So if there's there's three workspaces and we remove the second, we simply juggle the indexes like this:  [1, 2, 3] -> [1, 3, 2]. That way the preference isn't lost, but will be reused when a new workspace is added. To make the UX good we'd need some method to flip through previously used preferences though ofc., but at least we won't destroy stuff if someone tries out dynamic workspaces for a bit.

Issue #135, opened 2019-06-25
Trouble to switch window between monitors


egegunes:
I couldn't figure out how to switch windows or workspaces between monitors. Workspace and window change bindings for Gnome does not work. Can it be accomplished by a custom keybinding?

hedning:
The built in `super-shift-arrows` should work to move windows between monitors (though note that mouse warping isn't supported in wayland making it a bit awkward). If the monitors are scaled differently it might not work however as the resulting resize might push the window back to the original monitor (we simply listen on the event that tells us when windows enters a monitor). Dragging windows with the mouse also works.

There's currently no simple action to move/rotate workspaces between monitors, we should really get that done. It is however possible by pushing the workspace down to the stack (`super-above_tab`) and retrieving it on the other monitor (also with `super-above_tab`).

hedning:
We added a keybinding example which cycles through monitors. Add this to the `init` function in `user.js`:
```
   Extension.imports.examples.keybindings.cycleMonitor();
```
That should make `super+d` cycle through monitors.

egegunes:
I tried the keybinding but it selects the other monitor only momentarily.

hedning:
Right, I'm guessing the mouse pointer is stuck on the other monitor? The pointer will then easily trigger a motion event on the old monitor and bring focus back. On x11 we're able to warp the pointer to the correct monitor, but gnome on wayland doesn't support that yet unfortunately.

So, yeah, x11 is by far better when using multiple monitors a lot. Now we should probably add some lower threshold to the pointer motion making it less likely to trigger it without intent.

hedning:
Made a PR which should fix the example keybind on wayland (it will also try to reject random pointer movement).

egegunes:
I checked out motion-workspace-threshold branch and tried. Super+D works now!

hedning:
Cool, I'll push to master pretty soon.

egegunes:
master works fine but switching focus to a fullscreen window with Super+D causes #134

hedning:
Should be fixed #182, though there's still some funky top bar behavior. Really need to rewrite most of that logic I think :/

PR #139, opened 2019-07-01
Require a threshold motion to trigger workspace switch

hedning:
On multi monitor setups we usually want the monitor focus to follow the mouse pointer since non-active workspaces don't react to the pointer at all.

Unfortunately it's not possible to warp the pointer from gnome-shell extensions in wayland yet. The pointer will be left on an inactive monitor when switching workspaces using the keyboard. This makes it extremely easy to trigger an motion event and inadvertently switching monitor focus (in fact this simply happens the moment we activate the motion listener).

So add a motion threshold to filter out pointer motions without user intent.

ref #https://github.com/paperwm/PaperWM/issues/135

### DIFF
--- a/stackoverlay.js
+++ b/stackoverlay.js
@@ -76,9 +76,20 @@ class ClickOverlay {
 
         this.signals = new utils.Signals();
 
+        this._lastPointer = [];
         this.signals.connect(
             enterMonitor, 'motion-event',
-            () => {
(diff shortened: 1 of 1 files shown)


hedning:
closed in 8e667c0c119034551fc0c4c586fbeaa7170cc5a5

PR #157, opened 2019-07-14
Unmaximize window on cycleWidth

Pajn:
I find it very annoying when trying to do cycleWidth and nothing is happening. This is usually caused by the window being maximized for some reason. Simply cheeking if it is, and if so unmaximizing it, is a simple fix that works well.

### DIFF
--- a/tiling.js
+++ b/tiling.js
@@ -2447,6 +2447,10 @@ function cycleWindowWidth(metaWindow) {
         targetX = workArea.x + workArea.width - minimumMargin() - targetWidth;
     }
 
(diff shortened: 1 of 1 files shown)


olejorgenb:
Thanks.

We should probably do something similar for maximize-width.

Ideally we'd capture maximize completely as maximized windows look a bit funny with non-zero window gaps (the selection overlaps the panel)

Either by somehow stopping the maximze action and do maximize-width instead or maybe hide the panel (making it a step between fullscreen an maximzie-width)

PR #158, opened 2019-07-14
Remove unneded topbar workspace label

Spissable:
Hi,
when using a theme that messes with the style/positioning of the topbar, the label of the spaces won't perfectly position. For example with the Sweet-Dark shell theme:
label-on
Instead of trying to re-position it so it overlaps nicely, I just completely hid it, as I don't see any use for the label:
label-off

Cheers

### DIFF
--- a/topbar.js
+++ b/topbar.js
@@ -428,7 +428,7 @@ function enable () {
 
         imports.mainloop.timeout_add(0, () => {
             for (let [workspace, space] of Tiling.spaces) {
-                space.label.set_position(Math.round(r.x), Math.round(r.y));
+                space.label.set_position(Main.panel.actor.x + Math.round(r.x), Main.panel.actor.y + Math.round(r.y));
                 let fontDescription = label.clutter_text.font_description;
                 space.label.clutter_text.set_font_description(fontDescription);
             }})


olejorgenb:
The label is useful when changing workspaces so we can't simply remove it. We could show/hide the label when the workspace-switcher is active/inactive, but that is extra bookkeeping we'd ideally avoid since we still need the label and button to align pretty well - otherwise the animation would look a bit strange.

But it might be unrealistic to make the match perfect, so maybe we should reconsider and hide the label when workspace-switching is done. 

It's strange that the offset is that large though - the theme seem to move the panel slightly down and to the right which the relative transform should pick up. Seems like we don't get the allocation signal..

PS: There's an hidden setting for not touching the activities button at all that also hides the workspace lables: `dconf write /org/gnome/shell/extensions/paperwm/use-workspace-name false`

olejorgenb:
Seems we assume the panel is positioned at 0,0. Can you try this patch?

```patch
diff --git a/topbar.js b/topbar.js
index 4d022a562..ed16a4c4b 100644
--- a/topbar.js
+++ b/topbar.js
@@ -428,7 +428,7 @@ function enable () {
 
         imports.mainloop.timeout_add(0, () => {
             for (let [workspace, space] of Tiling.spaces) {
-                space.label.set_position(Math.round(r.x), Math.round(r.y));
+                space.label.set_position(Main.panel.actor.x + Math.round(r.x), Main.panel.actor.y + Math.round(r.y));
                 let fontDescription = label.clutter_text.font_description;
                 space.label.clutter_text.set_font_description(fontDescription);
             }})
```

Spissable:
The patch works a charm :). I updated the PR accordingly

topbar.js
```diff
@@ -428,7 +428,7 @@ function enable () {
 
         imports.mainloop.timeout_add(0, () => {
             for (let [workspace, space] of Tiling.spaces) {
-                space.label.set_position(Math.round(r.x), Math.round(r.y));
+                space.label.set_position(Main.panel.actor.x + Math.round(r.x), Main.panel.actor.y + Math.round(r.y));
```
hedning: Hmm, this will probably go wrong when the panel is on a monitor where `monitor.x != 0`.

olejorgenb: yeah, I realized that after merging, but it seems the actor is inside the panelBox so it shouldn't be a problem.

olejorgenb: That also means it's still not 100% robust ofc. But maybe we're lucky and themes can't change the panelbox.

PR #167, opened 2019-08-30
Fix crash under Gnome 3.34

Pajn:
`extenstionUtils.extensions` is undefined in 3.34 and thus crashes the extension.
`extensionUtils.getCurrentExtension();` was previously used in some places so I assume that it also exists in older versions or otherwise the current version of the extension should have already crashed there.

Window placement is messed up as well so this does not give full 3.34 compatibility but not crashing is a start.

### DIFF
--- a/examples/keybindings.js
+++ b/examples/keybindings.js
@@ -1,4 +1,4 @@
-var Extension = imports.misc.extensionUtils.extensions['paperwm@hedning:matrix.org'];
+var Extension = imports.misc.extensionUtils.getCurrentExtension();
 var Keybindings = Extension.imports.keybindings;
 var Main = imports.ui.main;
 var Tiling = Extension.imports.tiling;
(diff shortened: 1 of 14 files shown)


hedning:
Cheers.

The reason we've used the current access pattern for `Extension` is enabling hot reloading whole modules through `gnome-shell-mode`. So we need to move initialization to the module's `init` function (like we do in `user.js`) in order not to break that.

If a module doesn't have a init function adding one should just work out of the box.

hedning:
Right, might be easier to use something like this, as they've simply changed the way to look up extensions (found it when fixing gnome-shell-mode):
```
var Extension;
if (imports.misc.extensionUtils.extensions) {
    Extension = imports.misc.extensionUtils.extensions["paperwm@hedning:matrix.org"];
} else {
    Extension = imports.ui.main.extensionManager.lookup("paperwm@hedning:matrix.org");
}
```

While a bit verbose moving everything to `init` did cause some problems.

PR #179, opened 2019-10-04
Modernize the workspace menu

hedning:
- Remove the `Set ...` buttons, just update the name on the fly instead
- Remove the color selection, didn't really work all that great. Perhaps rather add 3 sliders ala. the volume?
- Add buttons to select other workspace settings 

image

### DIFF
--- a/topbar.js
+++ b/topbar.js
@@ -180,6 +180,8 @@ class WorkspaceMenu extends PanelMenu.Button {
                              'switch-workspace',
                              this.workspaceSwitched.bind(this));
 
+        this.menu.addMenuItem(new PopupMenu.PopupSeparatorMenuItem(_('Workspace Settings')));
+
         this.entry = new PopupMenuEntry(this._label.text);
         this.menu.addMenuItem(this.entry);
         let changed = () => {
--- a/topbar.js
+++ b/topbar.js
@@ -249,10 +249,12 @@ class WorkspaceMenu extends PanelMenu.Button {
         this.nextIcon.connect('clicked', () => {
             let space = Tiling.cycleWorkspaceSettings(-1);
             this.entry.label.text = space.name;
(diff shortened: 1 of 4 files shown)


topbar.js
```diff
@@ -188,10 +245,72 @@ class WorkspaceMenu extends PanelMenu.Button {
...
+            unbound = strip.slice(1);
+
+            nextSettings[1].set_int('index', wsI);
+            space.setSettings(nextSettings); // ASSUMPTION: ok that two settings have same index here
+
+            // Re-assign unbound indices:
+            for (let i = n; i < N; i++) {
+                unbound[i-n][1].set_int('index', i);
+            }
+            updateWorkspaceIndicator();
```
olejorgenb: Is `updateWorkspaceIndicator()` needed? `space.setSetting` should update the indicator.

topbar.js
```diff
@@ -188,10 +245,72 @@ class WorkspaceMenu extends PanelMenu.Button {
...
 
-        this.menu.box.add(this.prefsIcon, { expand: true, x_fill: false });
+        function rotated(list, dir=1) {
+            return [].concat(
+                list.slice(dir),
+                list.slice(0, dir)
+            );
+        }
+
+        let cycle = (dir=1) => {
```
olejorgenb: Move to a util to facilitate usage from keybindings?

hedning: Yes, probably settings.js, not entirely sure where it fits in the best.

topbar.js
hedning: Nope, doesn't look like it is.

topbar.js
hedning: Move to tiling.

hedning:
Dropped the color picker for now, but left the code in, can see if we can integrate back in later.

PR #186, opened 2019-10-16
Better workspaces-only-on-primary support

hedning:
This adds better support for workspaces-only-on-primary:
- moving windows with grab
- moving between monitors (without or without a window)
- Only works properly when secondary monitors are above or below primary :-1: (Though this isn't a new bug) (I'm seeing some weirdness using a nested gnome-shell to test, which I'm unable to reproduce in a real session).

We now only handle grabs in `window-entered-monitor`, which should hopefully solve windows accidentally moving between workspaces when changing monitors :tada: 

~I opted to remove setting  `workspaces-only-on-primary` to false as a recommended setting, as things should honestly work a bit smoother with workspace only on primary.~

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ You can then enable the extension in Gnome Tweaks, or enable if from the command
 gnome-shell-extension-tool -e paperwm@hedning:matrix.org
 ```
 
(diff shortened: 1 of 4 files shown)


hedning:
OK, pretty certain this makes workspaces-only-on-primary simply better, while not causing any regressions. And it should fix accidentally moving windows when changing monitors :)

There's some visual wonkyness if the primary monitor is small. That can make new windows spawn on the secondary, causing a buggy frame and new window animation, it's not a new issue though as far as I can tell.

PR #188, opened 2019-10-17
Undefined workspace settings handling

olejorgenb:


settings.js
```diff
@@ -133,6 +149,9 @@ function getNewWorkspaceSettings(index) {
     list.push(uuid);
     workspaceList.set_strv('list', list);
     settings.set_int('index', index);
+    settings.set_string('color', getUnusedColor());
+    let n = findWorkspaceSettingsByName(/^Unnamed .*/).length;
+    settings.set_string(`Unnamed ${n+1}`)
```
hedning: Is there anything which prevents us from using `Workspace ${i}` here?

olejorgenb: No, as long as the list is "monotonic" (which it is atm.) that would also work

hedning: Right, I'd vote for `Workspace n` in that case, as that makes it clearer that it's a workspace related.

jtaala:
Closing this as #584 addresses this issue - as well as work with dynamic workspaces (and injecting workspaces between other workspaces - e.g. non-monotonic workspace naming).

PR #193, opened 2019-10-26
wip: Smart cycle width position

hedning:
Every keybind action gets it's own context, which lives as long as the action is repeated. Use this infrastructure to provide eg. better cycleWindowWidth.


hedning:
Doesn't intergrate with floating windows yet.

jtaala:
Closing as this PR describes the current behaviour with (unless I'm misunderstanding it) - e.g. cycle widths and other actions remain in context and can be repeated until modifier is released etc.

PR #202, opened 2019-11-22
gnome-shell-extension-tool -> gnome-extensions

tycho01:
addresses warning `gnome-shell-extension-tool is deprecated, use gnome-extensions instead`.


hedning:
Good catch, though perhaps we should wait until 3.36 as `gnome-extensions` are only supported in 3.34.

hedning:
Added a better installer in https://github.com/paperwm/PaperWM/pull/234, which take cares of enabling paperwm (and does it consistently), so  `gnome-shell-extensions-tool` is no longer mentioned.

PR #207, opened 2019-12-05
Always keep a minimum number of workspaces

hedning:
Dynamic workspaces are really just superior, but it can be really useful having your favorite workspaces in predictable order. Respect the exesting `num-workspaces` when using dynamic workspaces.


hedning:
merged to develop in 27f8d957c76fb81f8b089b83b2c0892e2968f424

Issue #113, opened 2019-05-13
Is it planned to make paperWM more mouse friendly ?


inkhey:
Hi, PaperWM look like an interesting idea :smiley: . I was searching for a way to have some king of intelligent tiling (with resizing) on gnome-shell and i end up here (more than current half tiling, as i want to have more than 2 windows on one screen). That's was not really what i expected but that's really interesting.

What interesting me is solution about using both tiling (for nice windows placement) and mouse (because finding good keybinding between applications is complex and i'm currently using 2 different keyboard layout).

PaperWM is really nice, i like it but as far as i go, i didn't see a way to move easily windows order, i see mostly keyboard shortcut but not some much mouse interaction. Is there also a way to allow to switch windows using mouse wheel (by placing the mouse a some place or with a key)  ?

hedning:
Yeah, reordering windows using the mouse is on the agenda, but we haven't done anything about yet.

> allow to switch windows using mouse wheel

We support 3-finger swiping with a touchpads and ideally we would support `super+mouse-wheel` to scroll through the tiling, but it's unfortunately not supported by gnome-shell: https://gitlab.gnome.org/GNOME/mutter/issues/207

Having some sort of gui widget that listens to scrolling is possible (eg. you can scroll through workspaces on the workspace name button in the top bar), though not sure where it would fit nicely.

olejorgenb:
PS: I'm sure you know, but all keyboard shortcuts are configurable using either the extension settings (GUI) or dconf-editor[1]. I know it's a pain to reassign a ton of shortcuts though :)

[1] so it is possible to create a script that switches between sets of bindings.

olejorgenb:
Quick prototype of "scroll the topbar to switch windows": 

Scrolling while the mouse pointer is over the topbar selects the next/prev window and brings up a minimap (enters a navigation mode).

While the minimap is open:

- Clicking the topbar close the minimap and focus the selected window
- Pressing <kbd>Return</kbd> close the minimap and focus the selected window
- Pressing <kbd>Esc</kbd> close the minimap and go back to the focused window
- Clicking a non-selected window in the minimap scrolls to that window
- Clicking the selected window in the minimap close the minimap and focus the window (or dobbelclicking a unselected window)
- Cicking a unfocused window (ie. in the actual tiling) focus that window and close the minimap
- Scrolling over the minimap selects the next/previous minimap

[...]

PR #225, opened 2020-01-12
drag and drop windows

olejorgenb:
Ref: #113


hedning:
Short snippet to show how it works:
dnd

PR #255, opened 2020-02-25
add resize functions to inc/dec height/width

xkonni:
As discussed on zulip, keybinds now default to <Super>=/- and <Super><Shift>=/-.


jtagcat:
Discussion source?
Also be vary of keyboard layouts, on mine, I would have to press Ctrl+Shift+0 and Ctrl+- - rather map the keys, not the characters.

xkonni:
> Discussion source?
> Also be vary of keyboard layouts, on mine, I would have to press Ctrl+Shift+0 and Ctrl+- - rather map the keys, not the characters.

There's a topic on zulip, general.
And yes, those keys may be different in different layouts. For me = and - are primary keys and right next to one another (us layout). However, the values used here are just defaults and can be easily changed in the preferences. Personally I don't use them and mapped them to vim-like bindings.

Which keys would you suggest instead?

schemas/org.gnome.shell.extensions.org-scrollwm.gschema.xml
```diff
@@ -172,6 +172,26 @@
...
+      <summary>Increment window height</summary>
+    </key>
+
+    <key type="as" name="resize-h-dec">
+      <default><![CDATA[['<Super><Shift>minus']]]></default>
+      <summary>Decrement window height</summary>
+    </key>
+
+    <key type="as" name="resize-w-inc">
+      <default><![CDATA[['<Super>equal']]]></default>
```
hedning: We want to have `plus` too here, as there's layouts with unshifted plus.

xkonni: okay. but then it's the same keybind as resize-h-inc for e.g. US layouts with plus being <Shift>equal. or am i missing something?

hedning: Ah, right, that's a problem.

xkonni: I mean we could go for ctrl instead of shift and it would be okay. Maybe not nice to reach, but without conflicts and then everyone can change their bindings.

hedning: Did a quick test, seems to work correctly. Mutter doesn't care about shifting/case eg. `<super>r` and `<super>R` is the same keybinding.

hedning: Actually, it looks like `<super>plus` also works in us layout. Ie. `<super>plus` and `<super>equal` is the same binding in us layout.

xkonni: ah nice. okay, then we'll go for that.

hedning:
Squashed the commits to develop, thanks :)

xkonni:
awesome, thank you ;)

PR #259, opened 2020-03-08
Drag and drop windows while navigating

hedning:
Mutter's window grab blocks all other input, but it's possible to deactivate a grab from js.

So to make navigation while dragging work we implement the whole process using clone's and clutter events. The native `grab-op-begin` signal is only used to initiate a grab (we can't listen to arbitrary button events on WindowActors) which we immediately cancel in favor of the before mentioned implementation.

In addition to navigation while grabbing, we now support `<super><ctrl>button-press` as a shortcut to go straight into DnD mode. This can be very handy when you want to move a floating window into the tiling.

### DIFF
--- a/grab.js
+++ b/grab.js
@@ -211,6 +211,7 @@ var MoveGrab = class MoveGrab {
             return;
         this.dnd = true;
         global.display.end_grab_op(global.get_current_time());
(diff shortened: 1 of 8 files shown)


smichel17:
All of these commits are already in `develop`. They must have been rebase-merged, so the hashes don't match, but if you search by commit message, you can confirm they are all there.

PR #264, opened 2020-03-11
gnome shell background -> GNOME Shell background

yegortimoshenko:
Changes "GNOME Shell" name capitalization to match upstream project name: https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/master/README.md


hedning:
Cheers :)

PR #272, opened 2020-03-17
Suggest a way for X11 users to determine the WM_CLASS.

johnmee:


README.md
```diff
@@ -198,7 +198,7 @@ It's possible to create simple rules for placing new windows. Currently most use
     });
 ```
 
-The `wm_class` of a window can be looked up by clicking <kbd>Super</kbd><kbd>Insert</kbd> and then checking the value of `metaWindow.wm_class` in emacs or looking glass.
+The `wm_class` of a window can be looked up by clicking <kbd>Super</kbd><kbd>Insert</kbd> and then checking the value of `metaWindow.wm_class` in emacs or looking glass, or typing `xprop | grep WM_CLASS` at the command line then clicking on the window.
```
hedning: It's probably good to mention that this only works on X11 windows.

It might make sense to direct users to the windows tab in looking glass, `Alt-F2 lg`, which gives you an overview of all windows and their app-id and wm-class since that works in wayland too.

johnmee: Thanks.  Qualified it for X11.  I'm not familiar with Looking Glass (or wayland) so will not try to provide instructions for those.

smichel17 (approved):
Looks good to me. At some point we may wish to split some of the documentation out of the README, but Today Is Not That Day!

smichel17:
Actually, it looks like this was already addressed in https://github.com/paperwm/PaperWM/commit/3ad3100413a4247e133b21de23a8005a72af4c54.

PR #279, opened 2020-03-27
uninstall.sh: Actually remove $EXT

jtagcat:
uninstall.sh currently exits with `rm: cannot remove '/home/jc/.local/share/gnome-shell/extensions/paperwm@hedning:matrix.org': Is a directory` Adding -R will allow removing the directory and depending on distribution, -f disallows asking for 'if you want to delete this file in the directory' for each file deleted.

### DIFF
--- a/uninstall.sh
+++ b/uninstall.sh
@@ -19,4 +19,4 @@ if [[ "$LINK" != "$REPO" ]]; then
     echo "$EXT" does not link to "$REPO", refusing to remove
     exit 1
 fi
-rm $EXT
+rm -Rf $EXT


hedning:
Right, `uninstall.sh` assumes you've installed paperwm using `install.sh`, ie. using a symlink to the repo. And because we install from a git repo with potentially local changes we don't want to kill the repo. Eg. it wouldn't be nice if `make uninstall` killed the source.

`gnome-extensions uninstall` does a recursive remove and I managed to kill my git repo once with it, that's the main reasoning behind not using `-R`. Now we could add `-f`.

jtagcat:
How would you even install it standalone?
I downloaded the install.sh, executed it and got many 404 errors.

jtagcat:
remind me to come back to this, to learn / test how install.sh works (where does it git clone to, it does not use git and so on)?
git clone not working: need to ssh-add my key, but apparently gnome session crashes, when I try to use the compose key (what the ssh key password needs)
remind me to find out where to report the compose key, ssh, gnome-terminal or gnome-shell bug, f32 beta love

hedning:
@reacocard found a satisfying solution for this in #312  so closing.

Issue #308, opened 2020-05-21
Top bar disappears when moving between monitors


elshimone:
There appears to be an extension conflict between PaperWM develop branch (404d50c41675f0169018ce8ecaa13cb589825389) and Dash to panel v37.

I have "Make top bar follow montior focus" off in PaperWM.  

In dash to panel, I have "display the main panel" on set to default, and "Display panels on all monitors" set to off.  I have played around with the options a bit to see if I can get a combination which works, but to no effect.

You can see when I move my cursor from the primary to the secondary monitor on the left, the top bar disappears.

GNOME Shell 3.36.1
simon@tachikoma ~/d/PaperWM>                                            (PaperWM) 20:30:53

They are both great extensions, FWIW I have disabled dash to dock for now :)

![paperwm_dash_to_dock_trimmed](https://user-images.githubusercontent.com/698207/82598406-89600500-9ba2-11ea-87e1-02d2daa12e1d.gif)

hedning:
Right, «top bar follows focus monitor» still moves the top bar, but always to the primary monitor: https://github.com/paperwm/PaperWM/blob/b33f2d859dc0eaced1d338e336928fa78943e32d/topbar.js#L624-L633

I'd guess this issue would be fixed by leaving the top bar alone in the above code, and only move it to the primary monitor when `topbar-follow-focus` is turned off:
https://github.com/paperwm/PaperWM/blob/b33f2d859dc0eaced1d338e336928fa78943e32d/topbar.js#L540-L549

elshimone:
Thanks for the pointer, this fixes it for me, but I don't completely understand what the topbar_follow_focus setting should do.

jtaala:
Fixed by #309.

PR #309, opened 2020-05-22
308 Topbar lost when moving between monitors

elshimone:
See #308  Seems to be a conflict with dash to dock.  Fix is to only move topbar
when topbar_follow_focus is toggled off.

### DIFF
--- a/topbar.js
+++ b/topbar.js
@@ -540,7 +540,7 @@ function enable () {
     signals.connect(Settings.settings, 'changed::topbar-follow-focus', (settings, key) => {
         let monitors = Tiling.spaces.monitors;
         if (!settings.prefs.topbar_follow_focus) {
-            moveTopBarTo(Main.layoutManager.prim);
+            moveTopBarTo(Main.layoutManager.primaryMonitor);
         }
         let to = setMonitor(Main.layoutManager.focusMonitor);
         let space = monitors.get(to);
--- a/topbar.js
+++ b/topbar.js
@@ -540,15 +540,13 @@ function enable () {
     signals.connect(Settings.settings, 'changed::topbar-follow-focus', (settings, key) => {
         let monitors = Tiling.spaces.monitors;
         if (!settings.prefs.topbar_follow_focus) {
-            let to = moveTopBarToPrimary(Main.layoutManager.focusMonitor, moveTopBar);
+            let to = moveTopBarToPrimary(Main.layoutManager.focusMonitor);
             let space = monitors.get(to);
             updateWorkspaceIndicator(space.workspace.index());
-            for (let [workspace, space] of Tiling.spaces) {
-                space.layout();
-            }
-
         }
-
+        for (let [workspace, space] of Tiling.spaces) {
+            space.layout();
+        }
     });
 
     signals.connect(panelBox, 'show', () => {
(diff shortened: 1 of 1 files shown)


topbar.js
```diff
@@ -622,8 +625,15 @@ function setWorkspaceName (name) {
 }
 
 function setMonitor(monitor) {
-    if (!prefs.topbar_follow_focus)
```
hedning, line 625: When `topbar_follow_focus` is true we need to move the top bar to `monitor`. 

 `panelMonitor` also needs to be set as there's other code which depend on knowing which monitor the top bar is on (not necessary each time, but it doesn't hurt).

topbar.js
```diff
@@ -539,11 +539,14 @@ function enable () {
 
     signals.connect(Settings.settings, 'changed::topbar-follow-focus', (settings, key) => {
         let monitors = Tiling.spaces.monitors;
-        let to = setMonitor(Main.layoutManager.focusMonitor);
-        let space = monitors.get(to);
-        updateWorkspaceIndicator(space.workspace.index());
-        for (let [workspace, space] of Tiling.spaces) {
-            space.layout();
+        if (!settings.prefs.topbar_follow_focus) {
```
hedning, line 542: We want to do everything here when `topbar_follow_focus` is true too. (eg. layouts should be run to fix up window heights).

So we only want to move the top bar to the primary monitor in this if block.

I guess it makes sense to generalize `moveTopBarToPrimary()` into `moveTopBarTo(monitor)`, that way `setMonitor` cas use it too when top bar should follow focus.

elshimone:
Ok I understand what the topbar_follow_focus setting does, and it appears to work with dash to dock disabled (i.e. the top bar moves to the focussed monitor).  But if I turn on topbar_follow_focus and dash to dock then I get the original issue :(  At least I can turn off the setting for my own use, but I'd like to fix this properly.

hedning:
Yeah, better support for dash-to-panel would be nice, but I'm not really sure if it's reasonable to fix it when top bar follow focus unfortunately, though I might be wrong here.

As I understand it dash-panel have a lot of options, one of which puts a panel on every monitor. In that case paperwm shouldn't be moving the top bar around at all, but without some ad-hoc dash-to-panel fix we can't necessarily know what to do.

I guess we could turn `topbar-follow-focus` off if we detect dash-to-panel, that's at least a minimal hack. 

(There's also an issue with how we handle the workarea, which makes having the panel on the bottom broken, I think the underlying issue might be fixed in newer gnome shells though making it possible to rely entirely on the reported workarea).

topbar.js
```diff
@@ -622,8 +623,14 @@ function setWorkspaceName (name) {
 }
 
 function setMonitor(monitor) {
-    if (!prefs.topbar_follow_focus)
-        monitor = Main.layoutManager.primaryMonitor;
```
hedning, line 626: This needs to happen to. `setMonitor` is called when changing workspaces, and without it `panelMonitor` will be incorrect.

topbar.js
```diff
@@ -539,13 +539,14 @@ function enable () {
 
     signals.connect(Settings.settings, 'changed::topbar-follow-focus', (settings, key) => {
         let monitors = Tiling.spaces.monitors;
-        let to = setMonitor(Main.layoutManager.focusMonitor);
-        let space = monitors.get(to);
-        updateWorkspaceIndicator(space.workspace.index());
```
hedning, line 544: We still want these lines too. `setMonitor` updates `panelMonitor` which the tiling code uses to know which monitor the panel is located. `updateWorkspaceIndicator` refreshes the workspace name in the top bar.

topbar.js
```diff
@@ -539,13 +539,15 @@ function enable () {
 
     signals.connect(Settings.settings, 'changed::topbar-follow-focus', (settings, key) => {
         let monitors = Tiling.spaces.monitors;
+        if (!settings.prefs.topbar_follow_focus) {
+            moveTopBarTo(Main.layoutManager.prim);
```
hedning, line 543:
```suggestion
            moveTopBarTo(Main.layoutManager.primaryMonitor);
```

hedning:
Cheers :) I'll try to test it out quickly and then merge.

hedning:
Right, finally got a chance to take a look at this.

Seems like it mostly works as intended, but due to the way we handle monitor workarea there's some wonkyness left to be sorted out (the window is forced down on the second monitor). We should really just trust what the workarea method, I'll need to look closer why we made things as they are (my recollection is that fullscreen in older gnome shell versions caused the workarea to change which could cause trouble).

Though, what I mentioned above isn't a regression, so I think it's fine to merge this as is, and fix the remaining problems later.

hedning:
Cheers :)

Sorry it took so long to merge.

PR #312, opened 2020-05-31
fix uninstall.sh

reacocard:
fixes the errors
"rm: cannot remove '/home/reacocard/.local/share/gnome-shell/extensions/paperwm@hedning:matrix.org': Is a directory"
and
"rm: remove write-protected regular file '/home/reacocard/.local/share/gnome-shell/extensions/paperwm@hedning:matrix.org/.git/objects/pack/pack-3c3158f6621fdd04f2b4c3870a93934d2f6b695f.pack'?"
by adding -r and -f respectively to rm.

### DIFF
--- a/uninstall.sh
+++ b/uninstall.sh
@@ -19,4 +19,12 @@ if [[ "$LINK" != "$REPO" ]]; then
     echo "$EXT" does not link to "$REPO", refusing to remove
     exit 1
 fi
-rm $EXT
+if [ -L "$EXT" ]; then
+    rm "$EXT"
+else
+    read -p "Remove $EXT? (y/N): " -n 1 -r
+    echo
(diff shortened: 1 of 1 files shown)


hedning:
See https://github.com/paperwm/PaperWM/pull/279#issuecomment-604957573 for the reasoning why I'm hesitant to do it this way.

reacocard:
Ah, that makes sense. I could make the logic a bit smarter, i.e. test whether it's a symlink and if it is, just rm the link, otherwise rm -rf the whole directory. Does that sound reasonable?

hedning:
Yeah, that's reasonable. I'd be in favor of adding a confirmation before removing the directory though, so what's happening is clear.

reacocard:
I've updated it with the test logic and confirmation, please take another look.

hedning:
Cheers :)

PR #346, opened 2020-10-28
Add option to disable showing scratch windows in overview.

### LINKED ISSUE #340: Hide scratch windows in overview
Is it possible to hide scratch windows in the GNOME activities overview? It feels weird to have the windows follow me through every workspace, and they take up a lot of space in the overview. I don't see the point of them being there, when you can just bring the entire scratch layer to the front at any time to see what windows are in it.

Thank you so much for this extension, its really amazing and I've gotten such a great setup now on GNOME.

Pacman99:
Closes #340 

This reworks the scratch in overview option as a drop down menu that lets the user choose between never, always, or only showing scratch windows in overview. 

To track the 'never' option, I added a boolean setting 'disable-scratch-in-overview'. The existing 'only-scratch-in-overview' is used to track the 'only' option. If both are set to true, which can only be done manually, the default is only showing scratch windows.

I'm using this on gnome 3.36, so I'm not sure what changes are necessary for 3.38. But I think the GUI and settings logic could be useful either way.


hedning:
Nice, the menu selection makes a lot of sense.

Will hopefully work in 3.38 (been a slow poke upgrading this cycle, so don't have reliable access at the moment).

PR #353, opened 2020-11-29
fix workspace-label border

christopher-l:
Removes small bottom border on workspace label that is visible when not covered by the top bar (especially on multi-monitor setups).

Before:
image

After:
image


hedning:
Cheers, haven't seen that artifact, but I don't use multi monitor setups much, and commit makes sense.

PR #358, opened 2020-12-20
Add "StackUnstack" action in Examples.

mrspikeroot:
If the focused window is not stacked, it stacks it at the bottom of an adjacent column
If the focused window is stacked, it moves it out into its own column.


mrspikeroot:
This demonstrates a different way to manage pulling windows in and out of stacked columns.

hedning:
Cheers 😊

PR #360, opened 2020-12-23
Fix/worspace switching glitch

christopher-l:
I experience a bug on a two-monitor setup when directly jumping to workspaces (either by using a keybinding or when clicking on notifications) that lie on the monitor that is not currently active:
image
Moving the cursor in this state to the second monitor causes further glitching of workspaces an windows.

The bug can be reproduced quite reliably by doing the following:
- Have open windows on workspaces 1 to 4
- Restart Gnome Shell (<kbd>Alt</kbd> + <kbd>F2</kbd>, <kbd>r</kbd>, <kbd>Enter</kbd>)
  > workspace 1: visible on monitor 1
  > workspace 2: visible on monitor 2
  > workspace 3: on monitor 1
  > workspace 4: on monitor 1
- Go to monitor 2 and switch to workspace 4 using <kbd>Super</kbd>+<kbd>Above_Tab</kbd>
  > workspace 1: visible on monitor 1
  > workspace 2: on monitor 2
  > workspace 3: on monitor 1
  > workspace 4: visible on monitor 2
- Now switch workspaces directly using keyboard shortcuts assigned to "Switch to workspace x" in Settings
  - switch to workspace 1
  - switch to workspace 2
  - switch to workspace 3

I don't completely understand the involved code, but I could eliminate the bug with the first commit in this pull request.

-----

The second commit changes the behavior when switching workspaces on the active monitor. Before, switching, e.g., from workspace 1 to workspace 5 would trigger an animation that scrolls quickly past all workspaces in-between on the same monitor, adding a lot of visual clutter. This commit makes it use the same more subtle animation that appears when switching to a workspace on another monitor. There appear to be some situations in which this animation is not performed correctly (happened before when switching to another monitor as well), but I cannot reproduce this reliably.

### DIFF
--- a/tiling.js
+++ b/tiling.js
@@ -1588,6 +1588,8 @@ class Spaces extends Map {
             // Only start an animation if we're moving between workspaces on the
             // same monitor
             this._initWorkspaceSequence();
+        } else {
+            this.selectedSpace.setMonitor(this.selectedSpace.monitor, false);
         }
 
         this.stack = this.stack.filter(s => s !== toSpace);
@@ -1995,6 +1997,8 @@ class Spaces extends Map {
             return;
         }
 
+        this._updateMonitor();
+
         Tweener.addTween(to.actor,
                          { x: 0,
                            y: 0,
@@ -2019,6 +2023,14 @@ class Spaces extends Map {
             above = above.get_next_sibling();
         }
 
+    }
+
+    _updateMonitor() {
+        let monitorSpaces = this._getOrderedSpaces(this.selectedSpace.monitor);
+        let currentMonitor = this.selectedSpace.monitor;
+        monitorSpaces.forEach((space, i) => {
+            space.setMonitor(currentMonitor, false);
+        });
     }
 
     addSpace(workspace) {


hedning:
Cheers for tracking down and fixing  :)

The multi monitor code is indeed a bit obtuse and messy unfortunately (and neither me nor @olejorgenb really use a multi monitor setup much). 

I'll try to take some time to test your fix and go through the code again soon, so I'll have a better handle on what's actually wrong.

The sequence vs. most recently used animation commit should probably be split into another commit though :) While it's a bit swooshy when going to through several workspaces my impression is that it's a preferred default. 

Thanks again for contributing, will try to merge soon :)

christopher-l:
> The sequence vs. most recently used animation commit should probably be split into another commit though :) While it's a bit swooshy when going to through several workspaces my impression is that it's a preferred default.

I see. Should I remove  https://github.com/paperwm/PaperWM/pull/360/commits/59c5652dcd473a22b2d906585c5658e4f5fd6466 from the pull request?

hedning:
Yeah, I'm open to be convinced changing the animation, but it doesn't really fit this pr.

hedning:
OK, I'll just merge and try to figure out the underlying cause later :)

christopher-l:
Thanks ^^
You will probably find some things to improve, but this made working with the PaperWM a whole lot easier for me for the moment.

Issue #318, opened 2020-06-29
Small offset on workspace name


Natchuz:
This small visual bug appears all workspaces except on the first you see after launching Gnome. It's also visible when cycling, that mentioned workspaces have their names a little bit moved. However I was not able to take example screenshot, since it's impossible capturing when holding super key. This bug happens currently on Fedora 32, Gnome 3.36.3. I also noticed this bug, before I moved to Fedora, on Ubuntu 19.10 (I don't remember which Gnome version it was).
After quick extension re-enable, bug dissapears.

![Screenshot from 2020-06-29 19-25-37](https://user-images.githubusercontent.com/30700722/86036561-609f1b00-ba3e-11ea-946c-3d68c3b085c0.png)

hedning:
Right, I've noticed this when launching nested gnome-shell, on secondary monitors.

Are you seeing this with only a single monitor and are you seeing it on wayland or x11?

Natchuz:
Normally I use one monitor, so I plugged second cable and it still happens on both outputs, no matter if it's x11 or wayland.

christopher-l:
For me, this happens when new workspaces are added dynamically ("Dynamic Workspaces" activated).

jtaala:
Hey all, this shouldn't happen on recent supported versions of PaperWM (e.g. on `develop`, `release`, or on PaperWM installed via https://extensions.gnome.org/extension/6099/paperwm/).

Let me know if anyone is still seeing this, otherwise will close this issue in a few weeks or so.

christopher-l:
Looks good to me now.

jtaala:
> Looks good to me now.

Thanks @christopher-l.

PR #361, opened 2020-12-23
Add options to isolate monitors and hide empty spaces

christopher-l:
This adds two options to the "Toggles" section of the extension settings:
image
The options change the behavior when switching workspaces by stack using e.g. <kbd>Super</kbd>+<kbd>Above_Tab</kbd>.

### Isolate Monitors
I use a multi-monitor setup and tend to leave workspaces assigned to a given monitor. <kbd>Super</kbd>+<kbd>Above_Tab</kbd> is the quickest way to switch recent workspaces, but often I end up switching to a workspace that was assigned to another monitor by mistake, switching to the wrong workspace *and* assigning the workspace the wrong monitor. Showing only workspaces that belong to the active monitor avoids this kind of mistake. 

### Hide Empty Spaces
At the moment, I use 10 static workspaces. Showing all of them in the recent-workspaces switcher adds a lot of clutter and makes it hard to find workspaces, that actually have windows on them. Additionally, selecting an empty workspace from the recent-workspaces switcher is almost never what I want to do. When I want an empty workspace, I can still get to one by using the ordered-workspace switcher, by selecting it directly using a shortcut, or assign a shortcut especially for going to an empty workspace.

## Considerations
This pull request solves my problems, but one could consider a few changes:
- Make some of these options default instead of having toggles
- Find a better category in the settings widget / make the effect of the settings clearer
- Add an option to show only a single empty workspace in the switcher
- Apply settings also to the ordered-workspace switcher

I don't have a good idea which of the proposed changes might be useful, but I'm happy to adapt the pull request if needed.


hedning:
Yep, isolating workspaces for super+above_tab is  probably worth an option :+1: And might be a good default when we support moving workspaces between monitors better in the future.

I think a better default solution for the empty workspace is  __always__ sorting empty spaces below populated ones, this is a simple change that doesn't suffer hiding things from the user. In general we do try to limit the amount of options as it adds a fair bit of complexity in testing and maintenance. So I strongly prefer behavior that works well as default and gets out of your way.

Hope that makes :)

christopher-l:
> And might be a good default when we support moving workspaces between monitors better in the future.

For me, a [custom keybinding](https://github.com/christopher-l/dotfiles/blob/5f610b782105335e15aca9b20481a3e360b4ff2f/config/paperwm/user.js#L108-L136) works quite well for thits. I found that there are situations, where I like the current behavior with workspaces from all monitors being shown, though.

> I think a better default solution for the empty workspace is always sorting empty spaces below populated ones

That would definitely help. For now, I would still prefer to only show a single empty workspace, but this wouldn't be a problem with dynamic workspaces. Currently, mainly https://github.com/paperwm/PaperWM/issues/318 keeps me from using that.

PR #364, opened 2021-01-04
Fix showing margin on display edges when switching apps

basdelfos:
When cycling through applications and reaching the begin or the end, didn't display the horizontal margin.

Before:
image

After:
image

### DIFF
--- a/tiling.js
+++ b/tiling.js
@@ -2519,10 +2519,10 @@ function ensuredX(meta_window, space) {
         x = 0;
     } else if (index == 0 && x <= min) {
         // Always align the first window to the display's left edge
-        x = min;
+        x = min + prefs.horizontal_margin;
     } else if (index == space.length-1 && x + frame.width >= max) {
         // Always align the first window to the display's right edge
-        x = max - frame.width;
+        x = max - frame.width - prefs.horizontal_margin;
     } else if (frame.width > workArea.width*0.9 - 2*(prefs.horizontal_margin + prefs.window_gap)) {
         // Consider the window to be wide and center it
         x = min + Math.round((workArea.width - frame.width)/2);
--- a/tiling.js
+++ b/tiling.js
@@ -2515,7 +2515,7 @@ function ensuredX(meta_window, space) {
     let workArea = space.workArea();
     let min = workArea.x;
     let max = min + workArea.width;
-    if (meta_window.fullscreen) {
+    if (meta_window.fullscreen || meta_window.maximized_horizontally) {
         x = 0;
     } else if (index == 0 && x <= min) {
         // Always align the first window to the display's left edge
@@ -2526,7 +2526,6 @@ function ensuredX(meta_window, space) {
     } else if (frame.width > workArea.width*0.9 - 2*(prefs.horizontal_margin + prefs.window_gap)) {
         // Consider the window to be wide and center it
         x = min + Math.round((workArea.width - frame.width)/2);
-
     } else if (x + frame.width > max) {
         // Align to the right prefs.horizontal_margin
         x = max - prefs.horizontal_margin - frame.width;


hedning:
Thanks for taking the time to contribute, unfortunately this is intended behavior. The main points are these:
- The lack of a gap lets the user know they're at the start/end of the space
- It makes it easier to scroll in the first window as you've then got access to an edge

While I'm aware some people would much rather prefer your solution, I'd really like a solution that makes it possible to use the left/right margin to control this. Hope that makes sense :)

jtaala:
Hey @hedning - I know this has been here a while, but happy to address this.

Yes, this seems to be a user preference thing.  I certainly understand and agree with the intended behaviour (and I do appreciate the extra visual information it provides)  

What about having the left/right edge margin user settable?  We already have the horizontal/top/bottom margins settable in the extension settings ui.   We could call it something like "Left tiling edge margin" and "Right tiling edge margin" and have the default 0 (current behaviour).  Or simplify it to just one input "Tiling edge margin".

That way user can set their preferred behaviour.  I admit, I think I would prefer having it > 0 (so I could see my pretty borders...) but not quite as much as normal (so I can still detect when I'm at the edge of my tiled windows).

I'm happy to implement this in the settings ui.  @smichel17 and others - any thoughts/comments?

smichel17:
This was also one of the first things that seemed broken to me when I started using PaperWM. I didn't realize exactly what the trigger conditions were for not showing the margin, so it seemed like sometimes it just didn't work. It's especially jarring for full-width windows, since I understood `<Meta-f>` as "fullscreen (ish)". Even today I almost always use `<Meta-c>` immediately after to center fullscreen windows.

**I think that we don't need to hide the border to communicate "This is the last window" to the user.**

Specifically, when I move my mouse to the very edge of the screen, when there is another window offscreen, a preview is shown. When I click, the preview disappears. I would suggest that instead, the preview should show the next window offscreen. That way you'll know it's the last window when no more preview appears.

Lythenas:
> Specifically, when I move my mouse to the very edge of the screen, when there is another window offscreen, a preview is shown. When I click, the preview disappears. I would suggest that instead, the preview should show the next window offscreen. That way you'll know it's the last window when no more preview appears.

This is already the way it is (at least for me). If there is a window offscreen in that direction a preview is shown. If there is no window in that direction nothing is shown.

But I agree, I think this should be the default behavior, because it feels like a but that sometimes there is a border and sometimes there isn't. Maybe other users would disagree but I feel like I also don't need a setting for this and can just change the behavior.

But also an indicator that there are no more windows in that direction might be nice (but I think not that important). I would imagine more like a flashing screen border or a wiggle when trying to navigate an there are no more windows.

smichel17:
@Lythenas The preview of the next window stays visible after you click?

I think a screen recording might be worth a thousand words here. At 3 seconds in, I click, and the preview disappears. The next window's preview reappears shortly after, when I move my mouse away and back. I'm saying I shouldn't need to move my mouse to get it to show up.

https://user-images.githubusercontent.com/6563664/215345762-aa98bd85-f5d0-45e8-b576-cf2981c0957a.mp4

Lythenas:
Ah that's true. I never noticed it because I rarely navigate with the mouse.

But also that seems like a separate thing/bug from the display edge margin. Maybe create a separate issue for this, if one does not already exist.

smichel17:
Thanks for the suggestion, I opened https://github.com/paperwm/PaperWM/issues/470 and https://github.com/paperwm/PaperWM/issues/471. I think if they were implemented, this could be merged as-is.
Or we could merge first, fix later :smiling_imp:

tiling.js
```diff
@@ -2515,18 +2515,17 @@ function ensuredX(meta_window, space) {
...
+    if (meta_window.fullscreen || meta_window.maximized_horizontally) {
         x = 0;
     } else if (index == 0 && x <= min) {
         // Always align the first window to the display's left edge
-        x = min;
+        x = min + prefs.horizontal_margin;
     } else if (index == space.length-1 && x + frame.width >= max) {
         // Always align the first window to the display's right edge
-        x = max - frame.width;
+        x = max - frame.width - prefs.horizontal_margin;
```
smichel17, line 2525: Does it work to remove these special cases and let the code below handle it?

jtaala:
Hey all, I appreciate the viewpoints we're getting on this.

I've provided #467 which hopes to address this (@Lythenas check it out and let us know).  The reason I provided a simple user-defined option (in the settings UI) to change this is that there are valid opinions for/against this approach - and it seems quite up to user preference.

For example, @hedning described the original intent/reasons for the design and I understand that.  Others seem to prefer it being set the same as horizontal margin etc.  I actually prefer it at 5px - since I can see my nice border (refer to the video in #467), but still also visually see in a static way (without needing a popup/preview etc.) if I'm against the tiling edge.

Hence, given the divide, why wouldn't we provide a simple toggle like #467 provides? - that way, it's up to the user what they prefer.

Anyways, if I had an issue with just relying on the window left/right preview as a substitute is that it's only shown when switching windows... similar to only seeing where you are during switching with the minimap.  I like being able to get visual feedback at all times whether I'm up against a tiling edge without needing to switch or to just remember that.

jtaala:
P.S. if we do keep #467, we could change the default to 10px (which is the default for the horizontal-margin etc.) for new users (or if we think the majority of users prefer it) - without losing our heritage! (i.e. other preferences, like @hedning, myself, and likely others).

jtaala:
> Thanks for the suggestion, I opened #470 and #471. I think if they were implemented, this could be merged as-is. Or we could merge first, fix later smiling_imp

Please see #473 and #476, which fix/completes these issues!

jtaala:
Hey all, I've created PR #488 to replace this (mainly because @basdelfos's original fork/branch is no longer available).

#488 simply removes the `if` cases that cause no margin to be shown - additionally simplifying that area of code (thanks @smichel17!).

Planning on closing this PR in favour of #488.

Cheers!

tiling.js
jtaala: Why yes, yes it does!  See #488 - thanks @smichel17.

PR #368, opened 2021-01-30
Set actor variable for use after `if`

badboy:
Looks like this was recently changed here: bcd92857ae267bb561ee5e96591eeaaf606bd81d

Without this the last line would raise an error because `actor` is undefined.


hedning:
Cheers, must have moved lines inside the if without testing or something 🙃

Issue #355, opened 2020-12-07
Delayed window focus when switching with unusual keybindings


paradajz:
I am using crown on Logitech Craft keyboard to select windows - this keyboard sends VolDown / VolUp events when crown is rotated left/right to PC and I've mapped those keys to behave the same way as Super+Left / Super+Right. When switching the windows with Super+Left/right combination, I'm holding the Super button, press the left or right key few times until I select the window I want, then I release the Super key. The window is focused as Super key is released, which is great. However, with VolDown / VolUp keys combination, there is a delay of about 2 seconds before the window gets focused, which isn't really convenient. Is this something inherent to the way Gnome Shell works or I could maybe modify the PaperWM code somewhere to disable focus delay?

olejorgenb:
Not that I'm aware of. I can reproduce the problem here though. There's some special code for modifer-less keybindings in `navigator.js` (noModsTimeout), but it works fine when binding to eg. `F12`.

gnome-shell spams alot of stuff to the log when using the volume keys though:

```
des. 07 23:14:12 workole-nix gnome-shell[1830]: Window manager warning: Overwriting existing binding of keysym 31 with keysym 31 (keycode a).
des. 07 23:14:12 workole-nix gnome-shell[1830]: Window manager warning: Overwriting existing binding of keysym ff09 with keysym ff09 (keycode 17).
des. 07 23:14:12 workole-nix gnome-shell[1830]: Window manager warning: Overwriting existing binding of keysym ff54 with keysym ff54 (keycode 74).
des. 07 23:14:12 workole-nix gnome-shell[1830]: Window manager warning: Overwriting existing binding of keysym ff52 with keysym ff52 (keycode 6f).
des. 07 23:14:12 workole-nix gnome-shell[1830]: Window manager warning: Overwriting existing binding of keysym 37 with keysym 37 (keycode 10).
des. 07 23:14:12 workole-nix gnome-shell[1830]: Window manager warning: Overwriting existing binding of keysym 39 with keysym 39 (keycode 12).
des. 07 23:14:12 workole-nix gnome-shell[1830]: Window manager warning: Overwriting existing binding of keysym 30 with keysym 30 (keycode 13).
des. 07 23:14:12 workole-nix gnome-shell[1830]: Window manager warning: Overwriting existing binding of keysym ff54 with keysym ff54 (keycode 74).
[...]

PR #372, opened 2021-03-06
Remove navigator timeout for keybindings without mods

paradajz:
This simple change will disable timeouts when using keybindings without modifiers for switching between windows, eg. Volume Up/Down keys. This resolves https://github.com/paperwm/PaperWM/issues/355.


hedning:
Cheers :) Lets just go with 0 delay, as it seemed hard to hit something that would work as intended with a bit more.

PR #392, opened 2021-10-18
Add instruction on windows matching

soraxas:
I was trying to find a way to put a window that has no defined `wm_class` to scratch layer. I had thought of using `title` to match the window and only realised that it had already been implemented without being mentioned in the README.

This PR adds some hints to such an approach for window matching.


smichel17 (approved):
Looks good, but let's merge this after #415, so it's working better before we put more eyes on it.

smichel17:
Thanks, and sorry for taking so long to get around to this (but hey, at shorter than it's been sometimes :)

PR #396, opened 2021-12-23
Support GNOME 40

smichel17:
This is to merge back the community effort to support GNOME 40— https://github.com/PaperWM-community/PaperWM/. See https://github.com/paperwm/PaperWM/issues/376#issuecomment-988178131 for more context.

I'm marking this as a draft because I haven't tried it yet and am not sure what else is needed for it to be merge-ready… but when it is ready, @hedning it _may_ be easier to review the [merged PRs](https://github.com/PaperWM-community/PaperWM/pulls?q=is%3Apr+is%3Amerged+) in the other repo individually, rather than doing it all in one go.

Fixes #376.

PR #397, opened 2021-12-25
Feature/disable workspace animation

YannikSc:
- Disables the workspace animation through a new module
- Fixes keyboard shortcuts, that are using the `grave`/`Above_Tab` key

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -32,7 +32,7 @@ if (!global.workspace_manager) {
  */
 var modules = [
     'tiling', 'navigator', 'keybindings', 'scratch', 'liveAltTab', 'utils',
-    'stackoverlay', 'app', 'kludges', 'topbar', 'settings','gestures'
+    'stackoverlay', 'app', 'kludges', 'topbar', 'settings','gestures', 'animations'
 ];
 
 /**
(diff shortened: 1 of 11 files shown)


mskorzhinskiy:
I've tried your patch, @YannikSc, working great so far. However I must ask: what the difference between this PR and the #396? As I understood, yours are based on it?

If so, may be it worth to suggest this patch to the https://github.com/PaperWM-community/PaperWM/pulls instead?

extension.js
```diff
@@ -32,7 +32,7 @@ if (!global.workspace_manager) {
  */
 var modules = [
     'tiling', 'navigator', 'keybindings', 'scratch', 'liveAltTab', 'utils',
-    'stackoverlay', 'app', 'kludges', 'topbar', 'settings','gestures'
+    'stackoverlay', 'app', 'kludges', 'topbar', 'settings','gestures', 'animations'
```
mskorzhinskiy, line 35: I would also recommend splitting the changes into two separate commits, it would improve experience working with the history: merging, rebasing, reverting changes and so forth.

https://www.aleksandrhovhannisyan.com/blog/atomic-git-commits/

YannikSc:
I messed something up I would say. This PR was supposed to be in the community fork.
And by now I have seen, that this (functional) change is already in the community repo.
So I think this PR can be closed.

PR #414, opened 2022-09-06
allow overriding stylesheet

soraxas:
https://github.com/PaperWM-community/PaperWM/pull/13


smichel17 (approved):
As I wrote in that PR,

> I think this is a good feature to have. Especially, it provides an easier on-ramp for people to become involved in PaperWM development (tweaking their own styles).

The counterpoint would be: it's not that much harder to edit `~/.local/share/gnome-shell/extensions/paperwm@hedning:matrix.org/stylesheet.css` directly. However, this makes it easier to retain your customizations across upgrades.

I'll leave this open for around a week in case anyone else wants to make an argument against it, then merge.

PR #415, opened 2022-09-06
Fix scratch layer matching

soraxas:
https://github.com/PaperWM-community/PaperWM/pull/12

### DIFF
--- a/settings.js
+++ b/settings.js
@@ -302,16 +302,17 @@ function findConflicts(schemas) {
 */
 var winprops = [];
 
-function _winprop_match_p_helper(user_attr_prop, win_prop) {
+function _winprop_match_p_helper(user_attr_prop, _win_prop) {
     // matches user-supplied conditions with null checking
     if (user_attr_prop) {
+        let win_prop = String(_win_prop);
         if (user_attr_prop.constructor === RegExp) {
             // treat prop condition as regex
             if (!win_prop.match(user_attr_prop))
                 return false;
         } else {
             // treat prop condition as string
-            if (user_attr_prop !== String(win_prop))
+            if (user_attr_prop !== win_prop)
                 return false;
         }
     }
--- a/settings.js
+++ b/settings.js
@@ -302,24 +302,30 @@ function findConflicts(schemas) {
 */
 var winprops = [];
 
-function winprop_match_p(meta_window, prop) {
-    let wm_class = meta_window.wm_class || "";
-    let title = meta_window.title;
-    if (prop.wm_class.constructor === RegExp) {
-        if (!wm_class.match(prop.wm_class))
-            return false;
-    } else if (prop.wm_class !== wm_class) {
-        return false;
-    }
-    if (prop.title) {
-        if (prop.title.constructor === RegExp) {
-            if (!title.match(prop.title))
+function _winprop_match_p_helper(user_attr_prop, win_prop) {
+    // matches user-supplied conditions with null checking
+    if (user_attr_prop) {
+        if (user_attr_prop.constructor === RegExp) {
+            // treat prop condition as regex
+            if (!win_prop.match(user_attr_prop))
                 return false;
         } else {
-            if (prop.title !== title)
+            // treat prop condition as string
+            if (user_attr_prop !== win_prop)
                 return false;
         }
     }
+
+    return true;
+}
+
+function winprop_match_p(meta_window, prop) {
+    // check wm_class
+    if (!_winprop_match_helper(prop.wm_class, meta_window.wm_class))
+        return false;
+    // check wm_title
+    if (!_winprop_match_helper(prop.title, meta_window.title))
+        return false;
 
     return true;
 }
--- a/settings.js
+++ b/settings.js
@@ -311,7 +311,7 @@ function _winprop_match_p_helper(user_attr_prop, win_prop) {
                 return false;
         } else {
             // treat prop condition as string
-            if (user_attr_prop !== win_prop)
+            if (user_attr_prop !== String(win_prop))
                 return false;
         }
     }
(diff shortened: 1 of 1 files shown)


settings.js
```diff
@@ -302,28 +302,35 @@ function findConflicts(schemas) {
...
+        let win_prop = String(_win_prop);
+        if (user_attr_prop.constructor === RegExp) {
+            // treat prop condition as regex
+            if (!win_prop.match(user_attr_prop))
                 return false;
         } else {
-            if (prop.title !== title)
+            // treat prop condition as string
+            if (user_attr_prop !== win_prop)
                 return false;
```
smichel17, line 316:
```suggestion
            return user_attr_prop !== win_prop;
```

settings.js
```diff
@@ -302,28 +302,35 @@ function findConflicts(schemas) {
...
 
+function winprop_match_p(meta_window, prop) {
+    // check wm_class
+    if (!_winprop_match_p_helper(prop.wm_class, meta_window.wm_class))
+        return false;
+    // check wm_title
+    if (!_winprop_match_p_helper(prop.title, meta_window.title))
+        return false;
+
+    return true;
```
smichel17, line 331:
```suggestion
    const classMatches = _winprop_match_p_helper(prop.wm_class, meta_window.wm_class);
    const titleMatches = _winprop_match_p_helper(prop.title, meta_window.title);
    return classMatches && titleMatches;
```

This is *slightly* less efficient since both values are always calculated. I think it's worth the readability trade-off, but an alternative is

```suggestion
    return (
        _winprop_match_p_helper(prop.wm_class, meta_window.wm_class)
        && _winprop_match_p_helper(prop.title, meta_window.title)
    );
```

settings.js
```diff
@@ -302,28 +302,35 @@ function findConflicts(schemas) {
...
-        if (prop.title.constructor === RegExp) {
-            if (!title.match(prop.title))
+function _winprop_match_p_helper(user_attr_prop, _win_prop) {
+    // matches user-supplied conditions with null checking
+    if (user_attr_prop) {
+        let win_prop = String(_win_prop);
+        if (user_attr_prop.constructor === RegExp) {
+            // treat prop condition as regex
+            if (!win_prop.match(user_attr_prop))
                 return false;
```
smichel17, line 312:
```suggestion
            return win_prop.match(user_attr_prop);
```
The logic here can be simplified. Also, I think the code here is self-explanatory enough.

smichel17: The whole thing could be condensed further, like,
```
return user_attr_prop.constructor === RegExp
    ? win_prop.match(user_attr_prop)
    : user_attr_prop === win_prop;
 ```
 but in my opinion this is only shorter, not simpler or more readable.

File: settings.js [hunk 4]
```diff
@@ -302,28 +302,35 @@ function findConflicts(schemas) {
...
-        if (!wm_class.match(prop.wm_class))
-            return false;
-    } else if (prop.wm_class !== wm_class) {
-        return false;
-    }
-    if (prop.title) {
-        if (prop.title.constructor === RegExp) {
-            if (!title.match(prop.title))
+function _winprop_match_p_helper(user_attr_prop, _win_prop) {
+    // matches user-supplied conditions with null checking
```
smichel17, line 306: A more descriptive name allows the comment to explain _why_, not just _what_
```suggestion
function condition_matches_winprop(condition, _win_prop) {
    // A null condition matches any value
```
(also requires corresponding name changes below, both in this function and where it is called)

settings.js
```diff
@@ -302,28 +302,35 @@ function findConflicts(schemas) {
...
+    if (user_attr_prop) {
+        let win_prop = String(_win_prop);
```
smichel17, line 308: What's the desired behavior here? In the original code, we had `let wm_class = meta_window.wm_class || "";`, so if the window value was null/undefined, you'd get `wm_class = ""`. Now with `String(_win_prop)`, you'll end up with `"null"` or `"undefined"`.

Is win_prop ever something other than string|null|undefined? If so, you could do `_win_prop ? String(_win_prop) : "";`. Otherwise, I guess `_win_prop || ""` is probably the desired behavior? I suppose a more careful version would be `typeof _win_prop === "string" ? _win_prop : ""`.

settings.js
```diff
@@ -302,28 +302,35 @@ function findConflicts(schemas) {
...
+        if (user_attr_prop.constructor === RegExp) {
```
smichel17, line 309: This is just a question/comment, not a request.

Does `user_attr_prop instanceof RegExp` work here instead? I guess that would allow classes which inherit from RegExp instead of only RegExp itself.

smichel17 (changes requested):
Hi, I'd like to make a few more changes. I know I've already beat up this code a lot, and some of the lines I commented on are carry-overs from the original style, not even something you wrote. Putting my comments all together ends up looking something like this:

```
function condition_matches_prop(condition, _prop) {
    // null condition matches any value
    if (typeof condition !== 'string') {
        return true;
    }

    let prop = String(_prop);

    if (condition.constructor === RegExp) {
        return prop.match(condition);
    }

    // treat prop condition as string
    return condition === prop;
}

function winprop_match_p(meta_window, prop) {
    return (
        condition_matches_winprop(prop.wm_class, meta_window.wm_class)
        && condition_matches_winprop(prop.title, meta_window.title)
    );
}
```

settings.js
smichel17: Checked, it does. Also this fixes the null issue without requiring an explicit null check, which actually keeps the semantics closer to the original code. I think I'm actually just going to apply this fix to the original code, because then it truly just is a fix, whereas actually everything written here so far changes the semantics.

smichel17:
I pulled the `instanceOf` fix into https://github.com/paperwm/PaperWM/pull/392 instead

PR #417, opened 2022-09-07
Explain support for gnome versions before 40

smichel17:
[I'll wait for someone to review/"approve" this PR before I merge it.](https://github.com/paperwm/PaperWM/issues/407#issuecomment-1214966337)

Closes #411

PR #418, opened 2022-09-18
Gnome 42 support

### LINKED ISSUE #409: Support GNOME 42
Most of the code to do this already exists, but it is scattered around with a messy history and needs to be consolidated so it can be merged back here.

For now, if you are on GNOME 42 and you want to use a working PaperWM, use the `gnome-42` branch of @ccope's fork: https://github.com/ccope/PaperWM/tree/gnome-42

edit: Also, you will almost certainly want to install the **[vertical overview](https://github.com/github.com/RensAlthuis/vertical-overview)** extension alongside PaperWM.

hedning:
Fixes the buggy shadow when navigating workspaces by removing the offending box-shadow (probably due to changes in the upstream css engine). The border doesn't really stand out anymore if using the colored backgrounds.

Fixes DnD support. The `ActionDispatcher` is now shared between the keyboard and DnD since `stage.grab` now grabs all pointer input too, ie. the DnD and the keybind dispatcher need to listen for events on the same actor.

No longer does a disable/enable cycle on screen lock: https://gjs.guide/extensions/upgrading/gnome-shell-42.html#metadata-json

Also picked up the mentioned code from https://github.com/ccope/PaperWM/pull/5 (couldn't test the touch support at the moment so left it off).

Only supports 42, so I created a gnome-40 branch, and will create a gnome-42 branch when this is merged.

closes #409, #417


hedning:
Still relies on vertical-overview for gesture support, will probably fix that at some point.

PR #421, opened 2022-09-28
wip: Added touch scroll gestures support

Thesola10:
This allows the user to scroll between windows by swiping on the topbar (or panel). This can be useful for convertibles, or desktops with touch screens.

### DIFF
--- a/gestures.js
+++ b/gestures.js
@@ -144,6 +144,49 @@ function horizontalScroll(actor, event) {
     }
 }
 
+let walk = 0;
+let sdx = null;
+function horizontalTouchScroll(actor, event) {
+  const type = event.type()
+  const [myx, myy] = event.get_coords();
+
+  switch (type) {
+    case Clutter.EventType.TOUCH_BEGIN:
+      this.vx = 0;
+      dxs = [];
+      dts = [];
+      sdx = myx;
+      walk = 0;
+      start = this.targetX;
+      this.hState = Clutter.TouchpadGesturePhase.UPDATE;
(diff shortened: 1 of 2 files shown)


Thesola10:
This current implementation only catches the touch event on the topbar itself. If the gesture leaves the topbar area, it isn't cancelled for some reason, which leaves the workspace in a weird space until the next scroll or touchScroll event.

I guess I'd have to catch the gesture on the global desktop and check if it started on the topbar area, so the scroll gesture continues after leaving the area.

This likely also does not behave properly with "topbar follows monitor focus", it would probably just scroll the primary display even if the topbar is on the secondary display.

paradajz:
Is this configurable? I use this extension:
https://extensions.gnome.org/extension/1616/panel-scroll/

I'd hate to see PaperWM overriding that behaviour.

Thesola10:
@paradajz

> I'd hate to see PaperWM overriding that behaviour.

I don't think it will, this only changes response to touchscreen events, nothing else.

walkiry:
> Is this configurable? I use this extension: https://extensions.gnome.org/extension/1616/panel-scroll/
> 
> I'd hate to see PaperWM overriding that behaviour.

That's a nice extension, however, does not keep track of the order in which they are display, rather when they were displayed.

smichel17:
I don't understand the purpose of this. What does it provide over 3 finger swipe (on wayland)?

Thesola10:
It's for touch screens, not touchpads.

The three finger swipe doesn't exist on touchscreens.

PR #428, opened 2022-10-18
Claim to support Gnome 43

tadfisher:
Seems to be working fine so far, I've been using it for a few days.


smichel17 (changes requested):
Agreed. As long as it works well-enough, we can address bugs via normal issue reports.

Can you also update the versions listed in the README?

gombosg:
https://github.com/paperwm/PaperWM/pull/437 Should address this.

PR #437, opened 2022-11-06
fix: allow loading settings from system directory

gombosg:
Hey! I'm working on packaging for Fedora. In this case we're going to move the `gschema` file to a central system directory and not the local one.

Based on how Pop Shell extension is doing the same.

https://github.com/pop-os/shell/blob/886a069c0582b371e90ac3602b1747ea5fba616c/src/settings.ts#L33-L38

Video: https://www.loom.com/share/71a0ee0eb34043c3b228765a789b4196

### DIFF
--- a/settings.js
+++ b/settings.js
@@ -15,6 +15,7 @@ if (imports.misc.extensionUtils.extensions) {
     }
 }
 
+const Me = imports.misc.extensionUtils.getCurrentExtension();
 var Gio = imports.gi.Gio;
 var GLib = imports.gi.GLib;
 var Gtk = imports.gi.Gtk;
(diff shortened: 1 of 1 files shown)


smichel17 (commented):
As someone not fluent in "Gnome" could you "translate" this for me? In, say, a default Fedora install, which path(s?) would this look in before and after the change?

edit: ideally also you could amend your commit to add the example paths in the commit message

jtaala:
Closing given nil response.

carlwgeorge:
I don't know what happened to @gombosg, but I can answer the question about the paths.  Extensions distributed on https://extensions.gnome.org place the schema file inside the extension directory.

```
❯ unzip -l just-perfection-desktopjust-perfection.v29.shell-extension.zip | grep schema
        0  07-06-2024 13:16   schemas/
    13877  07-04-2024 05:21   schemas/org.gnome.shell.extensions.just-perfection.gschema.xml
```

When packaged as a system-wide extension, the schema file is placed in `/usr/share/glib-2.0/schemas/`, which is shared between all extensions (and applications too).

```
❯ rpm -ql gnome-shell-extension-just-perfection | grep schema
/usr/share/glib-2.0/schemas/org.gnome.shell.extensions.just-perfection.gschema.xml
```

What this patch is doing is checking for the schemas subdirectory in the extension directory, and if it isn't found then checking the shared location.  The logic seems sound to me.  I'm interested in reviving the work to package this in Fedora, so having this merged would help.  Would the developers be interested in a new pull request that rebases @gombosg's work onto the develop branch?

jtaala:
Feel free to create a PR for this - note though that rebasing as is wouldn't work, as from Gnome 45, Gnome moved to ESM modules, so things like `const Me = imports.misc...` won't work.

But it should be simple enough to look at current way of importing other modules in `settings.js` and adapting to suit.

Thanks!

PR #438, opened 2022-11-06
feat: experimental GNOME 43 support

gombosg:
Video: https://www.loom.com/share/b263c439d7fc4f869602f8bd97895f36

Fedora 37 is coming with GNOME 43 so let's start supporting that.


smichel17:
Thanks!

PR #440, opened 2022-11-06
chore: rename schemas file

gombosg:
This is just a cosmetic improvement, the compiled file doesn't change at all after compiling.


smichel17:
Makes sense to me. PaperWM hasn't been [ScrollWM](https://github.com/paperwm/ScrollWM) for a while.

PR #443, opened 2022-11-08
fix: moving windows between workspaces should not take windows in between

### LINKED ISSUE #442: Moving a window to a distant workspace via keybindings takes windows from workspaces in between
## Summary

When moving a window to another workspace at least 2 away, the "Move window one workspace up/down" keybindings will take windows from workspaces in between.

For example, if I have workspaces A, B, C, and I want to move a window from A to C, I want to press "Move window one workspace down" twice: one to go from workspace A->B, and one to go from workspace B->C. Since I have the keybinding set to Super+Shift+PageDn, I expect to hold down Super+Shift and then tap PageDn twice to move just that window and no other windows.

## Steps to reproduce

- Create three workspaces
- Open windows in the first two workspaces
- In the first workspace, hold the modifier key part of the "Move window one workspace down" keybinding and tap the "activator" key (not sure terminology here). The navigator should move to the second workspace.
- In the second workspace, continue holding the modifier keys and tap the "activator" key again

## Expected behavior

In the "Move window one workspace up/down" keybinding, pressing and holding modifier keys and tapping the "activator" key the first time should take the focused window from the original workspace and enter the workspace navigation mode.

While the user holds the modifier keys, the move operation is considered in-progress, and PaperWM should not take additional windows when the user presses the "activator" keys to move between workspaces, unless the user explicitly presses the distinct "Take window" keybinding. 

When the user releases the modifier keys, PaperWM should insert the taken windows into the destination workspace.

## Actual behavior

PaperWM will take the window from the first workspace and navigate to workspace B. When the user presses the activator key a second time, PaperWM will also take a window from workspace B and add it to the stack of windows to move. This is not what I intend, since I don't want to modify windows in workspaces I'm paging through.

## Mitigation

I thought this behavior might be unintentional, so I put together a quick fix on my local repo. I'll open a PR and link to it here.

lediur:
## Description

This PR adds a check to the `selectSequenceSpace` function in `tiling.js` so that it only takes windows during a move operation if there are no windows in the current navigator's `_moving` stack. 

This allows the move operation to take the focused window from the workspace it was originally triggered in, while preventing it from taking additional windows from workspaces that the user might navigate through on the way to their destination.

## Motivation

Fixes #442. See that issue for a more detailed description.

### DIFF
--- a/tiling.js
+++ b/tiling.js
@@ -1835,10 +1835,15 @@ var Spaces = class Spaces extends Map {
         let from = monitorSpaces.indexOf(this.selectedSpace);
         let newSpace = this.selectedSpace;
         let to = from;
+
         if (move && this.selectedSpace.selectedWindow) {
-            takeWindow(this.selectedSpace.selectedWindow,
-                       this.selectedSpace,
-                       {navigator: Navigator.getNavigator()});
+            const navigator = Navigator.getNavigator();
+
+            if (navigator._moving == null || (Array.isArray(navigator._moving) && navigator._moving.length === 0)) {
+                takeWindow(this.selectedSpace.selectedWindow,
+                    this.selectedSpace,
+                    {navigator: Navigator.getNavigator()});
+            }
         }
 
         if (direction === Meta.MotionDirection.DOWN)
--- a/tiling.js
+++ b/tiling.js
@@ -1842,7 +1842,7 @@ var Spaces = class Spaces extends Map {
             if (navigator._moving == null || (Array.isArray(navigator._moving) && navigator._moving.length === 0)) {
                 takeWindow(this.selectedSpace.selectedWindow,
                     this.selectedSpace,
-                    {navigator: Navigator.getNavigator()});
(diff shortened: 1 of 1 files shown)


azmeuk:
I can confirm this fixes the behavior for me (GNOME 43, Wayland, AMD Radeon™ RX 550 Series)

jtaala:
Same here - confirm fixes #442 for me.  Many thanks @lediur!

lediur:
@jtaala thanks! anything I need to do to get this mergeable? do you mind giving it a review :)

jtaala (approved):
Thanks @lediur for your patience.
This is good, this behaviour does fix the issue - that is, when moving a selected window up/down to another workspace (using the workspace keybinding for `Move window one workspace up/down`)  - only `take` the window if no other window is selected (checks look good - e.g. check for window selection is if _moving is null or is an empty array).

I've also been using this for the last few months and can confirm no regressions or issues seen.

Issue #441, opened 2022-11-08
Windows switching 


xvladqt:
Hi, i have a bug with paperwm after upgrade gnome to 43, when i use hotkeys for switch windows after release button they rollback to original window. 

Video demo - https://drive.google.com/file/d/1ORxH4GBS_-8ItRen61SBhUXlpJcllWCr/view

mateotomas2:
Same problem here using super + left / right. Super + tab works.

autumn-birds:
I am also having this issue!  Super+tab works for me as well, though it's not ideal.

It started to occur post-rebooting after a full system upgrade.  My setup is Arch on zfs, so I have a snapshot of how the system was previously. Here are the versions of gnome-shell on the main system and on the snapshot--

*main system:*
```
Name            : gnome-shell
Version         : 1:43.1-1
Description     : Next generation desktop shell
[...snip...]
Optional Deps   : gnome-control-center: System settings [installed]
                  evolution-data-server: Evolution calendar integration
                  [installed]
                  gst-plugins-good: Screen recording [installed]
                  gst-plugin-pipewire: Screen recording
                  gnome-bluetooth-3.0: Bluetooth support [installed]
                  power-profiles-daemon: Power profile switching
```

*pre-upgrade snapshot:*
```
Name            : gnome-shell
Version         : 1:43.0-2
Description     : Next generation desktop shell
[...snip...]
Optional Deps   : gnome-control-center: System settings [installed]
                  evolution-data-server: Evolution calendar integration
                  [installed]
                  gst-plugins-good: Screen recording [installed]
                  gst-plugin-pipewire: Screen recording
                  gnome-bluetooth-3.0: Bluetooth support [installed]
```

Hopefully this is helpful.  I'm going to look into whether downgrading just gnome-shell is possible and, if so, whether that helps.  My PaperWM appears to be up-to-date with current `develop`, except for some 'upstream' changes in README.md (I had manually merged in the supported-version bump PR for gnome shell 43.)

autumn-birds:
update: downgrading gnome-shell alone didn't help.  Booting from a clone (like a fork) of the previous snapshot did, so I assume a change in something other than gnome-shell *might* be causing this.  I don't have time to debug further and will just try to get used to using super+tab for now, but if it would be helpful I might be able to look at potentially coming up with a list of what packages changed at some point.

mateotomas2:
[...]

PR #446, opened 2022-11-29
fix: cannot switch windows using navigator on X11

mateotomas2:
## Solution
Just moved the `dismissDispatcher(Clutter.GrabState.KEYBOARD)` after the `nav.destroy()`

## Info
When the keyboard dispatcher is dismissed before destroying the navigator, a focus signal is emitted to the starting window. This causes a ensureViewport, that changes the selected window to the first one, overriding the selected one. Since the logic that activate the new window is done inside the destroy function, the selected window is wrong after the dismissDispatcher.

Not sure why this happens only on X11 and on the last GNOME version. I don't understand why the focused window is focused a second time, like if the navigator is stealing the focus.

### DIFF
--- a/navigator.js
+++ b/navigator.js
@@ -223,8 +223,8 @@ function getModLock(mods) {
     _finish(timestamp) {
         let nav = getNavigator();
         getNavigator().accept();
-        dismissDispatcher(Clutter.GrabState.KEYBOARD)
         !this._destroy && nav.destroy();
+        dismissDispatcher(Clutter.GrabState.KEYBOARD)
     }
 
     destroy() {


smichel17:
I'm not on gnome 43 yet so I cannot confirm the fix. However, I will run this for a few days to make sure it doesn't introduce any other regressions. @tadfisher @Gelbana @hedning I'd also appreciate if someone else could give this a code review— should be relatively quick (though not as quick as the 1-line change would suggest!)

xvladqt:
Thanks, it's fixed for me (#441)

scottames:
I'm running Gnome 43 and have been running on this branch for 1+ day now. Can confirm it fixes the issue and doesn't seem to introduce any further issues AFAICT.

azmeuk:
I can confirm this fixes the behavior for me too (GNOME 43, Wayland, AMD Radeon™ RX 550 Series)

Gelbana:
I've been running this for a week with no issues and I can't see any obvious regressions from this change.

smichel17:
Me too. No (new) issues.

PR #452, opened 2022-12-23
fix: hide topbar for fullscreened floating and scratch windows (fixes #451)

### LINKED ISSUE #451: Top bar does not hide when floating or scratch window goes full screen
## Summary
The gnome top bar does not hide when floating windows or windows on the scratch layer go full screen.

I often like to have a video (or something) playing on the scratch layer while I work.  Often (on a particularly good or action-packed part of a video...) I want to make it full-screen - it does go full-screen but the top bar is still there - ruining the moment! :-)

## Steps to reproduce
1. open a window (I would suggest mpv or a media player for example);
2. detach the window into the scratch layer;
3. make the window full-screen (for example, in mpv just double-click in the window);
4. observe that the gnome top bar does not hide;

## Expected behaviour
The top bar should hide (just like it does if the window was not in the scratch layer);

jtaala:
Fixes #451.

Checks if the window is fullscreen and on the scratch layer (or is fullscreen and floating), if so it hides the topbar.

### DIFF
--- a/topbar.js
+++ b/topbar.js
@@ -574,34 +574,35 @@ function disable() {
 }
 
 function fixTopBar() {
-    let spaces = Tiling.spaces
-    if (!spaces)
-        return;
-    let space = spaces.monitors.get(panelMonitor);
+    let space = Tiling.spaces?.monitors.get(panelMonitor) ?? false;
     if (!space)
         return;
+    // normal ==> not in overview AND not in tiling workspace view
     let normal = !Main.overview.visible && !Tiling.inPreview
-    let selected = spaces.monitors.get(panelMonitor).selectedWindow
+    // selected is current (tiled) selected window (can be different to focused window)
+    let selected = space.selectedWindow
+    // current focused window (can be different to selected tile window, e.g. focused scratch window)
     let focus = display.focus_window
+    // check if a scratch window is currently focused
     let focusIsScratch = focus && Scratch.isScratchWindow(focus)
-    let fullscreen = selected && selected.fullscreen && !(focusIsScratch);
-    let hideTopBar = !spaces.monitors.get(panelMonitor).showTopBar
-    // hide topbar for fullscreen scratch windows
-    if (focusIsScratch && focus.fullscreen) {
-        panelBox.hide();
-        return;
-    }
-    if (normal && hideTopBar) {
+    // check if currently fullscreened (check both focused scratch and tiled selected)
+    let fullscreen = focusIsScratch ? focus.fullscreen : selected && selected.fullscreen;
+    
+    // if normal view and current workspace is set NOT to show topbar ==> hide topbar
+    if (normal && !space.showTopBar) {
         // Update the workarea to support hide top bar
         panelBox.scale_y = 0;
         panelBox.hide();
         return;
     }
+
+    // if normal view and window (either tiled or scratched) is fullscreen ==> hide topbar
     if (normal && fullscreen) {
         panelBox.hide();
         return;
     }
 
+    // fallback: if here, not fullscreened etc. ==> show topbar
     panelBox.scale_y = 1;
     panelBox.show();
 }
--- a/topbar.js
+++ b/topbar.js
@@ -585,8 +585,8 @@ function fixTopBar() {
     // current focused window (can be different to selected tile window, e.g. floating/scratch window)
     let focused = display.focus_window;
     // check if focused window is floating or a scratch window
-    let focusIsFloatOrScratch = focused && space.isFloating(focused) || focused && Scratch.isScratchWindow(focused);
-    // check if is currently fullscreened (address both focused scratch / tiled selected windows)
+    let focusIsFloatOrScratch = focused && (space.isFloating(focused) || Scratch.isScratchWindow(focused));
+    // check if is currently fullscreened (address both focused scratch, floating or tiled windows)
     let fullscreen = focusIsFloatOrScratch ? focused.fullscreen : selected && selected.fullscreen;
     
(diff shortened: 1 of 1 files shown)


hallettj (dismissed):
I have been seeing this issue in a game running under Wine. I can confirm that this change fixes the problem. Once again I appreciate your fix!

jtaala:
Morning all! @Gelbana @tadfisher @ccope @smichel17  - here's another quick fix to review (if needed, noting that @hallettj approved this this PR previously - thanks @hallettj!) when someone has time.

Simply also hides the top bar when a window on the scratch layer (or a floating window) goes fullscreen.

topbar.js
```diff
@@ -586,6 +586,11 @@ function fixTopBar() {
     let focusIsScratch = focus && Scratch.isScratchWindow(focus)
     let fullscreen = selected && selected.fullscreen && !(focusIsScratch);
     let hideTopBar = !spaces.monitors.get(panelMonitor).showTopBar
+    // hide topbar for fullscreen scratch windows
+    if (focusIsScratch && focus.fullscreen) {
```
smichel17: One particular thing I would look for: `focusIsScratch` currently includes a check to make sure that `focus` is not null, so it's safe to check `focus.fullscreen` here. But there's no guarantee `focusIsScratch` will keep that check, so it would be easy to end up with a crash here from trying to access a property of null/undefined.

However this ends up, I'd either try to get it to match the `if (!focus) return` pattern so it's safe to use in the rest of the function, or pull things out so there's an explicit `focus && focus.fullscreen` somewhere or other.

smichel17 (changes requested):
This may work, but it makes already messy logic even messier.

Why do we exclude scratch windows from our `fullscreen` check variable, but then immediately check if a window is scratch and fullscreen? Why do we have a `hideTopBar` variable but then it's only one of three situations where we hide the top bar?

I think some combination of changing what goes into the intermediate variables and naming them more descriptively would do wonders here. Also (this one's a personal preference), I would swap the early `return`s at the end for a chain of `else if`s with the last two lines of `fixTopBar` going in the closing `else` (~~if there are still 3 cases post-refactor, at least~~ edit: actually I just strictly prefer it; I'd do it even with a single if/else).

topbar.js
jtaala: yes, good call tbh - it was/is quite messy and a refactor is probably in order!

I'll relook at the implementation of this and go for a cleaner approach/refactor.

smichel17: I think I'd start with something like this (if I'm guessing correctly about what `selected` vs `focus` means) and go from there. It doesn't fully address the null check issue, but it moved them closer together at least, and I think would simplify the logic below.
````diff
-let fullscreen = selected && selected.fullscreen && !(focusIsScratch);
+let fullscreen = focusIsScratch ? focus.fullscreen : selected && selected.fullscreen;
````

Comment (jtaala) [reply topbar.js]: Thanks @smichel17, so, I've simplified the current fixTopBar() function as a first step (see current commit) - and added comments detailing (my) current understading/interpretation of what is happening in the code.

Let me know if that aligns with what you're seeing/thinking (especially the diff between `selected` and `focused` window).

Using your suggested logic for `fullscreen` boolean (looks good to me!).

I've been testing this simplification (no issues so far for both tiled/scratch fullscreen windows) and will continue testing this.

Comment (jtaala) [reply topbar.js]: Thanks @smichel17 - please checkout the latest refactor.  Requesting a re-review.  Cheers!
edit: yes, chain of if/else's makes sense here and improves readability and flow.

**Comment by jtaala** [comment]
Huzzah! found a case not addressed: _floating windows_.  Floating windows are not scratch windows but should also have the topbar hide (floating windows can be fullscreen'd but their top is then not visible since the topbar is covering them).  Latest commit addresses these windows too.

An example of a floating window is the PaperWM extension settings - it opens and floats but is not on the scratch layer.

File: topbar.js [hunk 2]
```diff
@@ -574,30 +574,35 @@ function disable() {
 }
 
 function fixTopBar() {
-    let spaces = Tiling.spaces
-    if (!spaces)
-        return;
-    let space = spaces.monitors.get(panelMonitor);
+    let space = Tiling.spaces?.monitors.get(panelMonitor) ?? false;
```
jtaala, line 577: Just to explain this line (in case others aren't use to such js operators) - it's using an optional chain with the nullish coalescing operator `??`.

Basically, if `Tiling.spaces` is null (or undefined) then the nullish coalescing operator (`??`) means `space` will be assigned to `false`; otherwise it will return the left hand side of `??`, with the optional `?.` function applied, meaning `space` will then be assigned the value of `Tiling.spaces.monitors.get(panelMonitor)` (since `Tiling.spaces` is not `null` and not `undefined`).

jtaala:
@Gelbana @tadfisher @ccope @smichel17 @hallettj - any further comments on this one?  I believe it's ready to go and the rewrite is much clearer as to the intent and function of `fixTopBar` (plus it now catches, I believe, all the cases - e.g. if a window is fullscreen'd AND (tiled or floating,or on the scratch layer).

smichel17 (dismissed):
Maybe a little overcommented now (main useful distinction was focused vs selected, since the latter is a paperwm-specific concept that I didn't understand at first). But looks good.

jtaala:
> Maybe a little overcommented now (main useful distinction was focused vs selected, since the latter is a paperwm-specific concept that I didn't understand at first). But looks good.

Lol - I actually read through it again yesterday and thought the same thing!  

I'll clean up the comments (not sure it it would need to be approved again though - don't want to make more work for you!).

smichel17 (approved):

PR #454, opened 2022-12-26
Make minimap scale (size) user-configurable through PaperWM extension settings

### LINKED ISSUE #457: Right-side window preview (StackOverlay) is broken / not showing
## Summary
The right-side window preview (StackOverlay) that should appear when you have multiple windows tiled and you place mouse cursor on the right side of the screen does not show.  Left side shows and behaves as expected.

## Steps to reproduce
1. open a bunch of windows so that some windows are no longer showing;
2. place your mouse cursor on the left-side of the screen - observe that a window preview (or the next left window) shows;
3. move to the left most window (or just left a few times/windows);
4. now place your cursor on the right side of the screen;
5. observe that no window preview is shown;

## Expected behaviour
Right-side window previews should be shown (according to [triggerPreview()](https://github.com/paperwm/PaperWM/blob/7c0863c944a02d4e8095034403bff6ade3579091/stackoverlay.js#L270-L274) in stackoverlay.js).

jtaala:
This PR makes the mini-map size user configurable.  It also fixes #457 and uses the user defined mini-map scale value for StackOverlay window previews as well.

I prefer the mini-map tile size to be a little larger than the current default (15%).  This PR adds a setting to PaperWM extension settings which allows users to select their own mini-map scale setting (from 5% to 95%).   

This setting corresponds to the "scale" of mini-map tiles as compared to their actual window size (e.g. "15" corresponds to a mini-map tile size being approximately 15% of the actual window size).

Changes are applied immediately (no need to logout/login etc.).

The default value is set to the original MINIMAP_SCALE value that was defined in minimap.js.

image


jtaala:
@Gelbana @tadfisher @ccope @smichel17  - here's a new feature to review when someone has time.

This one allows users to set the minimap scale in extension settings.

_Note this PR also implements fix #458 and sets the the StackOverlay preview size to the user defined scale._

jtaala:
@smichel17 to align with the other PRs (like #453 and #456) I've changed the setting value display (in extension settings) to also be percentage format (e.g. `30` instead of `0.30`).  Internally, we use standard format though (e.g. `0.3`).

Cheers!

jtaala:
All, this has been up and ready to go for 3 weeks - did the invites for comment and haven't heard any complaints etc.

Have been testing this new feature in anger for the past 3 weeks and haven't seen any regressions or issues.  Will merge this now.

smichel17:
This looks good to me, no suggestions. I have two comments for the future, though:

1. The preferences screen is starting to get large. Not to the point where it is a problem, but if/when more settings are added, we'll have to take a pass at organizing them. Next time you want to add a setting, let's have a more detailed discussion. In short, I'm a fan of apps that have more settings than are exposed in their UI (e.g. set them via user.js or directly modifying dconf values, without putting them in the official settings dialog). I think it's a good balance between flexibility and not overwhelming the user when they first open the settings. Firefox's `about:config` is a good example of this.
    - I think it's good for this to be in the UI but, on reflection, "useful window widths" might be more appropriate to leave as a "hidden" setting, where perhaps we can fix the problem by providing better defaults, instead of asking users to pick their own values.
2. In the future we might want to make this setting affect the size of other UI elements created by the app, and rename it to something like "UI scale".

jtaala:
Yeah, good points - this setting seemed simple enough and fit well enough to add it to the "Settings" section (p.s. it used to be called "Toggles").

Re the "Useful window widths/heights" - yeah I struggled a bit with knowing where to stick them... I toyed with creating a new settings section or tab - in the end I thought where they ended up was kind of okay (but not ideal).

smichel17:
Agreed that it's good enough for now, with how few settings there are. The more settings, the better they need to be organized.

PR #456, opened 2022-12-27
Make "useful window widths" and "useful window heights" user-configurable through PaperWM extension settings

### LINKED ISSUE #316: Add a setting for the default window size(s)
This scrollable tilling scheme seems to work really well with my super-ultrawide monitor. However, I feel that the smallest default width (seems to be about 30~40% of the screen width) is a bit too large, and I don't find anything in the settings to change it. It would be great if I can choose the sizes given by ctrl-r.

### LINKED ISSUE #402: Make Useful Window Widths/Heights configurable via the Extensions tool
https://github.com/paperwm/PaperWM/issues/285 shows that the settings exist, they're just hidden?

jtaala:
Resolves #316, resolves #402 

This PR makes the "useful window widths" and "useful window heights" user-settable through the extension pref settings ui.

Inputs supported are percentage or pixel values, e.g. `38.195%; 50%; 61.804%` or `400px; 800px; 900px`.  Mixed input types (e.g. input with `%` AND `px` in it) are not supported and will cause the input box to turn red (error css class), as will any invalid input:
image
image
image

Changes are applied immediately (no need to logout/login etc.).

Reset buttons are also provided which will reset width/height values back to the default PaperWM cycle values.

Feedback welcome!

image

### DIFF
--- a/Settings.ui
+++ b/Settings.ui
@@ -289,6 +289,8 @@
                         </child>
                         <child>
                           <object class="GtkEntry" id="cycle_widths_entry">
+                            <property name="width_chars">24</property>
+                            <property name="max_width_chars">24</property>
                             <layout>
                               <property name="column">2</property>
                               <property name="row">0</property>
@@ -305,7 +307,7 @@
               <object class="GtkListBoxRow" id="general_row_5">
                 <property name="activatable">False</property>
                 <property name="focusable">False</property>
-                <property name="tooltip_text" translatable="yes">Comma separated values representing the "useful window heights" that will be cycled through. Values can be either ratio or pixel values. Mixed ratio and pixel values are not supported.</property>
+                <property name="tooltip_text" translatable="yes">Semicolon separated values of "useful window heights" that will be cycled through. Values types can be either percentage (of available screen width), e.g. "50%" or pixel values, e.g. "500px". Mixed value types are not supported.</property>
                 <child>
                   <object class="GtkGrid" id="workspace_grid12">
                     <property name="focusable">False</property>
@@ -341,6 +343,8 @@
                         </child>
                         <child>
                           <object class="GtkEntry" id="cycle_heights_entry">
+                            <property name="width_chars">24</property>
+                            <property name="max_width_chars">24</property>
                             <layout>
                               <property name="column">1</property>
                               <property name="row">0</property>
--- a/Settings.ui
+++ b/Settings.ui
@@ -249,6 +249,110 @@
                 </child>
               </object>
             </child>
+            <child>
+              <object class="GtkListBoxRow" id="general_row_4">
+                <property name="activatable">False</property>
+                <property name="focusable">False</property>
+                <property name="tooltip_text" translatable="yes">Ratio values (comma or whitespace separated) representing the "useful window widths" that will be cycled through</property>
+                <child>
+                  <object class="GtkGrid" id="workspace_grid3">
+                    <property name="focusable">False</property>
+                    <property name="margin_start">12</property>
+                    <property name="margin_end">12</property>
+                    <property name="margin_top">6</property>
+                    <property name="margin_bottom">6</property>
+                    <property name="column_spacing">32</property>
+                    <child>
+                      <object class="GtkLabel">
+                        <property name="focusable">False</property>
+                        <property name="hexpand">1</property>
+                        <property name="label" translatable="yes">Useful window widths</property>
+                        <property name="xalign">0</property>
+                        <layout>
+                          <property name="column">0</property>
+                          <property name="row">0</property>
+                        </layout>
+                      </object>
+                    </child>
+                    <child>
+                      <object class="GtkGrid">
+                        <layout>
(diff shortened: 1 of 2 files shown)


jtaala:
@ccope could you review / test this one when you get a chance?

jtaala:
@Gelbana @tadfisher @smichel17  - just FYI, asked ccope to review this one (since he requested it in #402).

Thoughts or any improvements suggestions super welcome!

Settings.ui
```diff
@@ -249,6 +249,110 @@
                 </child>
               </object>
             </child>
+            <child>
+              <object class="GtkListBoxRow" id="general_row_4">
+                <property name="activatable">False</property>
+                <property name="focusable">False</property>
+                <property name="tooltip_text" translatable="yes">Comma separated values representing the "useful window widths" that will be cycled through. Values can be either ratio or pixel values. Mixed ratio and pixel values are not supported.</property>
```
smichel17: > ratio or pixel values

For consistency with #453, I'd suggest only supporting `___px` and `__%`, not plain numbers.

prefs.js
```diff
@@ -109,6 +109,36 @@ var SettingsWidget = class SettingsWidget {
             this._settings.set_int('vertical-margin-bottom', bottomMargin.get_value());
         });
 
+        let cycleWidths = this.builder.get_object('cycle_widths_entry');
+        let widthArray = this._settings.get_value('cycle-width-steps').deep_unpack();
+        cycleWidths.set_text(widthArray.toString().replaceAll(',', ', '));
+        cycleWidths.connect('changed', () => {
+            let varr = cycleWidths.get_text()
+                .replaceAll(',', ' ')
+                .split(/\s+/)
```
smichel17: > can be entered as comma (or whitespace) separated values

I would be stricter with how you parse this.

1. Require a certain delimiter, rather than allowing the user to mix whitespace and commas. Fine to also allow the other one to be present optionally (`, ` separators are easier for *me* to read so I'd like if they were *supported*), but I would require one or the other as the character we actually split on. I don't think there's any edge cases where this matters right now; it's just a case of "why make parsing harder for ourselves if it's not strictly necessary?"
    - I recommend making the whitespace the optional bit. I don't think it will be relevant here, but making whitespace not be a separator leaves open the _option_ of doing crazy things like `max 50% 1000px` as a value. Not saying we _should_ add anything along those lines, just that it's nice to have the flexibility if needed.

2. I'd use a semicolon as a separator, rather than a comma. In much of Europe, comma is a decimal place separator, so it's something we might someday want to support (`33,33%`).

```suggestion
        cycleWidths.connect('changed', () => {
            let varr = cycleWidths.get_text()
                .split(';')
                .map((s) => s.trim())
```

Note: this is (obviously) not the only change that would be needed, just a convenient place to hang this thread.

smichel17 (changes requested):

Settings.ui
jtaala: yeah, I agree, especially if #453 switching to 'px' or '%'.

Cheers.

prefs.js
jtaala: Yep, makes sense.  The original implementation was quite lenient on input format since I only wanted some delimited number array (hence replacing and splitting on whitespace chunks and then mapping to Number worked well).  Given we'll be switching to `%` and `px` strings trimming and processing seems in order.

jtaala: Okay!  requesting a re-review here.  Have implemented all these suggestions.  Parsing is stricter and am enforcing units on each input.  Thanks @smichel17!

smichel17:
On reflection, I use "useful window widths" a _ton_ but I don't think I've ever actually used "useful window heights" (and in fact, I didn't realize it was a thing until just now). Is it something you added because you wanted to change it, or because you were also making the window widths configurable and it felt like the heights should be, too?

jtaala:
> On reflection, I use "useful window widths" a _ton_ but I don't think I've ever actually used "useful window heights" (and in fact, I didn't realize it was a thing until just now). Is it something you added because you wanted to change it, or because you were also making the window widths configurable and it felt like the heights should be, too?

Heights were there, I actually use them quite a bit as I have a layout I like for several applications and being able to set a specific height (that I use) was part of the early intent of this PR.  So bit of a mix - they were there (so might as well have a setting for them like widths), plus I wanted to change/use them anyway.

jtaala:
Yep, I guess I wouldn't mind have the width/height settings hidden / or changing them through dconf, a few thoughts though:

- having them in settings allows changing them without needing a shell-reload (which is handy for fine-tuning without needing restart/login-logout if in wayland)... but there's ways to keep that behaviour with changes via gsettings/dconf;
-  it's much easier to let users know when they input invalid inputs if in settings (e.g. input boxes turn red) - users will probably try to input/mix % and px values (or some other invalid inputs) and currently the box turns red letting them know when inputs aren't right.  We could throw an error user notification when they try to use the invalid widths/heights though... so maybe this one isn't such a big deal.

PR #458, opened 2022-12-28
fix: right-side StackOverlay window preview not showing

### LINKED ISSUE #457: Right-side window preview (StackOverlay) is broken / not showing

jtaala:
Fixes #457.

Sets right-side window preview's x position directly (using the used scale and `clone.width` instead of `get_transformed_size()`).

### DIFF
--- a/stackoverlay.js
+++ b/stackoverlay.js
@@ -268,7 +268,7 @@ var StackOverlay = class StackOverlay {
 
             let monitor = this.monitor;
             if (this._direction === Meta.MotionDirection.RIGHT)
-                x = monitor.x + monitor.width - clone.get_transformed_size()[0];
+                x = monitor.x + monitor.width - (0.15 * clone.width);
             else
                 x = monitor.x;
             clone.set_position(x, y);


jtaala:
@Gelbana @tadfisher @ccope @smichel17  - apologies for the lot of PRs I'm pinging you about! but here's another quick fix for review that fixes the right-side stack overlay from not showing.

I'm not 100% sure why `get_transformed_size()` for the clone stopped working - but it does seem to be related https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/4281 as I get similar stack trace errors in both x11/wayland.  

In any case the proposed fix should be a much more direct/robust way for calculating the clone's x position for the right-side preview.

jtaala:
No longer required as #454 was merged (which fixed #457).

PR #461, opened 2023-01-02
No longer minimise all scratch windows when a single scratch window is unscratched

### LINKED ISSUE #455: Unscratching a window causes all other windows on the scratch layer to minimise
## Summary
Having multiple windows on the scratch layer, and then unscratching a single window causes all other windows on the scratch layer to minimize.

This looks like the intended behaviour? (it's caused this [hide()](https://github.com/paperwm/PaperWM/blob/7c0863c944a02d4e8095034403bff6ade3579091/scratch.js#L138) method in the `toggle(metaWindow)` method in `scratch.js`).

Just wondering what the reasoning was/is for this behaviour.

## Steps to reproduce
1. open some windows and put them all on the scratch layer;
2. focus on a single window (on the scratch layer) and unscratch it;
3. observe that all other scratch windows minimise;

## Expected behaviour
I would have thought that unscratching a single window should just do that - and not anything else to other windows on the scratch layer.

Curious as to the intent or reasoning for minimizing other scratch windows.  There's probably a use-case I'm missing.

## Solution
It's a quick fix: simply removing the `hide()` method from https://github.com/paperwm/PaperWM/blob/7c0863c944a02d4e8095034403bff6ade3579091/scratch.js#L138 would stop this behaviour.

jtaala:
Resolves #455.

As mentioned in issue #455 I'm not sure why this behaviour was implemented - that is, what the use-case was for having every other scratch window minimize when one of the scratch windows is unscratched (which seems unexpected and annoying...).

### DIFF
--- a/scratch.js
+++ b/scratch.js
@@ -135,7 +135,6 @@ function unmakeScratch(metaWindow) {
 function toggle(metaWindow) {
     if (isScratchWindow(metaWindow)) {
         unmakeScratch(metaWindow);
-        hide();
     } else {
         makeScratch(metaWindow);
 


jtaala:
I'm going to merge this one in as there's been no comments and I still can't see the original intent for the previous behaviour (described in #455).

PR #464, opened 2023-01-14
Set only the workspace with focus as active

Lythenas:
This will set only the currently focused workspace active. I.e. the border of the active window on all non-focused workspaces is dimmed. This helps see which monitor is currently focused.

~~Note: The try-catch is likely not necessary but I don't really know what I'm doing here^^ So I'm open to suggestions.~~

### DIFF
--- a/tiling.js
+++ b/tiling.js
@@ -2734,6 +2734,12 @@ function ensureViewport(meta_window, space, force) {
 function updateSelection(space, metaWindow) {
     let clone = metaWindow.clone;
     let cloneActor = clone.cloneActor;
+
+    // first set all selections inactive
+    // this means not active workspaces are shown as inactive
+    setAllWorkspacesInactive();
+
+    // then set the new selection active
     space.setSelectionActive();
     if (space.selection.get_parent() === clone)
         return;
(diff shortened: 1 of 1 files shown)


tiling.js
```diff
@@ -2839,13 +2845,31 @@ function grabEnd(metaWindow, type) {
     inGrab = false;
 }
 
+function setAllWorkspacesInactive() {
+    try {
+        for (let i = 0; i < workspaceManager.get_n_workspaces(); i++) {
+            const ws = workspaceManager.get_workspace_by_index(i);
```
jtaala: maybe change this to `let` (moreso  convention, although `const` works here in this use case `let` is more appropriate), e.g.

```javascript
let ws = ...
```

tiling.js
```diff
@@ -2839,13 +2845,31 @@ function grabEnd(metaWindow, type) {
     inGrab = false;
 }
 
+function setAllWorkspacesInactive() {
+    try {
```
jtaala: my thoughts on the try catch: not really needed (although isn't hurting) - you already do a check on
```javascript
if (wm) {...
```

jtaala:
@Lythenas, I really like this PR btw - makes sense and does make it much easier to use with multi-monitors.  Good work!

We could also implement this as a user-selectable option in the PapermWM extension settings (in case others think this behaviour may depend on user preference).  

If this is the case, I'm happy to help implement it in extension settings (I'd vote for it to be on by default).

jtaala (changes requested):
Added a couple of suggested (small) changes in comments, e.g.

1. change `const ws =` to `let`;
2. remove the try/catch to simplify and make it more readable etc.

Just suggestions - any other thoughts anyone?

Lythenas:
I integrated your suggestions.

While a setting for this would certainly be possible, I personally don't think there is any downside to always activating it. But it changes previous behavior so maybe other users will want to disable it.

jtaala:
> I integrated your suggestions.
> 
> While a setting for this would certainly be possible, I personally don't think there is any downside to always activating it. But it changes previous behavior so maybe other users will want to disable it.

Yes, I would agree.  I guess if others request it as an option we could implement an option for them - I like it and also don't see a downside (feels like it should be the default behaviour anyway).

jtaala (approved):

smichel17:
Agreed, no need for a setting here.

PR #467, opened 2023-01-22
Adds user configurable tiling edge margins to PaperWM extension settings ui

jtaala:
Closes #364 

This PR adds user configurable tiling edge margins to the settings ui.  Edge margins are the minimum margin from screen edges to the leftmost and rightmost window edges.  Currently the edge margin is set to 0px (by design).

Edge margins were discussed in #364, with some users preferring having an edge margin.  This PR allows users to:

- select their preferred margins;
- preview the edge margins as the settings is changed;

### Video demonstration of changing edge margins:

https://user-images.githubusercontent.com/30424662/213923583-ffa15624-4ce0-419b-a31c-34dabf53959d.mp4


jtaala:
Hey @hedning, this was a while back but you responded regarding the design decisions re edge margins back in #364  - keen to get any thoughts re this PR which makes them user configurable.

jtaala:
@smichel17 - given our conversation in [#454](https://github.com/paperwm/PaperWM/pull/454) and [#456](https://github.com/paperwm/PaperWM/pull/456) re user settings - what do you think of this one?

I think this one definitely fits - as it's a margin. 

Also, interestingly I've noticed that it's one of the first things that seems "off" to new users, although it's intended behaviour to let users know when they are a tiling edge.  I prefer `5px` though - so I can see my nicely styled border but still see when at the tiling edge.

Anyways, we could leave this one there and move or remove the `Useful window widths/heights` settings to cleanup it up a bit (removing the widths/heights setting and users can use dconf to set those ones).

smichel17:
I think this is one we can fix without a setting. I'll reply in #364

jtaala:
> I think this is one we can fix without a setting. I'll reply in #364

Thanks @smichel17!  I don't think it's a bug though, it's a feature! :-).  I added a comment to #364 outlining why I chose to have it as a simple user-defined preference setting (which is fit nicely in the settings ui with the margin settings).

Lythenas:
I tested this and it seems to work without issues. I set the margin to 5px and I also have the other margins set to 5px. So there is no space between the borders and the edge of the screen.

jtaala:
As mentioned in #488, closing this in favour of #488.

Issue #425, opened 2022-10-13
install.sh does not automatically enable the extension


NicTanghe:
install
```
❯ ./install.sh
ln: failed to create symbolic link '/home/ntanghe/.local/share/gnome-shell/extensions/paperwm@hedning:matrix.org/paperwm@hedning:matrix.org': File exists

PaperWM runs best with some Gnome Shell settings changed:
 workspaces-only-on-primary off: Required for working multi-monitor support
 edge-tiling off: Natively tiled windows doesn't work in PaperWM
 attach-modal-dialogs off: Attached modal dialogs can cause visual glitching

Use recommended settings (generates a backup) [Y/n]: y
restore-gnome-shell-settings-2022-10-13.sh exists

Enable the extension [Y/n]? y
Trying to load and enable extension:
something went wrong:
(false, '')
Success
```

<details><summary>debug</summary>

```
❯ ./debug
Unit gnome-shell-x11.service could not be found.
Unit gnome-shell-wayland.service could not be found.
12:41:00: Running GNOME Shell (using mutter 42.5) as a Wayland display server
12:41:00: Device '/dev/dri/card0' prefers shadow buffer
12:41:00: Added device '/dev/dri/card0' (nvidia-drm) using non-atomic mode setting.
12:41:01: Created gbm renderer for '/dev/dri/card0'
12:41:01: Boot VGA GPU /dev/dri/card0 selected as primary
12:41:01: Disabling DMA buffer screen sharing (implicit modifiers not supported)
12:41:01: Using public X11 display :1024, (using :1025 for managed services)
12:41:01: Using Wayland display name 'wayland-0'
12:41:01: Unset XDG_SESSION_ID, getCurrentSessionProxy() called outside a user session. Asking logind directly.
12:41:01: Will monitor session c1
12:41:01: Extension apps-menu@gnome-shell-extensions.gcampax.github.com already installed in /usr/share/gnome-shell/extensions/apps-menu@gnome-shell-extensions.gcampax.github.com. /usr/share/gnome-shell/extensions/apps-menu@gnome-shell-extensions.gcampax.github.com will not be loaded
12:41:01: Extension background-logo@fedorahosted.org already installed in /usr/share/gnome-shell/extensions/background-logo@fedorahosted.org. /usr/share/gnome-shell/extensions/background-logo@fedorahosted.org will not be loaded
[...]

Issue #398, opened 2022-01-21
Install fails on OpenSUSE Tumbleweed


mtgcyoi:
Trying to install on OpenSUSE Tumbleweed fails:

`gareth@cube-server ~/PaperWM (master)> ./install.sh
ln: failed to create symbolic link '/home/gareth/.local/share/gnome-shell/extensions/paperwm@hedning:matrix.org': File exists

PaperWM runs best with some Gnome Shell settings changed:
 workspaces-only-on-primary off: Required for working multi-monitor support
 edge-tiling off: Natively tiled windows doesn't work in PaperWM
 attach-modal-dialogs off: Attached modal dialogs can cause visual glitching

Use recommended settings (generates a backup) [Y/n]: y
restore-gnome-shell-settings-2022-01-21.sh exists

Enable the extension [Y/n]? y
Trying to load and enable extension:
something went wrong:
(false, '')
Success`

paradajz:
just do `rm /home/gareth/.local/share/gnome-shell/extensions/paperwm@hedning:matrix.org` and try again

LovesTha:
I'm not terribly surprised that trying again gets the same result.

How do I get more information about what is going wrong?

LovesTha:
(ooops, I'm the OP, I just logged in to an account I rarely use, this is my normal account)

LovesTha:
No hints on steps to debug why the installer is failing?

gwenzek:
the installation also fails for me on Fedora. I don't even have an error message. The script just return with a failed exit code. Is this only supported on Ubuntu ? Why isn't it available on https://extensions.gnome.org/ ?

MerlinR:
Same issue on Debian 11.
```
Trying to load and enable extension:
something went wrong:
(false, '')
Success
```

jamestwebber:
I'm getting the same result (`(false, '')`) but I just updated to Ubuntu 22.04 and I'm installing with the `gnome-42` branch (gnome version 42.4).

mitchweaver:
Seeing this same issue on Fedora 36, reproducible on a clean install

azmeuk:
I can confirm this on Archlinux with GNOME 42.

aliaksandrsen:
I can confirm this on DEBIAN bookworm with GNOME 43.

savy-91:
I was able to install it today on Tumbleweed following these instructions https://github.com/paperwm/PaperWM/issues/425#issuecomment-1401509135

[...]

Issue #459, opened 2022-12-30
Unable to install on fedora 37 Gnome 43


ghost:
rying to load and enable extension:
Error: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name org.gnome.Shell was not provided by any .service files
something went wrong:

Success

jtaala:
Hey @ShabirK21, how are you trying to install it? (via a fedora repo) or by cloning this repo (if you're on Gnome 43 you'll need to be on the `develop` branch) and running the `install.sh` script?

ghost:
> Hey @ShabirK21, how are you trying to install it? (via a fedora repo) or by cloning this repo (if you're on Gnome 43 you'll need to be on the `develop` branch) and running the `install.sh` script?

Hi, I'm doing it by cloning this repo and running the script while in develop branch, also I'm using Gnome 43

jtaala:
Hmm, well, that error is only when trying to enable the extension via script (the install, or symlinking the extension to be picked up by gnome, is done prior).  Can you confirm if the extension actually got installed? (you'll need to restart gnome or login/logout, then check if it was installed by opening up the `extensions` app).

Note, either way it shouldn't have died like that, just want to see if there's a deeper issue and it didn't even install.

P.S. I'm not on, nor familiar with Fedora btw - so might be limited in debugging this one.

ghost:
> Hmm, well, that error is only when trying to enable the extension via script (the install, or symlinking the extension to be picked up by gnome, is done prior). Can you confirm if the extension actually got installed? (you'll need to restart gnome or login/logout, then check if it was installed by opening up the `extensions` app).
> 
> Note, either way it shouldn't have died like that, just want to see if there's a deeper issue and it didn't even install.
> 
> P.S. I'm not on, nor familiar with Fedora btw - so might be limited in debugging this one.

So i tried installing and restarted my device. It seems to have installed and is in available in extensions and works. Although i still got the error. Thanks

jtaala:
>...something went wrong:
> 
> Success

Let that be a lesson to everyone - things can (and will) go wrong in life... and you can still have success! :smiley:

jtaala:
I think this is a semi-known issue with the install script when it can't enable the extension (might even be if/when extensions are disabled)... it should prob be fixed (maybe if it errors out at extension enabling have it return something like `logout/login and enable it yourself... please`).

smichel17:
[...]

PR #472, opened 2023-01-29
Update install.sh error message

smichel17:
Hopefully this will help at least a little bit with confusion like https://github.com/paperwm/PaperWM/issues/398, explained in https://github.com/paperwm/PaperWM/issues/425#issuecomment-1284622580

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -77,8 +77,9 @@ if [[ "(true, 'true')" == "$RET" ]]; then
         gnome-shell-extension-tool --enable="$UUID"
     fi
 else
-    echo Something went wrong:
-    echo $RET | sed -e "s/(true, '\"//" | sed -e "s/\\\\n/\n/g"
-    
-    echo You may need to enable PaperWM manually
+    # String is already quoted and escaped; we want to print the unescaped version
+    unescaped="$(echo "$RET" | sed -e "s/(true, '\"//" | sed -e "s/\\\\n/\n/g")"
+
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -77,8 +77,8 @@ if [[ "(true, 'true')" == "$RET" ]]; then
         gnome-shell-extension-tool --enable="$UUID"
     fi
 else
-    echo something went wrong:
+    echo Something went wrong:
```
jtaala: Do we even need a line saying `something went wrong`?  We could replace it just with a echo saying "Could not enable the PaperWM extension automatically, you may need to enable PaperWM manually".

I think much of the confusion comes from seeing that "something went wrong" line and misinterpreting as PaperWM wasn't installed - rather than it was installed but couldn't be auto enabled for some reason.  The "something went wrong" line just seems to stick out.

That was my thought when I first installed PaperWM (I just read the "something went wrong" line, and even though it was followed by some JSON-esque string saying "false" I still thought "oh, crap, maybe paperwm is just broken") - and appears to be the thought that led to #459.

jtaala (commented):
Hey @smichel17, see comment left re the "something went wrong" line.

What do you think?

install.sh
smichel17: Yeah, that could work

smichel17: Updated, how's that?

jtaala (approved):
Looks good @smichel17!  I reckon that's a big improvement and should avoid confusion here.

Issue #344, opened 2020-10-23
Topbar turns transparent when activating PaperWM


vincentmathis:
![Before activating PaperWM](https://user-images.githubusercontent.com/33002126/96957587-92a32400-14ea-11eb-81e1-61319e18d23a.png)
![After activating PaperWM](https://user-images.githubusercontent.com/33002126/96957593-99ca3200-14ea-11eb-99b5-70667e1336ab.png)

hedning:
Yeah, this is intended (meant to show case the workspace color), but since transparency can be a bit tricky it isn't guaranteed to work well with all themes and backgrounds.

`Main.panel.actor.set_style('')` in your `user.js` enable function should probably be decent local fix.

Open to a PR to make it optional.

Thesola10:
@hedning I get it, but are you sure it's within the scope of PaperWM to make the topbar transparent? I'm having issues on launch with Dash to Panel's handling, and it can cause conflict with more extensions

hedning:
> I get it, but are you sure it's within the scope of PaperWM to make the topbar transparent?

Yeah, I can absolutly buy that argument, and I'll happily accept a PR that makes top bar transparency configurable :)

raiguard:
Bump. I would like the top bar to be more transparent than it currently is. It's strange that PaperWM modifies the top bar style at all, I don't see what you mean by "workspace color") when you can clearly see it around the borders of the windows already.

Issue #127, opened 2019-06-23
Make the default workspace color palette better


hedning:
Especially bad with the first workspace colored a somewhat dark blue which have a bad contrast against the selected window highlight.

PR #476, opened 2023-02-21
Show persistent viewport window position indicator in topbar (completes #471)

### LINKED ISSUE #471: Show an indication of viewport position when there are offscreen windows
The reason for not merging https://github.com/paperwm/PaperWM/pull/364 is that leaving the window border offscreen gives an indication of when you are at the "end" of the row of windows, without having to try and move to the next window in that direction. The goal of this issue is to find some alternative, so we can merge that PR.

An easy solution would be to make the side border of the outermost window a different color (e.g. black). Inelegant but gets the job done.

A more ambitious idea: make a portion of the top bar change color (or opacity?) based on which windows you're currently viewing. For example, if you have 4 full-width windows open, and the 3rd one is visible on screen right now, the top bar could look like this (using obnoxious colors to make it more obvious):

image

If you had 3 fullscreen windows open, with the center on screen, it would look like this:

image

If two windows, with the left one on screen:

image

…etc.

I'm not sure if this would be obnoxious or distracting in practice. It wouldn't necessarily need to affect the whole top bar, could be just a small amount along the top or bottom. An alternative would be to use some of the empty space in the top bar (to the left of the clock) to show a smaller-scale version of this, maybe using icons too, something like this (geary, gimp, and part of extensions would be visible in this example):

image

jtaala:
Fixes #471 

This PR adds a window position indicator bar as described in #471.

The bar style can be overriden with the css class `paperwm-window-position-bar`. 

I've also added a schema property to enable/disable showing the position bar.  
You can disable it from the command line with:
`dconf write /org/gnome/shell/extensions/paperwm/show-window-position-bar false` 
and enable it with 
`dconf write /org/gnome/shell/extensions/paperwm/show-window-position-bar true`

### See the window position bar (topbar blue coloured segment indicating window tiled position) in the video below:

https://user-images.githubusercontent.com/30424662/221416159-464d7512-5174-451b-9035-0ee84f9eb4ec.mp4

### Feedback welcome!
The indicator might not be wanted by some users - so, it might be worth adding a toggle in the paperwm preference ui.


Lythenas:
I think this looks very good.

But I have a comment about the naming: "Window indicator" is really generic and does not really tell you anything. So at the very least we should document it properly in the code (and probably readme) what exactly it is. Also I guess it looks effectively like a scroll bar, so maybe that is a better name for it?

Also maybe we need some discussion about the default color, but it is easily configurable so maybe not that important. The default color of the window selection comes from the `.tile-preview` class, which is defined in the gnome theme. Also if I remember correctly on one gnome version I was able to select a color in the gnome settings and it also changed the paperwm window selection color, but on gnome 43.2 I can't find that setting anymore. I guess what I'm getting at is: What should be the default color? The same as the window selection or another one?

jtaala:
Thanks @Lythenas ,

> But I have a comment about the naming: "Window indicator" is really generic and does not really tell you anything. So at the very least we should document it properly in the code (and probably readme) what exactly it is. Also I guess it looks effectively like a scroll bar, so maybe that is a better name for it?

I guess it does look like a scroll bar,  my only issue with referring to it as a  `scroll bar` is that that name has some meaning behind it  - e.g. users will likely try to click on it and drag it left/right (like s a scroll bar) .  What about something like `viewport position indicator`?  or maybe a mix: `viewport position bar`?  Actually that last one sounds pretty good(?).  Open to other names/ideas.

> Also maybe we need some discussion about the default color, but it is easily configurable so maybe not that important. The default color of the window selection comes from the `.tile-preview` class, which is defined in the gnome theme. Also if I remember correctly on one gnome version I was able to select a color in the gnome settings and it also changed the paperwm window selection color, but on gnome 43.2 I can't find that setting anymore. I guess what I'm getting at is: What should be the default color? The same as the window selection or another one?

I think using `tile-preview` style/colors as the default color is good here and thinking about it should have been the default I went with.  I just chose green because I wanted it to stand out a bit for showcasing what this thing is.  I'll change the default style and throw up another video tonight to see what it looks like.

Cheers!

jtaala:
Thoughts?  should the position bar be shown when there is only one window?  or only when there is at least two windows?

Showing with one window seems a bit much? (not too useful?).

jtaala:
Hey all!  I've been testing this in anger the last two weeks and has been working great.

I'm planning on merging this in one week, please post any comments or thoughts re this PR if you have any.

README.md
```diff
@@ -273,6 +273,26 @@ Keybindings.bindkey("<Super>j", "my-favorite-width",
...
+```
+dconf write /org/gnome/shell/extensions/paperwm/show-window-position-bar false
+```
+
+To enable the bar execute:
+```
+dconf write /org/gnome/shell/extensions/paperwm/show-window-position-bar true
+```
+
+You can modify and override the styles of both the coloured position bar and the dimmed "position bar backdrop" by overriding the `paperwm-window-position-bar` and `paperwm-window-position-bar-backdrop` CSS classes respectively (see `user.css` in [Development & user configuration](#development--user-configuration) section for more information).
```
Lythenas:
```suggestion
You can modify and override the styles of both the coloured position bar and the dimmed "position bar backdrop" by overriding the `paperwm-window-position-bar` and `paperwm-window-position-bar-backdrop` CSS classes respectively (see `user.css` in [Development & user configuration](#development--user-configuration) section for more information). The `paperwm-window-position-bar` will also inherit the selection color (same as window borders) from `tile-preview`.
```

Or something along those lines. Correct me if I'm wrong.

tiling.js
```diff
@@ -227,6 +243,10 @@ var Space = class Space extends Array {
             this.targetX = oldSpace.targetX;
         }
         this.cloneContainer.x = this.targetX;
+        
+        this.windowPositionBarBackdrop.width = this.monitor.width;
+        this.windowPositionBarBackdrop.height = TopBar.panelBox.height;
```
Lythenas, line 244: What happens if the monitor is "resized"? E.g. unplugged and another one plugged in, or the workspace moves to the laptop screen. Also what happens if I open the workspace on a different monitor than before?

I will check if this is a problem and then comment again.

tiling.js
```diff
@@ -1121,6 +1165,32 @@ border-radius: ${borderWidth}px;
...
+        // number of columns (a column have one or more windows)
+        let cols = this.length;
+        if (cols <= 1) {
+            this.windowPositionBar.hide();
+            return;
+        } else {
+            this.windowPositionBar.show();
+        }
+        
+        let width = this.monitor.width;
```
Lythenas, line 1255:
```suggestion
        let width = this.monitor.width;
        this.windowPositionBarBackdrop.width = this.monitor.width;
```

Also update the backdrop width here. To be safe.

topbar.js
```diff
@@ -532,7 +530,7 @@ function enable () {
             return;
         fixTopBar();
     });
-
+
```
Lythenas:
```suggestion

```

Whitespace error.

Lythenas (changes requested):
Overall I think the code looks good.

I will try this out for the next couple of days, before approving.

README.md
Lythenas: Also maybe add a (cropped) screenshot of the topbar. Nothing complicated. Just to make it trivial for the user to see what this feature is about.

tiling.js
jtaala: oohh - good question!  I'll see if there's a function of signal on monitor or resolution change and hook into that to set the backdrop width/height on that too.

I note that your suggestion below would help with that, in that the backdrop width would be set (even on a new monitor of diff size) when switching to a different window.

tiling.js
jtaala: I'll see if there's a better way to implement this (e.g. set width of backdrop on monitor.width change) - and/or can add this.  Actually I'll add this now and then see if it's better to resize backdrop on monitor/resolution change.

README.md
jtaala: Ah, I was actually going to do this.  I note though that the [media folder](https://github.com/paperwm/media) where README.md is actually in another repo that we don't have write access to.

I was thinking on creating a media folder in PaperWM (our ) repo and move those files over and update README.md and then add a screenshot.

topbar.js
jtaala: Implemented in [4d0eab4](https://github.com/paperwm/PaperWM/pull/476/commits/4d0eab4d073588bcaba8a434e3eb88c9866541af).

tiling.js
jtaala: Implemented in [4d0eab4](https://github.com/paperwm/PaperWM/pull/476/commits/4d0eab4d073588bcaba8a434e3eb88c9866541af).

tiling.js
jtaala: I implemented your suggested changes and did some light testing - so I change resolution to be smaller and then larger (hoping to see if the backdrop width was set for the smaller size) - it seemed to not be an issue (backdrop was correctly sized on resolution changes).

Will test with multiple monitors as well.

Did some testing - on multiple monitors the indicator and backdrop still show even if don't have a topbar... I think it'll be better to have it hide (backdrop/indicator) if no topbar is present on a space.

Lythenas:
On my work computer right now (running Gnome 42.5, if that matters) and I'm currently using dash-to-panel (don't remember exactly why that extension, probably just so I have the topbar on all monitors). In general it seems to work. But two small issues:

- The "Show Applications" button is misaligned. But only on the primary monitor. And only after switching focus to secondary monitor and back. (could be cause by dash-to-panel)
- The workspace name is now shown twice on the primary monitor and the name is now also shown on the secondary monitor (which I like) but is badly aligned.
- Also when switching workspaces (`Super + PageUp/Down`) the workspace name is also misaligned (similar to the second monitor) and there is a gray bar at the top (maybe the backdrop).

image

image

Regarding dash-to-panel:

I also had to manually set it transparent in its settings. But that is probably fine.

As far as I know gnome-shell has no generic topbar API. I'm not sure we should specifically support dash-to-panel. But I think we should either:

1. provide a way to show the topbar on all monitors (maybe that is now also possible with vanilla gnome-shell)
2. support at least one extension that does that.

I've also been working on a PR that would show the "Workspace X" menu on all topbars and I didn't find a way to generically get all topbar. I.e. I had to access `global.dashToPanel` which is set by that specific extension.

I don't have any particular reason to use dash-to-panel over any other extension. So I'm happy to switch to another. But we can also discuss this later.

jtaala:
> * The workspace name is now shown twice on the primary monitor and the name is now also shown on the secondary monitor (which I like) but is badly aligned.
> * Also when switching workspaces (`Super + PageUp/Down`) the workspace name is also misaligned (similar to the second monitor) and there is a gray bar at the top (maybe the backdrop).

Yes, PaperWM actually has two workspace names (one on the topbar and one on each space) - it just so happens that they align so it's harder to notice.  I never liked that, for example, the icon for #482 will hide when not in the workspace switching view.  I'll clean up the old workspace label code and get it to do something similar.

jtaala:
> As far as I know gnome-shell has no generic topbar API. I'm not sure we should specifically support dash-to-panel. But I think we should either:
> 
> 1. provide a way to show the topbar on all monitors (maybe that is now also possible with vanilla gnome-shell)
> 2. support at least one extension that does that.

~~We should create a new PR for this.  It could be quite a bit of work(?).  I don't use dash-to-panel and am not sure that PaperWM should change too much to work better with a particular extension (within reason of course).  Perhaps you could work on a PR for this one - I note that there was is a PR #420 that you could look at extending.~~

Please check out my latest commit - I've added initial multimonitor support for the position bar indicator (.e.g it shows on all monitors) - including with the `Make topbar follow monitor focus`.

jtaala:
> * The "Show Applications" button is misaligned. But only on the primary monitor. And only after switching focus to secondary monitor and back. (could be cause by dash-to-panel)

PaperWM doesn't have a "Show Applications" button does it?  Is this something that dash-to-panel adds?

jtaala:
> 1. provide a way to show the topbar on all monitors (maybe that is now also possible with vanilla gnome-shell)

Actually, I really like this.  I've been playing around with multi-monitors the current experience isn't great (with or without the indicator).  I can implement showing the window position indicator and backdrop on all monitors - makes everything much more consistent and then you can still see the window position bar for all monitors.

Plus, enabling `Make topbar follow monitor focus` enabled works much nicer.

jtaala:
~~NOTE: I'm still working through some bugs with the `Make topbar follow monitor focus` option... grrrr~~ :heavy_check_mark:

tiling.js
jtaala: Thanks @Lythenas, I believe I've now resolved these issues - resize/changing monitors now works as exepected (in my tests), and I've now added multi-monitor support (with the window position indicator now showing correctly across all monitors).

Also the two workspace names should be resolved too.

Please test a bit when you can!

Lythenas:
If I have a fullscreen application (e.g. a fullscreen youtube video, or just firefox with F11) the window indicator is not shown (or I guess it is below the window), which is good. But if I then focus a different monitor. The window indicator is shown above the fullscreen window. If I focus the fullscreen window again, it corrects itself (after a short delay). This also happens if `show-window-position-bar` is disabled (but just with the backdrop).

Also with your multi monitor changes the backdrop is also shown if I disable `show-window-position-bar`. I think we should (as far as possible) keep the old behavior if the feature is disabled.

And it seems like disabling the feature does not return the original opacity of the topbar.

With dash-to-panel there are still some issues (actually even more after the multi monitor changes^^). But I will try to think of something independent of this PR.

Lythenas:
Also I think we should decide if this will now be the new default behavior or if it will just be an addon feature.

In any case I think we should also add a setting for this in the UI. But that can be a separate PR when we know there are no more bugs.

Lythenas:
Not sure if this was broken by this PR, but the workspace border/header when navigating is different that it used to be:

Now:
Screencast from 2023-03-08 20-57-53.webm

Previous (from README):

jtaala:
> Not sure if this was broken by this PR, but the workspace border/header when navigating is different that it used to be:
> 
> Now: Screencast from 2023-03-08 20-57-53.webm
> 
> Previous (from README): 

Ah ha yes!  that was by design (shows the backdrop and indicator just like normal view).  My thinking:
- I wanted to have the workspace previews (when switching) look like the actual view;
- it's easier (for me at least) to see the workspace name, and much easier to see the "focus mode" icon in [#482](https://github.com/paperwm/PaperWM/pull/482), regardless of what workspace colour or background you use (e.g. light colours or light background - I prefer seeing my background).  I note issue #127 re workspace colours and some visual issues it can bring - this choice was to avoid/minimise this.  E.g. make the workspace name (and other stuff like PR [#482](https://github.com/paperwm/PaperWM/pull/482)) easy to see regardless of workspace colours);
- you can actually see and use the workspace indicator from switching view (give it a try with switching windows with shortcut keys from the workspace switching view.  It then looks the same as in normal views (albeit a zoomed out view) ;
- looks betterer! (although, this could just be an opinion of mine);

jtaala:
> If I have a fullscreen application (e.g. a fullscreen youtube video, or just firefox with F11) the window indicator is not shown (or I guess it is below the window), which is good. But if I then focus a different monitor. The window indicator is shown above the fullscreen window. If I focus the fullscreen window again, it corrects itself (after a short delay). This also happens if `show-window-position-bar` is disabled (but just with the backdrop).
> 

Good pickup!  Will fix that.

> Also with your multi monitor changes the backdrop is also shown if I disable `show-window-position-bar`. I think we should (as far as possible) keep the old behavior if the feature is disabled.

Yeah, sounds good - I originally thought to just leave the backdrop there but what you mentioned is probably a better choice (return things to how they were).

> And it seems like disabling the feature does not return the original opacity of the topbar.

Original opacity - you mean like being opaque?  Interestingly, PaperWM TopBar.js sets the opacity to be slightly transparent (assuming it will stay like that) but gnome can clobber it (e.g. without this PR, right after a login, try going into overview and then clicking a window - gnome will set the topbar opaque).  This PR fixes that long-standing issue (although it might not be what people are used to now?).

jtaala:
> If I have a fullscreen application (e.g. a fullscreen youtube video, or just firefox with F11) the window indicator is not shown (or I guess it is below the window), which is good. But if I then focus a different monitor. The window indicator is shown above the fullscreen window. If I focus the fullscreen window again, it corrects itself (after a short delay). This also happens if `show-window-position-bar` is disabled (but just with the backdrop).

Fixed this.

jtaala:
> Also with your multi monitor changes the backdrop is also shown if I disable `show-window-position-bar`. I think we should (as far as possible) keep the old behavior if the feature is disabled.

Have implemented this.  Now, disabling the window position bar will completely disable it and revert back to previous behaviour.

Although, the issue with the topbar opacity (starting out as transparent but then gnome cloberring it to opaque) has been fixed (as in, PaperWM will enforce that transparency as it was originally designed to).  I'm remiss to revert back to the broken behaviour  - if we want the topbar to be opaque like vanilla gnome then we should just remove that transparency code.  I do note that  #344 suggests NOT having PaperWM change transparency.  

Now, we use transparency to good effect with the window position bar (and I like it! lol) but, I'm open to suggestions re this, for example we could:

1. keep transparency and fix the code (current behaviour with this PR);
2. do away with transparent topbar and let users add that if wanted (via css styling in PaperWM or other extensions etc.);
3. hybrid approach: if window position bar is `disabled` then we could have an opaque topbar (vanilla gnome) and keep transparency for when window position bar is `enabled`;

Option 2 would simplify things I guess.  If not, will go ahead with the first option (since that's what we've got atm).

jtaala:
@Lythenas, fixed those issues you raised, could you give the latest a test when you can?  Cheers!

Lythenas (approved):
This looks good to me. I haven't noticed any more issues.

Thanks for the awesome feature.

jtaala:
> This looks good to me. I haven't noticed any more issues.
> 
> Thanks for the awesome feature.

Thanks @Lythenas!  I've just cleaned up some stuff and also I did end up adding the enable/disable toggle to the settings ui.  I believe this feature is a big enough, prominent enough, and changes enough of the previous behaviour  to justify adding it to the settings ui.

I also updated the README.md and added the video from this PR to it.

I'll merge this shortly.

PR #478, opened 2023-02-25
Check gnome version compatibility

### LINKED ISSUE #475: I could use better errors 🙈
I've updated PaperWM with `git pull` after more than a year, and the extension fails on startup.

All I see if a little one-line floaty thing from gnome shell.
If I click that I get a more detailed error, which still doesn't quite help me:

```
-En PaperWM
Error occured in tiling @init:

fields are not currently supported

safeCall@/home/<user>/.local/share/gnome-shell/extensions/paperwm@hedning:matrix.org/extension.js:60:22
run@/home/<user>/.local/share/gnome-shell/extensions/paperwm@hedning:matrix.org/extension.js:46:14
init@/home/<user>/.local/share/gnome-shell/extensions/paperwm@hedning:matrix.org/extension.js:100:9
_callExtensionInit@resource:///org/gnome/shell/ui/extensionSystem.js:435:50
loadExtension@resource:///org/gnome/shell/ui/extensionSystem.js:348:27
_loadExtensions/<@resource:///org/gnome/shell/ui/extensionSystem.js:601:18
collectFromDatadirs@resource:///org/gnome/shell/misc/fileUtils.js:27:17
_loadExtensions@resource:///org/gnome/shell/ui/extensionSystem.js:576:19
_enableAllExtensions@resource:///org/gnome/shell/ui/extensionSystem.js:610:18
_sessionUpdated@resource:///org/gnome/shell/ui/extensionSystem.js:641:18
init@resource:///org/gnome/shell/ui/extensionSystem.js:56:14
_initializeUI@resource:///org/gnome/shell/ui/main.js:257:22
start@resource:///org/gnome/shell/ui/main.js:146:5
@<main>:1:47
```

Versions:
* paper wm: f590f8b
* gnoem shell: whatever I get with ubuntu 20.04.x

Lythenas:
In `init` check the gnome shell compatibility (as specified in the manifest) and display a warning if it is not compatible.

This is useful/necessary because if you git pull PaperWM after you installed it in a compatible version, you might update to an incompatible version. E.g. I imagine the develop branch will eventually move to the next gnome version and that gnome version could be incompatible (like the branches for the other gnome versions).

Screenshot from 2023-02-26 19-25-05

(Note: I hardcoded the version string to get the screenshots)

The notification reuses the existing `errorNotification` so it will show an editor with the message when clicked.

Questions:
- Should there be a way to disable the warning?
- The notification disappears relatively quickly. But I don't see a way to make it stay.

Fixes #475

### DIFF
--- a/extension.js
+++ b/extension.js
@@ -90,7 +90,7 @@ function init() {
     Extension = imports.misc.extensionUtils.getCurrentExtension();
     convenience = Extension.imports.convenience;
 
-    checkGnomeShellVersionCompatibility();
+    warnAboutGnomeShellVersionCompatibility();
 
     if(initRun) {
         log(`#startup Reinitialized against our will! Skip adding bindings again to not cause trouble.`);
@@ -131,7 +131,9 @@ var GLib = imports.gi.GLib;
 var Main = imports.ui.main;
 var Config = imports.misc.config;
 
-function checkGnomeShellVersionCompatibility() {
+// Checks gnome shell version compatibility and warns the user when running on
+// and unsupported version.
+function warnAboutGnomeShellVersionCompatibility() {
     const gnomeShellVersion = Config.PACKAGE_VERSION;
     const supportedVersions = Extension.metadata["shell-version"];
     for (const version of supportedVersions) {
@@ -142,11 +144,16 @@ function checkGnomeShellVersionCompatibility() {
 
     // did not find a supported version
     print("#paperwm", `WARNING: Running on unsupported version of gnome shell (${gnomeShellVersion})`);
-    print("#paperwm", `WARNING: Supported versions: ${supportedVersions}`);
+    print("#paperwm", `Supported versions: ${supportedVersions}`);
     const msg = `Running on unsupported version of gnome shell (${gnomeShellVersion}).
 Supported versions: ${supportedVersions}.
-Please upgrade/downgrade your gnome shell to a supported version or upgrade/downgrade PaperWM.`;
-    errorNotification("PaperWM warning", msg);
+Click for more information.`;
+
+    const notification = notify("PaperWM Warning", msg);
+    notification.connect('activated', () => {
+        imports.misc.util.spawn(["xdg-open", "https://github.com/paperwm/PaperWM/wiki/Warning:-Running-on-unsupported-version-of-gnome-shell"]);
+        notification.destroy();
+    });
 }
 
 function getConfigDir() {
--- a/extension.js
+++ b/extension.js
@@ -90,6 +90,8 @@ function init() {
     Extension = imports.misc.extensionUtils.getCurrentExtension();
     convenience = Extension.imports.convenience;
 
+    checkGnomeShellVersionCompatibility();
+
     if(initRun) {
         log(`#startup Reinitialized against our will! Skip adding bindings again to not cause trouble.`);
         return;
@@ -127,6 +129,25 @@ function disable() {
 var Gio = imports.gi.Gio;
 var GLib = imports.gi.GLib;
 var Main = imports.ui.main;
+var Config = imports.misc.config;
+
+function checkGnomeShellVersionCompatibility() {
+    const gnomeShellVersion = Config.PACKAGE_VERSION;
+    const supportedVersions = Extension.metadata["shell-version"];
+    for (const version of supportedVersions) {
+        if (gnomeShellVersion.startsWith(version)) {
+            return;
+        }
+    }
+
+    // did not find a supported version
+    print("#paperwm", `WARNING: Running on unsupported version of gnome shell (${gnomeShellVersion})`);
+    print("#paperwm", `WARNING: Supported versions: ${supportedVersions}`);
+    const msg = `Running on unsupported version of gnome shell (${gnomeShellVersion}).
+Supported versions: ${supportedVersions}.
+Please upgrade/downgrade your gnome shell to a supported version or upgrade/downgrade PaperWM.`;
+    errorNotification("PaperWM warning", msg);
+}
 
 function getConfigDir() {
     return Gio.file_new_for_path(GLib.get_user_config_dir() + '/paperwm');


Lythenas:
@dimaqq can you check if this works for you?

dimaqq:
If I run PaperWM from this branch, I get 3 notifications:
1. Running on unsupported version of gnome shell (x,y)
2. Error blah blah tiling
3. Error blah blah tiling again (a bit different text)

So on one hand, this change kindof works, in that it works the user.
On the other, the code seems to fall through and other errors are shown.

I'm not sure if the latter is intended.

Lythenas:
So I intended it as only a warning and still load PaperWM like normal. Because I wasn't sure if the version check is always exhaustive. E.g. the next major gnome version might work just fine with PaperWM.

But we could also just abort loading if the version does not match. What do the others think?

jtaala:
> So I intended it as only a warning and still load PaperWM like normal. Because I wasn't sure if the version check is always exhaustive. E.g. the next major gnome version might work just fine with PaperWM.
> 
> But we could also just abort loading if the version does not match. What do the others think?

Yes, I agree with first sentiment - e.g. version 44 should (hopefully!) work just fine; also if it does have some issues then crowd-sourcing identification of those issues is usually a good thing  (i.e. allow the version to run so users can identify and submit issues etc.)

Also, I'm not sure about the `Please update/downgrade` suggestion in the notification - I don't mind it, it's just that in the incoming next gnome version (44) for example, say it works just fine (hey, I'm an optimist!) - then showing the `Please update/downgrade....` part might just confuse the user (?).

Also, there's the older branches (e.g. `gnome-3.38`), although those aren't be developed anymore (AFAIK) and don't have latest fixes/features etc.  We could put a message on there saying if you're using an older/unsupported version try that branch, and for newer versions we could ask them to please submit any issues found (so we can fix them etc.).

Just some ideas to consider.

Otherwise, I think this is good.  Happy to review it when you're ready so it can be merged.

Lythenas:
Yes you are right, that the message is a bit confusing. The space is a bit constrained in the notification. And once it initially hides (to the message tray) the user can't see the full message anyway (or at least I cant').

I will create a wiki page with a better explanation and link to that when clicking on the notification.

Lythenas:
I created a new wiki page https://github.com/paperwm/PaperWM/wiki/Warning:-Running-on-unsupported-version-of-gnome-shell and the notification will open this page when clicked.

Tell me if the wiki page should rather be in the readme.

Also should I add a try-catch around my function? All other code is run using safeCall, presumably to cause less problems with the gnome shell.

smichel17:
I'm not 100% sure when `init` is called vs `enable`, but generally I would suggest trying to move as much initialization code to `enable` as possible, so disabling and then re-enabling PaperWM results in a clean slate, similar to restarting gnome-shell. This would also help during development, especially on Wayland, to be able to test some changes without re-logging in gnome. Obviously it won't be perfect, but I think it can likely be better.

jtaala:
> I'm not 100% sure when `init` is called vs `enable`, but generally I would suggest trying to move as much initialization code to `enable` as possible, so disabling and then re-enabling PaperWM results in a clean slate, similar to restarting gnome-shell. This would also help during development, especially on Wayland, to be able to test some changes without re-logging in gnome. Obviously it won't be perfect, but I think it can likely be better.

Yes, so `init()` is run when the extension is actually loaded (i.e. when logging into a _new_ gnome session) - ~~it's also executed regardless of whether an extension is enabled or not (e.g. even for disabled extensions when you login).~~  _I was wrong here!, it is **NOT** run for disabled extensions on gnome login! (just did some testing here)._   It's run only _once_ per gnome session.

[GJS guidelines](https://gjs.guide/extensions/review-guidelines/review-guidelines.html#basics) state that `init`should only be used for "Gettext translations and static resources" which paperwm doesn't follow.  In the guideline's words:

>init() is called by GNOME Shell when your extension is loaded, not when it is enabled. Your extension MUST NOT create any objects, connect any signals, add any main loop sources or modify GNOME Shell here.

`enable` is run when an extension is enabled (i.e. turned on, e.g. after a disable) - and can be automatically run (by gnome-shell) during gnome locking/unlocking the screen.

@smichel17  - paperwm doesn't quite align with guidelines here and it does cause certain constraints (e.g. in order to load `user.css` ~~or `user.js`~~ _(other complexities with `user.js` since it's a module and hence loaded statically... so for that would need another approach)_ you actually need to run `init` which means a full logout/login for wayland and a gnome-restart for X11 - instead of the more preferable disable/enable).  

It's a separate issue (to this PR) but  I think we should probably create an issue for not aligning with guidelines and to fix this... although it would require a decent refactor of current code (I'm happy to give this a go since it annoys me that in wayland I have to logout/login and lose application states just to reload user styles etc.)

jtaala:
In any case, I'd suggest considering leaving this check in `init` - it's not adding/changing gnome-shell or connecting signals etc. (as per guidelines).

We could move it to `enable` but then users could  get spammed with that notification when enabling/disabling other extensions and/or locking/unlocking their session (unless you put a var/check on whether the notification has already been shown for a session - which adds complexity and just does what putting it in init would do in this case).

jtaala:
> I created a new wiki page https://github.com/paperwm/PaperWM/wiki/Warning:-Running-on-unsupported-version-of-gnome-shell and the notification will open this page when clicked.
> 
> Tell me if the wiki page should rather be in the readme.
> 
> Also should I add a try-catch around my function? All other code is run using safeCall, presumably to cause less problems with the gnome shell.

Nice work - I like this!  I think it's fine - shouldn't need a try/catch there (it's not doing anything too crazy) - I'll do some testing a bit today.  Re `safeCall`, not quite, it's specifically for running methods if they are present in other modules (e.g. run the `enable` or `disable` methods in all loaded modules if they are present in that module).

Lythenas:
As per jtaala's explanation. That was also mainly my reasoning for putting it in `init`. Although I didn't know init also runs when the extension is disabled.

Also regarding reloading user.css I had some code that I was playing around with that worked but I can't find it anymore... Maybe I will try to recreate it this week. It was relatively straight forward.

jtaala:
> As per jtaala's explanation. That was also mainly my reasoning for putting it in `init`. Although I didn't know init also runs when the extension is disabled.

**My apologies here!**  I was wrong re the `init` method being run for disabled extensions!  Just did some testing and re-looked at the doco (I misinterpreted the doco).  So, `init` is indeed only run once per gnome session.  From my testing it seems these are the cases when `init` is run:

- extension **IS** enabled and you log into a new gnome session --> `init` executed;
- extension **IS NOT** enabled when logging into gnome session **AND** you enable it (once logged in to gnome session) --> `init` is executed;

Otherwise, that would have caused issues for us since the version notification would have shown regardless of whether the extension was enabled or not (thankfully that's not the case!).

> Also regarding reloading user.css I had some code that I was playing around with that worked but I can't find it anymore... Maybe I will try to recreate it this week. It was relatively straight forward.

Thanks @Lythenas!

Yeah, thinking about it, `user.css` would be a good candidate for moving to `enable` (and not only just to align with GJS guidelines) since it's a stylesheet and could load it into the theme context on `enable` (in `extensions.js`).  Since there's already a check to ensure `enable` is only run (in `extensions.js`) if a `disable` was run previously then it should be fine (i.e. don't need to worry about css precedence/specificity with multiple changes being loaded) - maybe.  Anyway, you might have had another/better way!

jtaala (approved):

PR #482, opened 2023-03-04
Persistent centered mode (aka "Sticky Centering")

### LINKED ISSUE #299: Always Center
The more I use PaperWM, the more I play around with certain applications behind slightly different widths.

It would be really beneficial if there was an option to always center on the active window

### LINKED ISSUE #332: Option to center a window in focus.
New option in settings. If the Windows are centered immediately, it will make it much easier to navigate with the mouse on two monitors.
  
Now it is very difficult to aim the mouse at the very edge of the screen to switch to the next window without accidentally clicking on the second monitor.

jtaala:
Fixes #332, #299.

So, this PR resurrects and updates the `sticky-center` branch.  This basically continues to center windows (when switching to other windows) if the window was already centered (e.g. use the shortcut to center a window horizontally --> when switching to other windows they will be centered too).

~~The updates I added disables this behaviour when using the center shortcut again (on a window that is already centered) - in which case it will "uncenter" that window (move it back to the left edge).~~

_See the video below - in the video I switch between "focus" mode and normal mode (see leftmost topbar icon, which is clickable to switch between modes btw).  I'm clicking here but you can use a shortcut to switch between modes too:_

https://user-images.githubusercontent.com/30424662/223440683-e4cb3537-3f3b-49ac-ab0f-37ee12224a20.mp4

First up, I really like this behaviour when I'm more just focusing on one window/task at a time (instead of looking at multiple side-by-side windows).  However, ~~I'm setting this as a draft since I'm not entirely happy with it atm~~ - it's a departure from current behaviour which is to center the current window but not enforce centering view when switching, and will likely confuse users when they try it.

I would suggest:

- ~~having a separate shortcut to disable/enable a "focus center mode";~~ :heavy_check_mark: implemented
- ~~perhaps add an indicator icon (maybe at left-most position in the topbar) showing what mode PaperWM is in (e.g. centering mode or normal mode).  Or just have it show up when in centering mode.~~ :heavy_check_mark: implemented

Feedback please!


Lythenas:
> I would suggest:
> 
>     having a separate shortcut to disable/enable a "focus center mode";
>     perhaps add an indicator icon (maybe at left-most position in the topbar) showing what mode PaperWM is in (e.g. centering mode or normal mode). Or just have it show up when in centering mode.

I would definitively add your suggestions since it is quite different behavior between "center it one time" and "always center the currently focused window". Also maybe add an option to enable it by default, for people that want that.

Since this is also more like a "mode" now (that you can turn on/off). I think the the check `isCentered(last, space)` does not make that much sense. I think the "centering mode" state makes more sense to have as a variable in space or maybe globally.

jtaala:
Hey all, so I've implemented a nice(?) icon/indicator in the very left of the toolbar.  You can click it now and it will switch between the modes (normal and centered).

Give it a go and see what you think.  I'm liking it thus far.

jtaala:
I also added the shortcut `super`+`Shift`+`c` to toggle modes.

Lythenas:
I will start testing this for the next couple of days.

A couple of things I have noticed on first try:

- Is the icon on the secondary monitor supposed to be clickable? Because it isn't.  If not it's probably fine, the workspace name is also not clickable.
- Do we want to un-center a window if we disable sticky centering?
- Can we add a tooltip to the icon/button? E.g. say "Sticky center" or "Normal mode".

jtaala:
Thanks @Lythenas for testing, always apprecaited!  I really like the behaviour of this and use it quite a bit now.

> * Is the icon on the secondary monitor supposed to be clickable? Because it isn't.  If not it's probably fine, the workspace name is also not clickable.

~~Ah, yeah - so they're not actually real buttons - basically they're just for display.  If you switch on `Make topbar follow monitor focus` then they would work - since they would be the actual buttons on the topbar. Having said that - there's really nothing stopping us from making them "real" - will have a look at it.~~ _Icons on other monitors are now clickable_ :heavy_check_mark: 

_The shortcut to toggle modes will work though on other monitors though_.

> Do we want to un-center a window if we disable sticky centering?

~~I chose not too do that - it's gets a bit difficult since we'd likely want to track which direction is "centered" from - and that adds a little more complexity.  I suppose we could store a variable for `x` position before centering and then return the selected window back to that position on uncentering?  _will give this a go and see how it works - cheers!_~~ _un-centering now implemented_ :heavy_check_mark: 

> Can we add a tooltip to the icon/button? E.g. say "Sticky center" or "Normal mode".

Likely could - I haven't had the best luck with tooltips (I don't like the way it centers text - but that's mostly for multi line text).  Would happily accept a PR on that branch if anyone wants to give it a go.  _I'll also give this a go and shout out if run into trouble on it._

Lythenas:
> > Can we add a tooltip to the icon/button? E.g. say "Sticky center" or "Normal mode".
>
> Likely could - I haven't had the best luck with tooltips (I don't like the way it centers text - but that's mostly for multi line text). Would happily accept a PR on that branch if anyone wants to give it a go. _I'll also give this a go and shout out if run into trouble on it._

If it is hard to implement, it might not be worth it. I just thought the icon is probably not obvious.

> > Do we want to un-center a window if we disable sticky centering?
> 
> I chose not too do that - it's gets a bit difficult since we'd likely want to track which direction is "centered" from - and that adds a little more complexity. I suppose we could store a variable for `x` position before centering and then return the selected window back to that position on uncentering? _will give this a go and see how it works - cheers!_

I was thinking more like: If there is now empty space on the right/left when uncentering, move the window to the edge of the screen. I don't know how well remembering the original x position works. Because windows could be added/remove/moved since enabling the mode.

But also you don't have to spent that much time adding new features here. I was just posting what came to mind.

jtaala:
> But also you don't have to spent that much time adding new features here. I was just posting what came to mind.

No, this is appreciated - and I think it's a good idea to see what feels better here.  I'll try the `x` approach (since it's easy) and push it up to see how it feels - note, I'll just literally save the `x` position and then return whatever window is selected at the time (of undoing focus mode) to that position - might end up feeling weird but will see.

paradajz:
I thought I would really like this but quickly ended up just using the "normal" behavior. The feature works fine, but I don't like the fact that this icon is the first icon on the topbar instead of the workspace name like it is now - by now it's in my muscle memory to quickly drag the mouse all the way to the top-left position where the workspace name is and click it, which brings out the overview (when I don't want to use keyboard). This breaks that.

jtaala:
> * Do we want to un-center a window if we disable sticky centering?

Hey @Lythenas, hope the weekend's going well!  So, I've implemented this in the latest commits to this PR.  It's using a both approaches.  The current implementation basically remembers the direction it centered from (e.g. if original centering moved window left-to-right, then uncentering moves window back right-to-left) .  Now, this is for windows that are not the first or last in the space - for those, we'll move them to the closest edge (i.e. fill empty space).

This feels good to me, and pretty natural - it's a bit heuristic but I'm liking it so far - seems natural enough (without doing crazy things like tracking the `x` positions of each window when some other window was centered.

jtaala:
> I thought I would really like this but quickly ended up just using the "normal" behavior. The feature works fine, but I don't like the fact that this icon is the first icon on the topbar instead of the workspace name like it is now - by now it's in my muscle memory to quickly drag the mouse all the way to the top-left position where the workspace name is and click it, which brings out the overview (when I don't want to use keyboard). This breaks that.

Hey @paradajz,  ha... didn't think of that.  Well, a couple of options then - either allow disabling this feature (with a `dconf` settable setting), or if you like the feature but just not the icon placement we could place it _after_ the workspace name?

paradajz:
> or if you like the feature but just not the icon placement we could place it after the workspace name

Yeah I have nothing against the feature, in fact, I might use it from time to time, I would just prefer then icon to be after workspace name.

jtaala:
> Yeah I have nothing against the feature, in fact, I might use it from time to time, I would just prefer then icon to be after workspace name.

Fair enough mate!  I've now moved the focus mode icon to after the workspace name.  Give that a go and see how you like it.

jtaala:
> * Is the icon on the secondary monitor supposed to be clickable? Because it isn't.  If not it's probably fine, the workspace name is also not clickable.

Hey @Lythenas, the workspace name and focus-mode-icon on other monitors are now clickable.  Cheers.

jtaala:
@paradajz, I also added a setting to turn off the new focus mode icon for those that don't like it for whatever reason.  E.g. run:

`dconf write /org/gnome/shell/extensions/paperwm/show-focus-mode-icon false`

Lythenas:
I think this works very well.

Just one thing I noticed: The icon is different on Gnome 42 than on Gnome 43.
image

I don't care that much about the icon, but on Gnome 42 it definitively doesn't mean "centering"...

Is there anything we can do about that? Maybe ship our own icons? (could just be a copy of the gnome icons)

jtaala:
> The icon is different on Gnome 42 than on Gnome 43.

Oh that's pretty interesting.  So, I'm using the `preferences-desktop-multitasking-symbolic` icon for centering and the `sidebar-show-right-symbolic` for normal - yeah, I guess other themes versions may have a different icon for that?

> Is there anything we can do about that? Maybe ship our own icons? (could just be a copy of the gnome icons)

Sounds like might need to - ~~will have a look at the best way to do that.~~ **have now implemented this** :heavy_check_mark: 

Thanks @Lythenas!  Yeah, I'm quite enjoying it - I find myself using it fairly frequently on workspaces with quite disparate tasks (e.g. music, browsing etc.).

jtaala:
@Lythenas, can you test the latest (on Gnome 42 as well)?  I'm now packaging the icons in PaperWM so should show icons correctly regardless of gnome version or theme.

terlar:
The icons was showing correctly for me on gnome 42, also before this change. So probably some theming issue as suggested

jtaala:
> * Can we add a tooltip to the icon/button? E.g. say "Sticky center" or "Normal mode".

Implemented! (well, my version of a tooltip... I forgot how not-great the gtk tooltips are - had trouble with them so just created an implementation myself that I think fits better).

Let me know what you think (I wasn't sure what to put in the tooltip - maybe it's a bit verbose?).

I think we're getting close to completing this feature.

Lythenas:
One slight issue with the tooltip: It appears on the wrong monitor :D My topbar is on the right monitor and the tooltip shows on the left monitor. But where I would expect it if I hover over the icon on that monitor. It also doesn't show up if I hover over the icon on my secondary monitor (which is probably also fine, we can improve the behavior if/when we decide how to do topbars on a seconary monitor).

Also maybe make the tooltip background a bit more opaque. It was a little hard to read with my pinned firefox tabs in the background.

Regarding the wrong icons: It could very well be a theming issue (although I think I didn't change anything). But since we are not using the icons for their original purpose I think it is good to bundle them with PaperWM. Just one note on that: Are we allowed to just copy them? What is the license on them? Maybe we need to include a notice somewhere. I would maybe put it either in a comment in the svgs or in a separate file next to the icons.

jtaala:
> One slight issue with the tooltip: It appears on the wrong monitor :D My topbar is on the right monitor and the tooltip shows on the left monitor. But where I would expect it if I hover over the icon on that monitor. 

Oh lol, yes I see - good pickup! (I usually don't use multi monitors - will fix that). _FIXED_ :heavy_check_mark: 

> It also doesn't show up if I hover over the icon on my secondary monitor (which is probably also fine, we can improve the behavior if/when we decide how to do topbars on a seconary monitor).

Well, given that icon is clickable (even on other monitors) it prob makes sense to have the tooltip show up on that monitor too.

> Also maybe make the tooltip background a bit more opaque.

Good call.  Have a look at the `focus-button-tooltip` css class in `stylesheet.css` - it's set for `0.35` opacity - let me know a good new opacity!

> Are we allowed to just copy them? What is the license on them? Maybe we need to include a notice somewhere. I would maybe put it either in a comment in the svgs or in a separate file next to the icons.

Good call - they're from `Adwaita` theme, and looks like [LGPL license](https://github.com/GNOME/adwaita-icon-theme/blob/master/COPYING_LGPL)? I'll put a note in the resources folder referencing the license - I assume that's all that needs to be done?  Anyone have experience here?

jtaala:
Yes, we can copy and package the icons - see COPY terms here:

https://github.com/GNOME/adwaita-icon-theme/blob/master/COPYING

I've added an `ICONS.info` file to resources folder attributing to "Gnome Project" and put some links in there too.

Finally, I've now made the focus icon on secondary (etc.) monitor also show tooltip.

jtaala:
Thanks all for testing thus far - please give the latest a test; I believe it's ready!

resources/ICONS.info
```diff
@@ -0,0 +1,5 @@
+SVG icons adapted from "GNOME Project" adwaita-icon-theme (https://github.com/GNOME/adwaita-icon-theme).
+
+See http://www.gnome.org for more information.
+
+Please see terms, under which these icons have been copied, at https://github.com/GNOME/adwaita-icon-theme/blob/master/COPYING.
```
Lythenas:
```suggestion
The icons are licensed under Creative Commons Attribution-Share Alike 3.0 United States License.

To view a copy of the CC-BY-SA licence, visit
http://creativecommons.org/licenses/by-sa/3.0/ or send a letter to Creative
Commons, 171 Second Street, Suite 300, San Francisco, California 94105, USA.
```

Just to be safe.

Also maybe we should list the original names of the icons.

Lythenas (changes requested):
I think the behavior is now fine.

I will test the tooltip tomorrow and the probably approve.

jtaala:
Oki-doki, updated ICONS.info with reference to original icon filenames used.  Also, updated README.md with a small section on focus modes (including hiding the focus-mode button in topbar via dconf if users wish).

Lythenas (approved):
The tooltip works well now.

FYI this is the tooltip definition for Adwaita: https://gitlab.gnome.org/GNOME/libadwaita/-/blob/main/src/stylesheet/widgets/_tooltip.scss They are using 0.8 transparency. But I think it is fine the way it is.

PR #483, opened 2023-03-05
WIP: Show workspace indicator on all monitors

Lythenas:
Show the workspace indicator on all dash-to-panel panels (because that is what I am currently using).

If you know of a better way to apply these changes to all panels, that doesn't rely on checking for a specific extension, please let me know.

Some things are not yet working:

- "follow focus" currently doesn't work so great. Maybe I broke it or it just doesn't work when dash-to-panel is active.
- when switching workspace on one monitor (Super+PageUp/Down), the indicators on all monitors are updated. But corrected when to monitor get focus again.

### DIFF
--- a/topbar.js
+++ b/topbar.js
@@ -217,9 +217,9 @@ class WorkspaceMenu extends PanelMenu.Button {
         this.add_actor(this._label);
 
         this.signals = new Utils.Signals();
-        this.signals.connect(global.window_manager,
-                             'switch-workspace',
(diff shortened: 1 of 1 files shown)


jtaala:
Hey @Lythenas - should we close this one? (or are you still working/intending to work on it?).

Lythenas:
I stopped using dash-to-panel a while ago. So I will close this.

I still would like to have the date time on all monitors, but I will probably look at another way to implement that instead of using dash-to-panel.

PR #477, opened 2023-02-25
FIXES: (1) removed deprecated actor call; (2) implemented guard against multiple `disable`->`enable`->`disable` spam calls

jtaala:
This PR implements two fixes, as described below.  Also fixes #432

### FIX 1: removed a deprecated actor call that is spamming logs

https://github.com/paperwm/PaperWM/blob/f590f8b30f0c1962e2bc18f1a39355b0a72636cb/topbar.js#L567

is deprecated and should be replace with `Main.panel.statusArea.activities.show()`, which removes the following log spam every time paperwm is disabled.

```
Feb 25 15:54:15 jay-blade gnome-shell[1009490]: Usage of object.actor is deprecated for ActivitiesButton
                                                get@resource:///org/gnome/shell/ui/environment.js:387:29
                                                disable@/home/jay/.local/share/gnome-shell/extensions/paperwm@hedning:matrix.org/topbar.js:567:5
                                                safeCall@/home/jay/.local/share/gnome-shell/extensions/paperwm@hedning:matrix.org/extension.js:61:52
                                                run@/home/jay/.local/share/gnome-shell/extensions/paperwm@hedning:matrix.org/extension.js:46:14
                                                disable@/home/jay/.local/share/gnome-shell/extensions/paperwm@hedning:matrix.org/extension.js:122:12
                                                _callExtensionDisable@resource:///org/gnome/shell/ui/extensionSystem.js:125:32
                                                _onEnabledExtensionsChanged/<@resource:///org/gnome/shell/ui/extensionSystem.js:536:45
                                                _onEnabledExtensionsChanged@resource:///org/gnome/shell/ui/extensionSystem.js:536:24
                                                createCheckedMethod/<@resource:///org/gnome/gjs/modules/core/overrides/Gio.js:686:46
                                                disableExtension@resource:///org/gnome/shell/ui/extensionSystem.js:226:29
                                                DisableExtension@resource:///org/gnome/shell/ui/shellDBus.js:451:38
                                                _handleMethodCall@resource:///org/gnome/gjs/modules/core/overrides/Gio.js:324:38
                                                _wrapJSObject/<@resource:///org/gnome/gjs/modules/core/overrides/Gio.js:401:34

```

### FIX 2: implemented guard against multiple gnome spam calls `disable` -> `enable` -> `disable` etc.

PaperWM can go through multiple `enable` -> `disable` -> `enable` cycles (sometimes rapidly).  This is alluded to here:
[...]

PR #484, opened 2023-03-05
WIP: Add button in settings to reload css

Lythenas:
As it turns out reloading the CSS was easy. Just call `Main.loadTheme()`.

But I'm having some trouble with the UI. The button shows up, but for some reason the click-listener does not work. Also as far as I can tell `init` in prefs.js is also never called. I pieced together that this is automatically loaded by gnome-shell and not by us, but I can't get it to work.

@jtaala I would be grateful for some pointers here. Or if you want you can just implement it yourself and I will take a look at that.

### DIFF
--- a/Settings.ui
+++ b/Settings.ui
@@ -709,6 +709,12 @@
                 </child>
               </object>
             </child>
+            <child>
+              <object class="GtkButton" id="reload-css">
+                <property name="label">Reload CSS</property>
+                <property name="tooltip_text" translatable="yes">Reload all currently loaded CSS files (from all extensions and the user.css)</property>
+              </object>
+            </child>>
           </object>
         </child>
       </object>
--- a/extension.js
+++ b/extension.js
@@ -105,6 +105,13 @@ function init() {
 
 function enable() {
     log(`#paperwm enable ${SESSIONID}`);
+
+    let userStylesheet = getConfigDir().get_child("user.css");
+    if (userStylesheet.query_exists(null)) {
+        let themeContext = St.ThemeContext.get_for_stage(global.stage);
+        themeContext.get_theme().load_stylesheet(userStylesheet);
+    }
+
     if (enabled) {
         log('enable called without calling disable');
         return;
@@ -209,12 +216,6 @@ function initUserConfig() {
 
     if (hasUserConfigFile()) {
         Extension.imports.searchPath.push(getConfigDir().get_path());
-    }
-
-    let userStylesheet = getConfigDir().get_child("user.css");
-    if (userStylesheet.query_exists(null)) {
-        let themeContext = St.ThemeContext.get_for_stage(global.stage);
-        themeContext.get_theme().load_stylesheet(userStylesheet);
     }
 }
 
--- a/prefs.js
+++ b/prefs.js
@@ -1,3 +1,10 @@
+/*
+ * Defines the extensions settings window.
+ *
+ * This file is automatically loaded by gnome-shell when opening the settings
+ * for this extension (e.g. by clicking the "Settings" button in the Extensions
+ * app).
+ */
 
 const Gio = imports.gi.Gio;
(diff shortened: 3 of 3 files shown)


jtaala:
> @jtaala I would be grateful for some pointers here. Or if you want you can just implement it yourself and I will take a look at that.

Hey @Lythenas!  So, your button is actually working but there's a few issues in play here.

First, `prefs.js` is treated a little differently in gnome as compared to other modules in the extension.  It's handled by gnome and apparently doesn't have access to `gnome-shell` like the other extension modules do.  So, you can actually see your logging events but you'll need to do a `journalctl -f /usr/bin/gjs` to see them.

A couple of links re this here:
https://www.reddit.com/r/gnome/comments/w0r9sf/gjs_gnomeextensions_import_ui_modules/
https://discourse.gnome.org/t/how-to-see-logs-from-prefs-js/6079/2

That also leads  to the second problem - since we don't have access to `gnome-shell` from `prefs.js` then `imports.main.ui` won't work here (since `ui` is part of gnome-shell).

How the PaperWM picks up changes from the other prefs settings is through gsettings - basically `prefs.js` changes a setting via gsettings and PaperWM receives that change since it's defined in it's schema.  Kind of like if you open up `dconf` and change a PaperWM setting in there PaperWM will pick it up and react.

In any case, I'm thinking of a way around this - but might have to work through gsettings?

Hopefully there's a simpler way we can find for this use case.

Lythenas:
> First, prefs.js is treated a little differently in gnome as compared to other modules in the extension. It's handled by gnome and apparently doesn't have access to gnome-shell like the other extension modules do. So, you can actually see your logging events but you'll need to do a journalctl -f /usr/bin/gjs to see them.

Oh man, that is good to know...

We could always use dbus or something like that. But it might not be worth implementing something complicated. The user can also just manually do (which might be just as fast as opening the extension settings and clicking the button):
- Alt+F2
- lg
- `Main.loadTheme()`

So maybe we should just add that to the readme.

jtaala:
Wait, ah, it's `Main.loadTheme()` (not `Main.ui.loadTheme`) - I feel silly.  I also feel silly that it's as easy as calling `Main.loadTheme()`...).  Nice work finding that!

Anyways, yeah, thinking it likely isn't worth putting in prefs - we could put it in the `README.md` like you said, or just add it to the "enable()" method in `extensions.js` so users can just disable then enable PaperWM if they want to load themes again.

Interestingly, on that last approach, I'm getting a gnome crash, which I surmise is due to needing #477 since gnome is disabling/enabling PaperWM as many times as the number of extensions (see fix2 in #477).

jtaala:
> which I surmise is due to needing #477 since gnome is disabling/enabling PaperWM as many times as the number of extensions (see fix2 in #477).

Scrap that - maybe not.  Putting that command in `enable()` method works... but when you disable and then enable - and run that same command - you get a GJS error.  Actually, get the same thing when you disable, then enable, then try running that command from `alt`+`f2` - `Main.loadTheme()` so there's something deeper going on (prob with themes and cleanup when disabling PaperWM).

I'll dig a little more but you may be right - might not be worth pushing too much since could use that command from `lg` or just log out / log in.

Lythenas:
Interesting. I haven't gotten any errors yet when calling `Main.loadTheme()`. Even if there is a syntax error in the css. It will just "fail" silently.

FYI `Main.loadTheme()` is actually `imports.ui.main.loadTheme()`: https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/js/ui/main.js#L491

I will close this PR and make a new one with a comment in the README.

jtaala:
Ha, I wonder if I had something else weird in my branch at the time (was testing, playing around) - can't reproduce now (even after enable/disable).

Thanks @Lythenas.

Lythenas:
Closing in favor of #487

PR #473, opened 2023-01-31
Show next stackoverlay window preview when switching to offscreen window with mouse

jtaala:
Closes #470.

This PR automatically shows the next stackoverlay window preview (if window exists) when a user clicks with the current preview.  Previous behaviour did not show the next window preview meaning users had to move the mouse away, and then back against the screen edge to show the next preview.

Note: I've also removed the 2 second timeout on showing next window previews.  This feels better to me - but please comment if you think the original (2 second timeout for previews) is better.

_NOTE: this PR has been implemented in the [PaperWM-redux](https://github.com/PaperWM-redux/PaperWM) fork, which you can install if you want this, or any of [my other open PRs](https://github.com/paperwm/PaperWM/pulls/jtaala)._

PR #488, opened 2023-03-11
Replaces #364: removes the left and right tiling restriction that caused no margin to be shown

jtaala:
Closes #364.

This PR was created since @basdelfos develop branch/fork isn't available anymore :-(

Please see #364 for details of this.  It has been a long time coming.  Given the merging of #471 (an improved way to see the window position), and also the merging of #473, having no margin on the left/right tiling edge no longer provides the original benefit (of seeing when you're at an tiling edge).

This simply removes the cases where no left/right margin would be shown - which also cleans-up/simplifies this area of code (thanks @smichel17 ). 

Note: we will close #467 as it was originally a way to set this separately.  However we will be closing that in favour of this one.

### DIFF
--- a/tiling.js
+++ b/tiling.js
@@ -2883,12 +2883,6 @@ function ensuredX(meta_window, space) {
     let max = min + workArea.width;
     if (meta_window.fullscreen) {
         x = 0;
-    } else if (index == 0 && x <= min) {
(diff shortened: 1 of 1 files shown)


Lythenas (approved):
I have been running these changes for a couple of days and it seems to work without issues,

Issue #376, opened 2021-03-18
Support GNOME 40


gavr123456789:

mrspikeroot:
GNOME put out this info for extension developers
https://blogs.gnome.org/shell-dev/2021/03/20/extensions-rebooted-porting-your-existing-extensions-to-gnome-40/

hedning:
Have` started working on support for gnome shell 40 here: https://github.com/paperwm/PaperWM/tree/next-release

Things should start and launching windows should work, but there's quite a few bugs (eg. scrolling the workspaces seems to crash the process, and dragging windows are broken).

qwIvan:
> Have` started working on support for gnome shell 40 here: https://github.com/paperwm/PaperWM/tree/next-release
> 
> Things should start and launching windows should work, but there's quite a few bugs (eg. scrolling the workspaces seems to crash the process, and dragging windows are broken).

The configuration panel doesn't works now. How can I modify the configuration manually?

wbolster:
using dconf-editor, orn using the gsettings command line tool. you can check my wbolster/dotfiles repo (gnome subdirectory) for some examples

qwIvan:
> using dconf-editor, orn using the gsettings command line tool. you can check my wbolster/dotfiles repo (gnome subdirectory) for some examples

Animation of switch focus between two monitor is incorrect. How to disable animation?

qwIvan:
> using dconf-editor, orn using the gsettings command line tool. you can check my wbolster/dotfiles repo (gnome subdirectory) for some examples

![image](https://user-images.githubusercontent.com/7613160/113989011-99e4a100-9882-11eb-94be-556215f7cfdd.png)
The above command is refer from your dotfile, but it doesn't work.

wbolster:
ah yes, this extension's name is problematic because it contains a `:` (colon), which the `gsettings` tool treats as a path separator (`GSETTINGS_SCHEMA_DIR`). i reported this before in #322.

i use an ugly symlink based hack to work around it:

```
    # HACK: The colon in the name confuses the GSETTINGS_SCHEMA_DIR
    # environment variable used by the ‘gsettings’ tool, so trick
    # it into picking up the right directory anyway.
    ln -s 'paperwm@hedning:matrix.org/schemas' 'paperwm@hedning'
```

see https://github.com/wbolster/dotfiles/blob/main/gnome/install-extensions#L39-L42

qwIvan:
[...]

PR #489, opened 2023-03-11
Extension UUID name change to avoid path issues (remove `:` from `paperwm@hedning:matrix.org`)

### LINKED ISSUE #322: path name with colon causes problems with gsettings and GSETTINGS_SCHEMA_DIR
this extension uses a `:` (colon) character in its directory name in `~/.local/gnome-shell/extensions`, namely `paperwm@hedning:matrix.org`, which causes problems when using `GSETTINGS_SCHEMA_DIR=...` and the `gsettings` command line tool. the `:` is interpreted as a path separator.

this means `gsettings` cannot be used to configure this extension, as suggested in the readme.

since this is only the extension (out of quite a few on my system) using a non-standard character, i suggest to adhere to the established practices and conventions for gnome extensions, and use an identifier and directory name that do not cause problems.

jtaala:
Fixes #322.   Also mentioned in [#376](https://github.com/paperwm/PaperWM/issues/376#issuecomment-815561175).

As noted in #322, having a colon `:` in the extension uuid/name can cause issues with using gsettings(?) to change settings.

To avoid this, and any potential future issues with having a non-standard(?) extension uuid, this PR changes it to a more standard (note the `.` instead of a `:`) `paperwm@hedning.matrix.org`.

The only issue I can see, if that users would need to run `./install.sh` again and logout/login... but if not now, when? :stuck_out_tongue_winking_eye:

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -33,7 +33,7 @@ GNOME Shell 43.3
 PaperWM branch/tag: develop
 PaperWM commit: 223ff883bca9bf20dbf066eceda891cb6e8be931
 Enabled extensions:
-- paperwm@hedning:matrix.org
+- paperwm@hedning.matrix.org
 - switcher@landau.fi
 - dash-to-panel@jderose9.github.com
 - appindicatorsupport@rgcjonas.gmail.com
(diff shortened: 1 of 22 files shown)


Lythenas:
Is there some way we could notify the users that they need to reinstall? Or will gnome just refuse to load the extension after a git pull because id and folder names don't match anymore?

Also maybe @hedning wants to have a say here since his name is in the UUID.

Since we are changing the name anyway alternatives could be:

- `paperwm@hedning.github.com`
- or even just `paperwm@paperwm.github.com`

This is the pattern that e.g. dash-to-panel uses and it is also more or less suggested by https://gjs.guide/extensions/overview/anatomy.html#required-fields.

But keeping the `matrix.org` is also fine by me.

jtaala:
> Is there some way we could notify the users that they need to reinstall? Or will gnome just refuse to load the extension after a git pull because id and folder names don't match anymore?

I'll do some more testing when I can - but when I changed it all, I ran the install script and then the extension didn't show up until logged out and logged in. 

> Also maybe @hedning wants to have a say here since his name is in the UUID.

Good idea - happy to keep this PR open to gather any feedback (and maybe a response from @hedning).

I like your suggestions too.

jtaala:
Closing in favour of #556.

PR #491, opened 2023-03-15
Allow enabling the extension in Gnome 44

Lythenas:
This allows enabling the extension in Gnome 44 (even if there might be errors)

Do we also want to increase the version number of PaperWM to "44.0"? We can also wait until we officially support Gnome 44.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ We hang out on [zulip](https://paperwm.zulipchat.com).
 
 Clone the repo and check out the branch supporting the Gnome Shell version you're running.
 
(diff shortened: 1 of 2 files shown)


jtaala:
> Do we also want to increase the version number of PaperWM to "44.0"? We can also wait until we officially support Gnome 44.

Yes, let's wait until we confirm it's working fine on 44 before increasing the PaperWM version number - this is just to enable it so we can confirm (hopefully with the help of others submitted bug reports etc.).

jtaala (approved):
Looks good!

I'll start a topic on zulip to discuss moving forward how we want to manage gnome versions and their associated branches.

PR #493, opened 2023-03-19
Fix swipe gesture trackers and workspace switching / double workspace animation (no longer need `Vertical overview` extension to workaround)

### LINKED ISSUE #413: Wayland 3-finger swipe broken?
I've been testing out the Gnome 42 branch (referenced [here](https://github.com/paperwm/PaperWM/issues/409)) and it works pretty good. The only thing I've found to be broken is the 3-finger swipe is meant to be supported on Wayland (according to the README).

Is this a known problem or something with my local setup?

When I try 3-finger swipe sideways, the window jitter a bit and sometimes move but not reliably. I think it would be great if swiping through horizontally with 3 fingers would just scroll the way a webpage scrolls, i.e. _without_ snapping. The user's next click on the window will snap it into place anyway. This way, it would be super easy to just scroll through all windows that are open, skimming them as they go past and choosing the one that you want by stopping the scrolling and focusing a window.

jtaala:
Fixes #413.

This PR fixes/updates swipe gesture tracking disabling in PaperWM, and implements a fix for gnome resetting this swipe gesture tracking after showing overview.

PaperWM disables the **default** swipe gesture tracking for workspace switching (only applicable to Wayland), and instead uses it's own one.  However, gnome was resetting the default switching trackers after going into overview mode, resulting in it breaking the paperwm swipe tracking, and the weird double-switching animation bug.  Changes in this PR ensure default swipe tracking gestures for workspace switching remains disabled.

Previously, a way to alleviate this issue was by using the [Vertical overview](https://extensions.gnome.org/extension/4144/vertical-overview/) extension.  However, this was just working around the issue.  Since the extension enforces vertical workspaces, 3-finger swipe gestures (which usually switches to next horizontal workspace) are not applicable in vertical workspaces.

This PR alleviates the need to use the `Vertical overview` extension to avoid the aforementioned issue.

This makes it a bit easier for new users who are often confused (annoyed?) by the 3-finger window tiling swipe gesture breaking and the workspace switching with the "double workspace" animation that results.

~~Although [this comment](https://github.com/paperwm/PaperWM/issues/413#issuecomment-1320324268) suggests it's still currently needed to avoid an issue in multiple-monitor setups (where windows may not appear on the correct monitor when in overview mode).  However, that issue still seems to occur even with vertical overview for me (when I'm using multiple monitors).~~  Apologies!  I misread that comment - it's saying that issue still likely occurs even with vertical overview extension (which it looks like it does).

### DIFF
--- a/kludges.js
+++ b/kludges.js
@@ -55,7 +55,6 @@ if (!global.display.get_monitor_neighbor_index) {
     }
 }
 
-
 if (!global.display.set_cursor) {
     global.display.constructor.prototype.set_cursor = global.screen.set_cursor.bind(global.screen);
 }
@@ -339,7 +338,6 @@ function getSavedPrototype(obj, name) {
     return getSavedProp(obj.prototype, name);
 }
 
-
 function disableOverride(obj, name) {
     obj[name] = getSavedProp(obj, name);
 }
@@ -375,6 +373,11 @@ function restoreMethod(obj, name) {
 }
 
 var signals;
+var swipeTrackers = [
+    Main.overview?._swipeTracker,
+    Main.wm._workspaceAnimation?._swipeTracker,
+    Main.overview?._overview?._controls?._workspacesDisplay?._swipeTracker
+].filter(t => t !== undefined);
 function init() {
     registerOverridePrototype(imports.ui.messageTray.MessageTray, '_updateState');
     registerOverridePrototype(WindowManager.WindowManager, '_prepareWorkspaceSwitch');
@@ -390,7 +393,6 @@ function init() {
     registerOverridePrototype(WindowManager.WorkspaceTracker, '_checkWorkspaces', _checkWorkspaces);
     if (WindowManager.TouchpadWorkspaceSwitchAction) // disable 4-finger swipe
         registerOverridePrototype(WindowManager.TouchpadWorkspaceSwitchAction, '_checkActivated', () => false);
-
 
     // Work around https://gitlab.gnome.org/GNOME/gnome-shell/issues/1884
     if (!WindowManager.WindowManager.prototype._removeEffect) {
@@ -422,18 +424,10 @@ function init() {
         registerOverrideProp(imports.ui.viewSelector, "PINCH_GESTURE_THRESHOLD", 0)
     }
 
-    const overviewSwipeTracker = Main.overview._swipeTracker
-    if (overviewSwipeTracker) {
-        registerOverrideProp(overviewSwipeTracker, "enabled", false)
-    }
-
-    const workspaceSwipeTracker = Main.wm._workspaceAnimation?._swipeTracker
-    if (workspaceSwipeTracker) {
-        registerOverrideProp(workspaceSwipeTracker, "enabled", false)
-    }
-
-    if (Main.wm._swipeTracker)
-        registerOverrideProp(Main.wm._swipeTracker._touchpadGesture, "enabled", false);
+    // disable swipe gesture trackers
+    swipeTrackers.forEach(t => {
+        registerOverrideProp(t, "enabled", false);
+    });
 
     registerOverridePrototype(Workspace.Workspace, '_isOverviewWindow', (win) => {
         let metaWindow = win.meta_window || win;
@@ -585,7 +579,7 @@ function disable() {
     Main.layoutManager._updateHotCorners();
 }
 
-// 3.32 overivew layout
+// 3.32 overview layout
 function computeLayout(windows, layout) {
     let numRows = layout.numRows;
 
@@ -868,7 +862,6 @@ function _checkWorkspaces() {
     this._checkWorkspacesId = 0;
     return false;
 };
-
 
 function addWindow(window, metaWindow) {
     if (this._windows.has(window))
--- a/kludges.js
+++ b/kludges.js
@@ -374,10 +374,10 @@ function restoreMethod(obj, name) {
 
 var signals;
 var swipeTrackers = [
-    Main.overview?._swipeTracker,
-    Main.overview?._overview?._controls?._workspacesDisplay?._swipeTracker,
-    Main.wm?._swipeTracker,
-    Main.wm._workspaceAnimation?._swipeTracker
+    Main?.overview?._swipeTracker,
+    Main?.overview?._overview?._controls?._workspacesDisplay?._swipeTracker,
+    Main?.wm?._swipeTracker,
+    Main?.wm?._workspaceAnimation?._swipeTracker
 ].filter(t => t !== undefined);
 function init() {
     registerOverridePrototype(imports.ui.messageTray.MessageTray, '_updateState');
--- a/kludges.js
+++ b/kludges.js
@@ -375,8 +375,9 @@ function restoreMethod(obj, name) {
 var signals;
 var swipeTrackers = [
     Main.overview?._swipeTracker,
-    Main.wm._workspaceAnimation?._swipeTracker,
-    Main.overview?._overview?._controls?._workspacesDisplay?._swipeTracker
+    Main.overview?._overview?._controls?._workspacesDisplay?._swipeTracker,
+    Main.wm?._swipeTracker,
+    Main.wm._workspaceAnimation?._swipeTracker
 ].filter(t => t !== undefined);
 function init() {
     registerOverridePrototype(imports.ui.messageTray.MessageTray, '_updateState');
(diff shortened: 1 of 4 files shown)


kludges.js
```diff
@@ -375,6 +373,11 @@ function restoreMethod(obj, name) {
 }
 
 var signals;
+var swipeTrackers = [
+    Main.overview?._swipeTracker,
+    Main.wm._workspaceAnimation?._swipeTracker,
+    Main.overview?._overview?._controls?._workspacesDisplay?._swipeTracker
+].filter(t => t !== undefined);
```
Lythenas: Is this ok to be done on the top level or should it be done in enable/init? I'm unsure when the fields in Main.overview and Main.wm are set.

kludges.js
```diff
@@ -422,18 +424,10 @@ function init() {
...
-        registerOverrideProp(overviewSwipeTracker, "enabled", false)
-    }
-
-    const workspaceSwipeTracker = Main.wm._workspaceAnimation?._swipeTracker
-    if (workspaceSwipeTracker) {
-        registerOverrideProp(workspaceSwipeTracker, "enabled", false)
-    }
-
-    if (Main.wm._swipeTracker)
-        registerOverrideProp(Main.wm._swipeTracker._touchpadGesture, "enabled", false);
```
Lythenas, line 436: Is it correct to not include this in the swipeTrackers list? I would rather include to much than to little so we maybe support more gnome versions.

Lythenas (commented):
I haven't used the gestures before. I will try to test this on my laptop. I'm guessing the gestures work on trackpack and not just on a touch screen?

kludges.js
jtaala: Good question! - it's done at the top level so it can be accessed by `tiling.js` (e.g. - when re-enabling the overrides on `overview` hidden signal) - otherwise we would have just defined in `init` (i.e. if not used elsewhere).

kludges.js
jtaala: Ah, good pickup!  Yes, in (at least Gnome 43) it looks like it was moved to `Main.wm._workspaceAnimation?._swipeTracker` but yes, will include it there too.  The filter on `swipeTrackers` will filter it out if it's undefined (which it will be since we'll use an optional chain on it).

Thanks @Lythenas!

kludges.js
jtaala: > I'm unsure when the fields in Main.overview and Main.wm are set.

Oh, yes, they are set (by order, so are set first, well before referenced here)  :smiley:

jtaala:
> I haven't used the gestures before. I will try to test this on my laptop. I'm guessing the gestures work on trackpack and not just on a touch screen?

Yes, it's actually just the 3-finger swipe gesture (used to move the tiled windows left/right).  Without this PR (e.g. on current `develop` branch) try disabling `vertical overview` extension, going into overview / going out of overview (at least once), and then 3-finger swiping again, it'll do a workspace switch instead (default gnome action for 3-finger swipe on wayland).

Yes, just works on trackpad.

kludges.js
jtaala: Updated (re-added back in).

kludges.js
```diff
@@ -375,6 +373,12 @@ function restoreMethod(obj, name) {
 }
 
 var signals;
+var swipeTrackers = [
+    Main?.overview?._swipeTracker,
+    Main?.overview?._overview?._controls?._workspacesDisplay?._swipeTracker,
+    Main?.wm?._swipeTracker,
+    Main?.wm?._workspaceAnimation?._swipeTracker
+].filter(t => t !== undefined);
```
smichel17:
```suggestion
].filter(t => typeof t !== "undefined");
```
Technically someone can do `var undefined = "foo"`. Unlikely, but `typeof` will never break this way.

kludges.js
```diff
@@ -477,10 +478,8 @@ function enable() {
             disableOverride(Workspace.Workspace.prototype, '_isOverviewWindow');
         }
     }
-    signals.connect(settings, 'changed::only-scratch-in-overview',
-                    scratchInOverview);
-    signals.connect(settings, 'changed::disable-scratch-in-overview',
-                    scratchInOverview);
+    signals.connect(settings, 'changed::only-scratch-in-overview',scratchInOverview);
```
smichel17:
```suggestion
    signals.connect(settings, 'changed::only-scratch-in-overview', scratchInOverview);
```

topbar.js
```diff
@@ -201,11 +201,12 @@ class ColorEntry {
  */
 var FocusIcon = Utils.registerClass(
 class FocusIcon extends St.Icon {
-    _init(properties = {}, tooltip_x_point=0) {
+    _init(properties = {}, tooltip_parent, tooltip_x_point=0) {
         super._init(properties);
         this.reactive = true;
 
         // allow custom x position for tooltip
+        this.tooltip_parent = tooltip_parent ?? this;
```
smichel17, line 209: I like using `??` over `||`, but it would be good to check which version of gnome-shell adds support for it. No sense in supporting compatibility for older versions of gnome shell (e.g. looking for swipeTracker in different locations) if we're going to use js syntax that's not supported on those versions.

kludges.js
```diff
@@ -375,6 +373,12 @@ function restoreMethod(obj, name) {
 }
 
 var signals;
+var swipeTrackers = [
+    Main?.overview?._swipeTracker,
+    Main?.overview?._overview?._controls?._workspacesDisplay?._swipeTracker,
+    Main?.wm?._swipeTracker,
+    Main?.wm?._workspaceAnimation?._swipeTracker
```
smichel17: I think it would be good to comment which of these correspond to which gnome versions. Makes it easier to drop support for old versions when we're ready.

smichel17 (approved):
Only looked at the code, have not tested yet.

topbar.js
Lythenas: According to my very quick research (take with a grain of salt): ?? support was added in GJS v1.65.90 (https://gitlab.gnome.org/GNOME/gjs/-/blob/master/NEWS#L993), Gnome version 40.4 was still using GJS 1.65.1 (https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/40.4/meson.build#L26) and Gnome version 41.1 is using GJS version 1.69.2 (https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/41.1/meson.build#L27).

"Officially" (according to the readme) we only support gnome 43 (and plan to support 44) on develop. However I'm personally also using gnome 42 on my work laptop. I think we should maybe decide or state more clearly which versions are actively supported.

jtaala: Good find @Lythenas!

So it looks like suppport for `??` and `?.` were the added/applied in gnome `41.0`:

https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/40.10/meson.build#L26
vs.
https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/41.0/meson.build#L27

jtaala: So, technically, those changes should be fine in `41` onwards but we haven't tested it.  Sounds like @Lythenas has tested/used the latest stuff in `42`.

I'm torn here, it can be quite a bit of work to support older versions (especially if we're not running those older version), and in some cases can limit new implementations and improvements (that's prob why I like rolling-releases in general... since I maintain some packages in the AUR as well).

jtaala: Let's carry on this thread in zulip (I was going to put up a thread approach for when new gnome versions come out... e.g. for `44` do we just then take a snapshot of the current dev in a branch called `gome-43` and that's what people get if they're on `43`?).

kludges.js
jtaala: Thanks @smichel17, have implemented these changes (see latest commit).

kludges.js
jtaala: Thanks @smichel17, have implemented these changes (see latest commit).

kludges.js
jtaala: Thanks @smichel17, have implemented these changes (see latest commit).

Lythenas:
As far as I can tell this seems to work. But I haven't checked what the previous buggy behavior was or how it works on xorg. I only tested wayland.

jtaala:
> As far as I can tell this seems to work. But I haven't checked what the previous buggy behavior was or how it works on xorg. I only tested wayland.

All good - the behaviour which triggered this issue was in wayland (i.e. swipe tracking - which isn't available in X11).  I've been testing in both btw.

PR #500, opened 2023-03-26
FIX: after mouse grab/move-window end operation, ensure window is correctly activated

jtaala:
This PR fixes an observed issue when grabbing a window and moving it (e.g. while holding down `alt` and click-hold a window to move it with the mouse).

Currently the selected window is not selected/activated after the grab-end operation.  In an unselected window state, select left/right window keybinds don't work, nor can shortcuts can't be passed to the selected window.

This PR ensures the window is correctly activated after grab-end operations.

### DIFF
--- a/grab.js
+++ b/grab.js
@@ -9,6 +9,7 @@ var Meta = imports.gi.Meta;
 var Clutter = imports.gi.Clutter;
 var St = imports.gi.St;
 var Main = imports.ui.main;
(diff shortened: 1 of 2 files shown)


Lythenas (approved):
THB I couldn't reproduce the issue you are talking about. But as far as I can tell this PR does also not break anything.

Issue #347, opened 2020-10-28
Animations glitch on overview


TheMascarade:
Happening since updating to gnome 3.38.
OS: Fedora 33
Using latest release.
![2020-10-28-13-03-54](https://user-images.githubusercontent.com/31579020/97464726-08125880-1920-11eb-9773-e400e18721fa.gif)

azmeuk:
I can confirm this too with Archlinux.

hedning:
Yeah, upstream did a fairly substantial rewrite of the overview layout code in 3.38 which I haven't had the time to monkey patch correctly yet.

ejemba:
@hedning Do you think you can work on the rewrite because basically now nothin is working properly under Archlinux :'( 

Or should I revert my version of gnome shell ?

hedning:
@ejemba  I'm on 3.38 now, so will hopefully get the time/motivation to fix things.

In the meantime I'd double check that you've turned `only-scratch-in-overview`. Should look like this in paperwm-preference (if you're on develop):
![image](https://user-images.githubusercontent.com/71978/98576076-4f9eca00-22ba-11eb-88a9-99a3ead81ebd.png)

hedning:
OK, pushed 261ae9de8fcdefd86d70d52b72cd6019db4cf8d2, which should hopefully make things usable.

ejemba:
@hedning I just pull your commit and refresh my gnome-shell
I have the same scratch configuration than you.
It's strange if I have at least one window that is in fullscreen, the overview screen bugs.
If I reduce this same windows by lets say 200pixels then the overview is working again.

ejemba:
Always got the glitch in function of the disposition of the windows (cf previous message)
here are my error messages

```
.. gnome-shell[1079]: Can't update stage views actor overviewGroup is on because it needs an allocation.
.. gnome-shell[1079]: Can't update stage views actor overview is on because it needs an allocation.
.. gnome-shell[1079]: Can't update stage views actor overviewGroup is on because it needs an allocation.
.. gnome-shell[1079]: Can't update stage views actor overview is on because it needs an allocation.
.. gnome-shell[1079]: Can't update stage views actor overviewGroup is on because it needs an allocation.
.. gnome-shell[1079]: Can't update stage views actor overview is on because it needs an allocation.
.. gnome-shell[1079]: Can't update stage views actor overviewGroup is on because it needs an allocation.
[...]

PR #501, opened 2023-03-28
FIX and improve new window create/insert tween (animation)

### LINKED ISSUE #328: Visual artifacts when opening windows with mutter 3.36.4
3.36.3 (good)

3.36.4 (bad, notice the blue rectangles)

I bisected the mutter commit that introduced this down to https://gitlab.gnome.org/GNOME/mutter/-/commit/44a3358125317dbe2151cf0ac70ec2c382236881.

I can only reproduce this using PaperWM, vanilla GNOME doesn't seem to be affected.

jtaala:
Fixes #328.

This PR fixes and improves the new window create/insert tween.  Also resolves a noticable issue where part of current selected window is clipped during window creation.

See below for comparison between current tween and new tween:

**Note the transitions have been slowed down for demonstration purposes.  Also, I'm using a styled `selection` (pinkish colour - because I like it but also to see it better for this demonstration)**

### current transitions (slowed down so can see issues with certain applications):
https://user-images.githubusercontent.com/30424662/228228961-34250abe-7fb6-4141-a1c7-a55f23b54ec4.mp4

### new (improved) transitions (slowed down to see them better):
https://user-images.githubusercontent.com/30424662/228229082-d0088765-6cd8-499f-8c06-26dfba21952e.mp4

### DIFF
--- a/tiling.js
+++ b/tiling.js
@@ -1081,6 +1081,14 @@ var Space = class Space extends Array {
         }
     }
 
+    hideSelection() {
+        this.selection.style_class = 'background-clear';
+    }
+
+    showSelection() {
+        this.selection.style_class = 'paperwm-selection tile-preview';
+    }
+
     setSelectionActive() {
         this.selection.opacity = 255;
     }
@@ -2520,7 +2528,7 @@ var Spaces = class Spaces extends Map {
                So even if we set it again in `first-frame` that is too late
                since that happens _after_ mutter have drawn the frame.
 
-               So we kill visibily on the first the `queue-redraw`.
+               So we kill visibily on the first `queue-redraw`.
             */
             signals.connectOneShot(actor, 'queue-redraw', () =>  {
                 actor.opacity = 0;
@@ -2911,9 +2919,11 @@ function insertWindow(metaWindow, {existing}) {
 
     if (!existing) {
         actor.opacity = 0;
+        actor.visible = false;
         clone.x = clone.targetX;
         clone.y = clone.targetY;
-        clone.set_scale(0, 0);
+        clone.set_scale(0, 1);
+        space.hideSelection();
         Tweener.addTween(clone, {
             scale_x: 1,
             scale_y: 1,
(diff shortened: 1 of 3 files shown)


paradajz:
Does this also fixes https://github.com/paperwm/PaperWM/issues/347?

tiling.js
```diff
@@ -2911,16 +2919,19 @@ function insertWindow(metaWindow, {existing}) {
 
     if (!existing) {
         actor.opacity = 0;
+        actor.visible = false;
         clone.x = clone.targetX;
         clone.y = clone.targetY;
-        clone.set_scale(0, 0);
+        clone.set_scale(0, 1);
```
Lythenas, line 2925: I don't particularly care. But do we want to change this? It seems the graphical artifacts are just because of the selection background. Or am I misunderstanding something?

Lythenas (commented):
This seems fine to me. But I will test it tomorrow.

tiling.js
jtaala: I vote yes!  That was part of the "improves".  I think it looks much much better / smoother (especially with faster `animation-time` parameters).  Plus, it just looks cooler (IMHO) when you (for example) drag a browser tab (e.g. to make it into a new window) and it just expands naturally to the right.  Happy to hear contrary opinions though!

Might be a small change/improvement but [scout's rule!](https://auth0.com/blog/a-scout-approach-to-software-development/).

jtaala:
> Does this also fixes #347?

Not intentionally.  I had a quick look at that one and couldn't quite understand from the video what glitch they were referring to (was it the pdf viewer in the video that looked weird?).

If so, I haven't experienced that issue (i.e. I can't reproduce).  I'll comment on that issue directly.

tiling.js
Lythenas: With this we stretch the window is just one dimension and I think then I notice the stretching more in the text. But I think it is probably fine since most of the time I am only scrolling left to right this makes it a bit more consistent.

Lythenas (approved):

PR #504, opened 2023-04-03
Allow users to set the default focus mode

### LINKED ISSUE #503: Default Focus Mode should be configurable
It seems that PaperWM uses its default focus mode every time I log in. I much prefer the "Center" focus mode rather than the default behavior - it feels really nice on an ultrawide monitor! I believe there should be a setting or a way to easily configure which focus mode PaperWM uses by default. If this is already the case, then maybe we could add some documentation for configuring such a setting.

To be clear, I'm not proposing that we change the behavior of the "Default" or "Center" focus modes, only that we allow configuration of which focus mode is used on init.

jtaala:
Fixes #503 

This PR allows users to set the default focus mode (see #482) that is used when starting a gnome-session (with PaperWM enabled).

This was meant to be part of #482 but I completely forgot about it :disappointed:.  From the `README.md` of this PR:

> ### Setting the default focus mode
> 
> The default focus mode is the standard PaperWM focus mode (i.e. not centered).  This can be changed according to preference by setting the `default-focus-mode` setting via `dconf` or `gsettings`.  
> 
> To set the default focus mode to `CENTER`, execute the following from a terminal:
> ```
> dconf write /org/gnome/shell/extensions/paperwm/default-focus-mode 1
> ```
> 
> To undo to, or revert to the original PaperWM behaviour (by default), execute the following:
> ```
> dconf write /org/gnome/shell/extensions/paperwm/default-focus-mode 0
> ```
> 
> _Note: changing this setting during a PaperWM session will set all spaces to the new default focus mode._

This PR also fixes a few minor issues/annoyances, such as aligning with `id` name for `show-window-position-bar` schema property, and correctly showing enabled options in the `Settings ui` (e.g. they can show enabled but the slider is in the "off" position while it's background shows it's on).


Lythenas (approved):
Looks good to me.

Issue #268, opened 2020-03-13
When minimap is active, other hotkeys are disabled


setzer22:
Hi, first of all, many thanks for this awesome project! Now onto my issue/question:

I find it confusing that when I'm pressing Super+Left/Right (and the minimap is active) I cannot use other hotkeys, like for instance Shift+Super+Left/Right to change monitor. Is this intended or undesired?

Many thanks!

hedning:
Cheers, glad you like it :)

That's indeed not intended when it comes to the change monitor keybindings. Though I foresee some edge cases that might need fixing (eg. what happens when changing monitors while previewing workspaces with eg. `<Super>above_tab`).

In general we aim to make the hotkeys usable in as many scenarios as possible (you can for instance navigate, remove windows and spawn new windows while flipping through workspaces). There's some actions we haven't figured out how to integrate, mainly `<Super>Tab`, as we're not sure how to make the transition from linear to time based navigation make sense. 

For hotkeys that aren't ours the problem is unfortunately harder. Starting navigation (eg. with `<super>left`) starts a mini mode (like `<Super>Tab` does in plain gnome shell), where we need to take full control of the keyboard. And for some reason it isn't possible to just ask gnome shell what handler is naturally associated with a hotkey and then run it. So for native bindings we need to set up that manually per hotkey/action (this is done for the minimize action).

olejorgenb:
If anyone wants to take a crack at this:
- Download gnome-shell source code: https://gitlab.gnome.org/GNOME/gnome-shell
- Download mutter source code: https://gitlab.gnome.org/GNOME/mutter
- Find the mutter action name of the action you want to implement support for (usually possible by looking here `dconf-editor /org/gnome/desktop/wm/keybindings/` - the settings keyname is the action name)
- Find out how this action is implemented: Search the gnome-shell and mutter code for the action name. (`setCustomKeybindingHandler` in gnome-shell)
- Figure out how to trigger the action programatically
- If the action comes from mutter it's action id is listed in Meta.`KeyBindingAction`
- If the action comes from gnome-shell, use the `Keybindings.getActionIdByActionName` method to get the actionId
- See https://github.com/paperwm/PaperWM/blob/d47206cc017df0a8ce63f2612d25c1db7d609267/navigator.js#L168 for how minimize is done

hedning:
[...]

PR #510, opened 2023-04-07
Add option to allow all keybindings in the navigator

Lythenas:
This allows executing all actions while the "navigator" (minimap shown while switching windows) is open. If the action would not be allowed before, it will close the navigator and then execute the action.

This works only for the paperwm minimap that is shown when pressing e.g. `<Super>+Left`. Not for the minimap that is shown by gnome-shell (e.g. `<Alt>+<Tab>`).

Also unfortunately in my testing it only works for keybinding that have been registered with paperwm and not for arbitrary gnome-shell keybindings (see also #268 for this).

Also after doing e.g. `<Super>+Left` and then switching monitors with `<Super>+<Shift>+Left` I get the following warning in the logs:

```
Window manager warning: META_CURRENT_TIME used to choose focus window; focus window may not be correct.
```

But apart from the warning everything works as expected. I also found this issue about the warning: https://gitlab.gnome.org/GNOME/mutter/-/issues/1066

Example usage that was annoying before: Navigate to a window and move it to the other monitor. I.e. `<Super>+Left` followed by `<Super>+<Ctrl>+Left`. This now works without letting go of `<Super>`.


jtaala:
Cool! - any tips for how you use this? (e.g. what actions do you like doing while minimap is open just so I can test them).

schemas/org.gnome.shell.extensions.paperwm.gschema.xml
```diff
@@ -393,6 +393,11 @@
         <summary>Show the focus mode icon in the topbar</summary>
     </key>
 
+    <key type="b" name="allow-all-actions-in-navigator">
```
jtaala: Do we need this to be an user settable option (with a default of false)?

Is there a reason why we wouldn't want this behaviour by default?  e.g. I can't think of a reason why would not want other shortcuts executable while minimap is open.

jtaala (commented):
Looks good - see comment about just having this as default behaviour.

Re the warning - it's a warning and doesn't appear to have any other negative impacts and action appears to execute correctly :man_shrugging:

jtaala:
> Cool! - any tips for how you use this? (e.g. what actions do you like doing while minimap is open just so I can test them).

Never mind - saw an example in the code (switching monitors works fine).

schemas/org.gnome.shell.extensions.paperwm.gschema.xml
Lythenas: I don't know the original reason for separating the keybindings but IMO we can also just change the behavior and not create a user setting for this.

The only breakage I can think of (not tested) is that I think it would be possible (in user.js) to create a keybinding that is only active in the navigator but has the same keys as another keybinding. But I think you can work around that because the handler get passed `navigator` and I think you also globally check if the navigator is open in your custom handler.

jtaala: Agree, would make this default behaviour.  Test out your hypothesis as much as you can - but with `user.js`, users can implement unintended things/behaviours (including code that would completely break PaperWM) so  I wouldn't let concerns about users potentially implementing something in there stop us from simplifying, testing, and improving PaperWM with this PR.

Make changes you see fit (e.g. making this default) and I can help with testing this.

Lythenas: I removed the option and just made this the new behavior.

Lythenas:
Unfortunately I haven't found a way to activate arbitrary gnome actions. In my testing only other PaperWM keybindings worked. If you have any ideas here I can investigate further. Otherwise I think this is already a good improvement on the current behavior.

jtaala (approved):
Approved.

jtaala:
One thing I noticed is that cycling through window widths while minmap is showing  (I have incrementing window widths set to `ALT+E`)  causes the minimap to show some weird empty space where the window portion was (before resize) - that's not an issue with this functionality though (the fact that we can now do this while navigator is showing proves this works).

I might try to fix navigator a bit here.

jtaala:
> I might try to fix navigator a bit here.

Hey @Lythenas, so I just pushed a fix to the issue I mentioned.  Basically it just resets/refreshes the minimap when windows resize (if minimap is showing that is).  Just means if minimap will be shown correctly after a window resize event

I'm ready to merge this now once you approve.

Cheers.

Lythenas:
Good catch. I haven't tried that particular shortcut. Your fix works perfectly. Feel free to merge, unfortunately I still can't merge.

jtaala:
> unfortunately I still can't merge.

You should be able to merge now.  Give it a go when ready.  Let me know if still have issues.

jtaala:
> > unfortunately I still can't merge.
> 
> You should be able to merge now. Give it a go when ready. Let me know if still have issues.

After you approve my last push that is.

Lythenas:
> After you approve my last push that is.

Ah, I see. Because you made the last commit, you also can't approve. Is that correct?

Unfortunately Github does not allow me to approve my own PR... So @smichel17 pls help^^

jtaala:
> Unfortunately Github does not allow me to approve my own PR... So @smichel17 pls help^^

All good, I can push it.

PR #512, opened 2023-04-08
Also print if on wayland or Xorg in gather-system-info.sh

Lythenas:
This will also print a line:

```
Display server: Wayland
```

or

```
Display server: Xorg
```

depending on what the user is using.

---

Also regarding gnome 3.38 and the `gnome-extensions` command not being available: I found two fallbacks we could use:

1. `ls ~/.local/share/gnome-shell/extensions/` (lists also extensions that might not be enabled)
2. `gsettings get org.gnome.shell enabled-extensions` (list also system extensions)

I think I would prefer option 2. And maybe we would also want to print all extensions anyway and not just user extensions (as distro specific extensions are currently not visible).

### DIFF
--- a/gather-system-info.sh
+++ b/gather-system-info.sh
@@ -20,6 +20,13 @@ fi
 
 gnome-shell --version
 
+echo -n "Display server: "
+if [ "${XDG_SESSION_TYPE}" = "wayland" ]; then
+    echo "Wayland"
+else
+    echo "Xorg"
+fi
+
 echo -n "PaperWM branch/tag: "
 git symbolic-ref --short -q HEAD || git name-rev --tags --name-only --no-undefined "$(git rev-parse HEAD)"
 echo -n "PaperWM commit: "
--- a/gather-system-info.sh
+++ b/gather-system-info.sh
@@ -67,20 +67,17 @@ show_gnome_extensions() {
         # compare enabled extensions to installed extensions
         # because some uninstalled extensions could still be enabled because of
         # stale gsettings
-        local enabled=$(_enabled_extensions)
-        local installed=$(_installed_extensions)
-
-        for ext in $enabled; do
-            if [[ "${installed}" =~ "${ext}" ]]; then
+        for ext in $(_enabled_extensions); do
+            if is_extension_installed "$ext"; then
                 echo "- $ext"
             fi
         done
     fi
 }
 
-_installed_extensions() {
-    ls ~/.local/share/gnome-shell/extensions/
-    ls /usr/share/gnome-shell/extensions/
+is_extension_installed() {
+    local ext=$1
+    [[ -d "$HOME/.local/share/gnome-shell/extensions/$ext" ]] || [[ -d "/usr/share/gnome-shell/extensions/$ext" ]]
 }
 
 _enabled_extensions() {
@@ -89,7 +86,7 @@ _enabled_extensions() {
     # $matches contains lines with the extension uuid and lines with ", "
     mapfile -t matches < <(_global_rematch "$s" "'([^']*)'")
     for match in "${matches[@]}"; do
-        if [[ "$match" =~ "^," ]]; then
+        if [[ "$match" =~ ^,\s* ]]; then
             continue
         fi
         echo "$match"
--- a/gather-system-info.sh
+++ b/gather-system-info.sh
@@ -9,36 +9,99 @@
 # - awk
 
 REPO="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
-cd "${REPO}"
 
-echo "Please include this information in your bug report on GitHub!"
+main() {
+    cd "${REPO}"
 
-echo -n "Distribution: "
-if [ -f /etc/os-release ]; then
-    source /etc/os-release && echo "${NAME}"
-fi
+    echo "Please include this information in your bug report on GitHub!"
 
-gnome-shell --version
+    show_distribution
+    show_gnome_version
+    show_display_server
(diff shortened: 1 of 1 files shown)


jtaala (approved):
Excellent!  tested in both X11 and Wayland.

jtaala:
> I think I would prefer option 2. And maybe we would also want to print all extensions anyway and not just user extensions (as distro specific extensions are currently not visible).

I agree.  Are you thinking to only use `gsettings` for a gnome 3.38 branch or switch to using `gsettings` regardless (would prefer to have consistency if possible).

Up to you, implement what you see fit and I'll test again and merge when you're ready.

Lythenas:
The problem with the `gsettings` approach I've noticed is that is can include uninstalled extensions too. E.g. for me this prints

```
$ gsettings get org.gnome.shell enabled-extensions
['nohotcorner@azuri.free.fr', 'appindicatorsupport@rgcjonas.gmail.com', 'TaskBar@zpydr', 'topicons-redux@pop-planet.info', 'pop-shell@system76.com', 'switcher@landau.fi', 'gnome-shell-mode@hedning:matrix.org', 'paperwm@paperwm-redux.github.com', 'paperwm@hedning:matrix.org']
```

But the only extensions I have actually installed and enabled are appindicator, switcher and paperwm@hedning:matrix.org (also no other builtin extensions). And some of the extensions aren't even installed anymore.

I will try to find a way to only print actually enabled extensions.

jtaala:
> I will try to find a way to only print actually enabled extensions.

Give it a go - if it's too finnicky, it's fine to leave the current one using `gnome-extensions` (since it's working well!) - for the `gnome-3.38` branch, could use the `ls` method?

Lythenas:
The implementation is a bit hacky, so please test it. It seems to work well on gnome 43. But I haven't tested it on older gnome versions.

@fumbles if you don't mind, could you test this updated script on gnome 3.32?

jtaala (approved):
Works well.  Good job.

jtaala:
@Lythenas, do you want me to give [@fumbles](https://github.com/fumbles) a couple of days to test in 3.32 - or should merge this PR into `develop` now (which I'm happy to do).

Lythenas:
Maybe wait a couple of days. But you can merge it when you want to do a release.

fumbles:
Sorry @Lythenas  @jtaala  I had forgotten about this :sweat:  Everything has been working with no issues. About to upgrade to fedora 38 soonish since it's approved so we'll see how that goes.

From: [Ubuntu forum](https://askubuntu.com/questions/1133782/command-to-list-installed-and-enabled-gnome-extensions)

Your locally installed gnome Shell extensions, i.e., these that are installed for your user only, can be listed with the command:

`ls ~/.local/share/gnome-shell/extensions/`

You can find out which extensions are enables by querying a dconf setting:
`gsettings get org.gnome.shell enabled-extensions`

System wide installed gnome-shell extensions are listed with the command
`ls /usr/share/gnome-shell/extensions/`


`gsettings get org.gnome.shell enabled-extensions | sed 's/,/\n/g'`
```
['background-logo@fedorahosted.org'
 'places-menu@gnome-shell-extensions.gcampax.github.com'
 'apps-menu@gnome-shell-extensions.gcampax.github.com'
 'appindicatorsupport@rgcjonas.gmail.com'
 'caffeine@patapon.info'
 'drive-menu@gnome-shell-extensions.gcampax.github.com'
 'dash-to-dock@micxgx.gmail.com'
 'clipboard-indicator@tudmotu.com'
 'utcclock@injcristianrojas.github.com'
 'paperwm@hedning:matrix.org']
```

Lythenas:
Yes that is more or less what I'm doing. I'm listing all enabled extensions and then checking if they are actually installed. This was necessary on my machine because I guess I "uninstalled" some extension wrong (i.e. deleted folders). E.g. when switching from the paperwm-redux fork back to paperwm.

@jtaala I simplified the script a bit. But I think you can merge it after reviewing.

jtaala (approved):
Tested and works well.  Good work!

gather-system-info.sh
```diff
@@ -20,6 +20,13 @@ fi
 
 gnome-shell --version
 
+echo -n "Display server: "
+if [ "${XDG_SESSION_TYPE}" = "wayland" ]; then
```
smichel17: Just want to share one of my preferences for shell scripting, not asking for a change:
```suggestion
if test "${XDG_SESSION_TYPE}" = "wayland"; then
```
I really like the educational benefit of using `test` rather than single brackets.

- Helps build an intuition for "`if` just switches on the status code of the command" so it doesn't become a separate concept for people to understand when you want to use a non-builtin there.
- Possible to `man test` vs having to do a web search for how brackets work in [ba]sh
- Things like `-gt` being flags actually makes sense.
- Not significantly more verbose.

Lythenas: Actually it is better to use `[[` instead of `[`. Because `[` is a synonym for `test` and therefore a command. And it is therefore also evaluated like a command, which has in some cases subtle, inconsistent or surprising behavior. `[[` is a keyword and therefore everything in `[[ ... ]]` is treated as "one thing" and evaluated in a more consistent way. Usually there is no difference at all. But in cases there is, it is usually not at all obvious what is going on.

Also using `[[` you can also use `<`, `>`, `=~` (for regex), `==` and `!=`. So in that sense it behaves more like other languages.

For most things bash I usually just refer to https://devhints.io/bash Since I can't remember everything and I think the syntax is not very intuitive. But you can also use `man bash` and search for `CONDITIONAL EXPRESSIONS`.

Issue #433, opened 2022-10-29
Alternative layout suggestion for workspace indicator


jonseppanen:
I personally use things like: https://extensions.gnome.org/extension/3968/improved-workspace-indicator/ 
or even: https://extensions.gnome.org/extension/3952/workspace-indicator/
This lets me quickly switch workspaces with a click.
I can use these plugins at the moment with PaperWM, but I have to disable the code block:
`Main.panel.addToStatusArea('WorkspaceMenu', menu, 0, 'left');
    menu.show()`
Inside the topbar.js file, as I then no longer need this.
I guess I'm asking for either a way to disable the PaperWM workspace indicator in case of using another, or improve the paperWM one so it shows all workspaces and has clickable buttons.

jtaala:
Thanks for paitence @jonseppanen, closing this as a duplicate given #387.  

P.S. I'll do up a PR in the next week or so to be able to disable the workspace indicator like we disable  other topbar elements like in later PRs (e.g. #476, #482, #499).

PR #513, opened 2023-04-09
Adds option to disable the PaperWM topbar workspace name indicator/button

### LINKED ISSUE #387: Add option to entirely disable the workspace name / activities button
PaperWM usually has a workspace name in the top bar, along with workspace names baked into each workspace's background. I have a different wallpaper per workspace so I don't need any of these identifiers. I found a way to disable the workspace names, but unfortunately that switches the workspace name to "activites" akin to the stock gnome button.

However, this button is added by PaperWM, so normal extensions to remove this button don't work. I would love a way to completely disable this button, because I never use it and it's just sitting there being unused.

image

jtaala:
This PR adds a user-settable option which disables the workspace indicator.

Fixes #387 #433.

From this PR's changes to `README.md`:

> If you prefer to use another workspace indicator (or prefer none at all), you can hide this workspace name element from Gnome topbar by executing the following command from a terminal:
> 
> ```
> dconf write /org/gnome/shell/extensions/paperwm/show-workspace-indicator false
> ```


topbar.js
```diff
@@ -799,16 +806,34 @@ function fixTopBar() {
     }
 }
 
+function fixWorkspaceIndicator() {
+    prefs.show_workspace_indicator ? menu.show() : menu.hide();
```
Lythenas, line 810: This should just be an `if`, right?

topbar.js
```diff
@@ -799,16 +806,34 @@ function fixTopBar() {
...
     prefs.show_focus_mode_icon ? focusButton.show() : focusButton.hide();
     Tiling.spaces.forEach(s => s.showFocusModeIcon());
 }
 
+/**
+ * Sets the space focusModeIcon position based off the position
+ * of the focusButton (mimics it).
+ * @param {Number} timeout needed to allow update before querying position.
+ */
+function setSpaceFocusModeIconPositions(timeout=0) {
```
Lythenas: Do we need a specific timeout here? Or does `0` always work?

topbar.js
jtaala: well, that's an inline if.  But yes, it's an if.

topbar.js
jtaala: default is 0, but a small timeout is needed when workspace indicator emits `show` and `hide` signals.

See https://github.com/paperwm/PaperWM/pull/513/files#diff-b0cab10735fabfe8a6e4455997859f505dbbe12a813ba3227c2952ba5a89ecddR671-R672

jtaala: Yes, `0` always works, hence set it as default.

We also could do something like

```
function setSpaceFocusModeIconPositions(timeout) {
    timeout = timeout ?? 0;
    ...
}
```
Which is kind of equivalent (but maybe preferred since it would also cover the case where a developer might do a `setSpaceFocusModeIconPositions(undefined)` or `setSpaceFocusModeIconPositions(null)`) - I can change that one.

topbar.js
jtaala: I might have misunderstood your question(?).  It needs to be an `if-else`, since if we change the `show-workspace-indicator` option to false, we want it to `hide()`.

Hence the inline if(else) statement.

Let me know if that didn't answer the question.

Lythenas (approved):

jtaala:
Thanks @Lythenas for approving!

Note, I've just updated `README.md` and added this property to the new properties section from #514.  Please approve again when you have a chance and I'll merge this in. 

Thanks!

ChaiAK:
Hi, what happened to this, it's seemingly no longer a feature or in the readme?

jtaala:
@ChaiAK 

image

Enable that option to replace (and disable) workspace name indicator with standard gnome pill.

jtaala:
It's also in the readme:
https://github.com/paperwm/PaperWM?tab=readme-ov-file#the-workspace-stack--monitors

image

Issue #516, opened 2023-04-22
Gnome 44: PaperWM keyboard shortcuts clash with gnome default `super+left/right`?


jtaala:
I don't usually use `super+left/right` to switch to left/right window, but on my test setup for Gnome 44 I notice that sometimes when logging in the PaperWM default switch to left/right sometimes doesn't work - or on `super+left` it moves the window.

I've disabled the default `Super+left/right` gnome shortcut (which tiles the window left/right) and the issue doesn't appear to be happening now.  Not sure if that was the problem.

Anyone else on Gnome 44 also seeing this?

**To Reproduce**
Unfortunately I can't get this to reproduce consistently.   It happened periodically after logging in (didn't seem very consistent).  I also think I saw it happen after a lock-screen.

**Expected behavior**
PaperWM keyboard shortcut should always work.

**System information:**

```
Distribution: Fedora Linux 38
GNOME Shell 44.0
Display server: Wayland
PaperWM branch/tag: fixes-gnome-44-support
PaperWM commit: 56023a79c5f827749de2bd868208570553b60cca
Enabled extensions:
- paperwm@hedning:matrix.org
```

Lythenas:
I think we can just remove the conflicting shortcuts in the `set-recommended-gnome-shell-settings.sh` and document it with the other recommended settings.

fumbles:
My Super + Left/Right work fine but Shift+Super+< does not for me even after disabling everything I can find in the keyboard shortcut stuff.
 I pulled the fixes-gnome-44-support commit into my local develop branch..hence it says develop branch below.


```Distribution: Fedora Linux
GNOME Shell 44.0
Display server: Wayland
PaperWM branch/tag: develop
PaperWM commit: 56023a79c5f827749de2bd868208570553b60cca
Enabled extensions:
- paperwm@hedning:matrix.org
- drive-menu@gnome-shell-extensions.gcampax.github.com
- dash-to-dock@micxgx.gmail.com
- clipboard-indicator@tudmotu.com
- utcclock@injcristianrojas.github.com
- apps-menu@gnome-shell-extensions.gcampax.github.com
- background-logo@fedorahosted.org
- places-menu@gnome-shell-extensions.gcampax.github.com
- appindicatorsupport@rgcjonas.gmail.com
- caffeine@patapon.info
```

crispinb:
[...]

PR #518, opened 2023-04-30
Remove Gnome tiling left/right shortcuts

tlevani:
Based on the discussion in #516 I added gnome 44 shortcut removal using the same `set-with-backup` bash function used for setting other recommended gnome-shell settings. I haven't researched if there are any unwanted consequences of disabling these shortcuts like this, but it seems to work fine for me. It would be great if someone else with this issue could also test this before merging. Also, the backup `restore-gnome-shell-settings-YYYY-MM-DD.sh` file was successfully created with the original values stored inside.

### DIFF
--- a/README.md
+++ b/README.md
@@ -432,6 +432,8 @@ There's a few Gnome Shell settings which works poorly with PaperWM. Namely
   spanning all monitors
 - `edge-tiling`: We don't support the native half tiled windows
 - `attach-modal-dialogs`: Attached modal dialogs can cause visual glitching
+- `toggle-tiled-left`: Default GNOME keyboard shortcut `super+left` collides with a default PaperWM shortcut. We disable the GNOME shortcut
+- `toggle-tiled-right`: Default GNOME keyboard shortcut `super+right` collides with a default PaperWM shortcut. We disable the GNOME shortcut
 
 To use the recommended settings run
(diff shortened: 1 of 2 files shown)


Lythenas (changes requested):
I think setting the keybinds to empty is fine for the recommended settings.

Can you also add these settings to the list in the README? Under "Recommended Gnome Shell Settings". And mention that the default gnome keybindings clash with the default PaperWM keybindings.

jtaala:
Looks great @tlevani!

Question (for everyone), got a bit confused in my testing from `set-recommended-gnome-shell-settings.sh` [not applying settings if a restore file exists](https://github.com/paperwm/PaperWM/pull/518/files#diff-1f2ed945f06168492ef47f0468cd0c7b714af1b3962be302993b36bd44e4bedfL5-L8).

Basically, I ran the `set-recommended-gnome-shell-settings.sh` first to test and it worked fine and disabled the keybind clash  Then I reset the keybinds manually (in dconf) and ran the `install.sh` script and selected `y` for the recommended settings - but this time it didn't fix the keybind clash because a restore file already existed.

What do you think?  remove those lines so that whenever we run it it resets the settings?  To avoid overwriting original settings (with the PaperWM settings) when we rerun, we could just create a restore file IF a previous restore file doesn't exist.

Thoughts?

Lythenas:
Ah interesting. @jtaala I will open a separate issue for that since it doesn't really have anything to do with this PR: #520

Lythenas (approved):
Looks good to me now.

Issue #523, opened 2023-05-15
GNOME 44: `Super + Shift + Above_Tab` works in the forward direction when it should work in the backward direction


letavocado:
**Describe the bug**
<kbd>Super</kbd> <kbd> Shift</kbd>  <kbd>Above_Tab</kbd> works in the forward direction when it should work in the backward direction. Thus making no difference between the behavior of <kbd>Super</kbd> <kbd>Above_Tab</kbd> and <kbd>Super</kbd> <kbd> Shift</kbd>  <kbd>Above_Tab</kbd>

While <kbd>Super</kbd> <kbd> Ctrl</kbd> <kbd> Shift</kbd>  <kbd>Above_Tab</kbd> works as expected in backward cycle direction. Now the behavior works correctly, but sometimes the bug occurs.

**To Reproduce**
1. Switch to [fixes-gnome-44-support](https://github.com/paperwm/PaperWM/tree/fixes-gnome-44-support)
2. run './install.sh'
3. do a login/logout (maybe twice, if bug not occurred)
4. check behavior of described bug 

**Expected behavior**
<kbd>Super</kbd> <kbd> Shift</kbd>  <kbd>Above_Tab</kbd> should work in the backward direction 

**System information:**
`./gather-system-info.sh` output:

Distribution: Arch Linux
GNOME Shell 44.1
Display server: Wayland
PaperWM branch/tag: fixes-gnome-44-support
PaperWM commit: 477d546e5a78280cb324379a365225b0f702ad8d
Enabled extensions:
- paperwm@hedning:matrix.org
- clipboard-history@alexsaveau.dev
- blur-my-shell@aunetx
- gnome-ui-tune@itstime.tech
- Vitals@CoreCoding.com
- trayIconsReloaded@selfmade.pl
- switcher@landau.fi
- dash-to-dock@micxgx.gmail.com

**Additional context**
<kbd>Super</kbd> <kbd> Ctrl</kbd> <kbd> Shift</kbd>  <kbd>Above_Tab</kbd> works as expected in backward cycle direction. Now the behavior works correctly, but sometimes the bug occurs.

jtaala:
Thanks @letavocado, haven't been able to reproduce as yet - but will try again later on.

[...]

PR #525, opened 2023-05-18
Disable known conflicting keybindings

Lythenas:
See #523 for discussion.

@jtaala please check that I got all keybindings.


jtaala:
Thanks @Lythenas - however, I now believe these aren't needed anymore!

See https://github.com/paperwm/PaperWM/issues/523#issuecomment-1553850521.

I believe I've now found/fixed the root cause of why PaperWM's keybind conflict detection and overriding wasn't working (it should have picked these clashes up without needed them to be disabled).

PR #529, opened 2023-05-22
add left-handed mode

KiaraGrouwstra:


jtaala:
Hey @KiaraGrouwstra!  Thanks for the PR!  

I use similar keybinds (except more of a left-handed version of hjkl) so I know where you're coming from.  However, PaperWM includes a decent/simple way to add custom keybinds via the PaperWM settings, and doesn't require changing the schema and adding these keybinds as defaults.

I know it's probably preference here, but I'd prefer to keep a simple/default set of keybinds and then allow users to customise these via the settings preferences.

I'll leave this up though, and see if others chime in if they think it's worth adding this set of default keybinds.

Lythenas:
I agree with @jtaala I think we shouldn't add many more default keybindings. New keybindings could also conflict with other keybindings that the user has manually set. And I would like to avoid annoying existing users without a good reason.

There is also a wiki page about alternative keybindings: https://github.com/paperwm/PaperWM/wiki/Alternative-keybindings I think it would be fine to add the keybindings there and maybe we can make this page a bit easier to find. Since I also didn't know it existed before searching for it because of this PR.

jtaala:
Thanks @KiaraGrouwstra and @Lythenas.  Will close this PR.

@KiaraGrouwstra - let us know if we can help add these keybinds to https://github.com/paperwm/PaperWM/wiki/Alternative-keybindings.

PR #532, opened 2023-05-27
Generate restore script with second timestamp

### LINKED ISSUE #520: `set-recommended-gnome-shell-settings.sh` does not work if it was previously ran on the same day
**Describe the bug**
`set-recommended-gnome-shell-settings.sh` does not work if it was previously ran on the same day.

**To Reproduce**
1. run `set-recommended-gnome-shell-settings.sh`
2. Manually reset or change the gnome settings
3. run the script again
4. Recommended settings are not applied

**Expected behavior**
The script should always apply the recommended settings..

**System information:**

```
Please include this information in your bug report on GitHub!
Distribution: Arch Linux
GNOME Shell 43.5
Display server: Wayland
PaperWM branch/tag: gather-system-info-update
PaperWM commit: 1ab44cddc3778276039a59b9316c87f8cb3fbcea
Enabled extensions:
- paperwm@hedning:matrix.org
- switcher@landau.fi
```

**Additional context**

This is cause by the following lines:
https://github.com/paperwm/PaperWM/blob/1130b663806d6daba1c8355c4b99c287d237e918/set-recommended-gnome-shell-settings.sh#L3-L8

So you only run into this behavior if you run the script multiple times on the same day.

IMO we can just remove this check and add the time to the restore file name.

I can create a PR for this once we decide what to do.

Lythenas:
Fixes #520 

When executing the `set-recommended-settings.sh` it will now generate a file with a name like `restore-gnome-shell-settings-2023-05-27T17:11:49+02:00.sh`.

I chose second precision because that seemed sufficient and the `--iso-8601` flag has no millisecond precision only nanoseconds and that seemed excessive.

### DIFF
--- a/set-recommended-gnome-shell-settings.sh
+++ b/set-recommended-gnome-shell-settings.sh
@@ -1,9 +1,9 @@
 #!/usr/bin/env bash
 
-RESTORE_SETTINGS_SCRIPT="restore-gnome-shell-settings-$(date +%F).sh"
+RESTORE_SETTINGS_SCRIPT="restore-gnome-shell-settings-$(date --iso-8601=s).sh"
 
-if [[ -e $RESTORE_SETTINGS_SCRIPT ]]; then
-    echo "$RESTORE_SETTINGS_SCRIPT exists"
+if [[ -e "${RESTORE_SETTINGS_SCRIPT}" ]]; then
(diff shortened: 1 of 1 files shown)


jtaala (approved):
Tested and looks good!

PR #515, opened 2023-04-22
Fixes: add Gnome 44 support

jtaala:
This PR fixes reported issues in Gnome 44: fixes #494, fixes #495, closes #490, fixes #523. 

Note: these are compatibility fixes to support Gnome 44 in develop branch (i.e. Gnome 43 should still work fine with these changes).

Gnome 44 has a new (improved?) mechanism for grabs (e.g. grabbing and moving windows), and no longer uses `end_grab_op`.

This PR also adds a polyfill to support Gnome 44's removal of `Meta.later_add` (which is replaced with `global.compositor.get_laters().add(...)`.  

Note: [GJS porting guide](https://gjs.guide/extensions/upgrading/gnome-shell-44.html#meta-later-add-and-meta-later-remove) suggests using `Meta.Laters.add` but in testing the latest Gnome 44 (and what was released in Fedora 38), this was not present.  The implemented approach is used heavily in the latest/official [gnome-shell source code](https://gitlab.gnome.org/search?search=global.compositor.get_laters()&nav_source=navbar&project_id=546&group_id=8&search_code=true&repository_ref=main) (which is why I moved to using that).

~~Note<sup>2</sup>: for the first fix (`end_grab_op`) - it still not quite perfect, it seems to require a second click to activate a window (after moving a window) - but at least it doesn't completely/irreversibly lose mouse click control. 
 Will look at this and see how to improve here.~~  _**fix now implemented for this**_  :heavy_check_mark:

PR #537, opened 2023-05-29
Fix conflicting keybindings on Gnome 43

### LINKED ISSUE #533: Conflicting keybindings are not working in gnome 43
**Describe the bug**
Conflicting keybindings (e.g. Super+Shift+Left/Right to navigate monitors) are not working if they are also bound in gnome itself.

**To Reproduce**
1. Press a conflicting keybinding (e.g. Super+Shift+Right)
2. Nothing happens

Note that not even the action it is bound to in gnome is executed.

Removing the conflicting keybinding in gnome fixes this issue.

**Expected behavior**
PaperWM keybindings should take precedence.

**Screenshots**

**System information:**

```
Distribution: Arch Linux
GNOME Shell 43.5
Display server: Wayland
PaperWM branch/tag: restore-faster
PaperWM commit: 0b73c0b1b0a41b6113fdf9f3d85ca8269a7c2157
Enabled extensions:
- paperwm@hedning:matrix.org
- switcher@landau.fi
```

**Additional context**
Note that the commit is from PR #532 which is one commit ahead of `develop`.

@jtaala you recently made changes to fix the issue of conflicting keybindings on Gnome 44. Could this have broken the behavior on Gnome 43?

Also note that I am now updating to Gnome 44 and now have only a machine with Gnome 44 and Gnome 42 to test. But I wanted to put this up. In case it is easy to do, I think we should fix this. Otherwise, I think the gnome43 branch works fine.

Lythenas:
Revert the one commit that fixed conflicting keybindings on Gnome 44 from #515 

Fixes #533 

Notes regarding testing: You need to relogin after you enable conflicting keybindings. They don't seem to be picked up without relogin if you reset your keybindings in the Gnome settings.

### DIFF
--- a/settings.js
+++ b/settings.js
@@ -242,7 +242,7 @@ function keystrToKeycombo(keystr) {
         keystr = keystr.replace('Above_Tab', 'a');
         aboveTab = true;
     }
-    let [ok, key, mask] = Gtk.accelerator_parse(keystr);
+    let [key, mask] = Gtk.accelerator_parse(keystr);
 
     if (aboveTab)
         key = META_KEY_ABOVE_TAB;
--- a/keybindings.js
+++ b/keybindings.js
@@ -574,8 +574,32 @@ function overrideAction(mutterName, action) {
     actionIdMap[id] = action;
 }
 
+/**
+ * Disables known keybind clashes with default gnome operations.
+ */
+function disableKnownClashes() {
+    // disable specific actions known to cause issues
+    const mutterbinds = convenience.getSettings('org.gnome.mutter.keybindings');
+    [
+        'toggle-tiled-left', 'toggle-tiled-right'
+    ].forEach(key => {
+        try {
+            mutterbinds.set_value(key, new GLib.Variant('as', []));
+        } catch (e) {}
+    });
+
+    const wmbinds = convenience.getSettings('org.gnome.desktop.wm.keybindings');
+    [
+        'switch-group', 'switch-group-backward',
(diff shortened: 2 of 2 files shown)


settings.js
```diff
@@ -242,7 +242,7 @@ function keystrToKeycombo(keystr) {
         keystr = keystr.replace('Above_Tab', 'a');
         aboveTab = true;
     }
-    let [ok, key, mask] = Gtk.accelerator_parse(keystr);
+    let [key, mask] = Gtk.accelerator_parse(keystr);
```
jtaala, line 245: Thanks @Lythenas!

This should be the only change (revert) needed to fix gnome 43, 42 keybind regressions.

jtaala (changes requested):

jtaala:
Thanks @Lythenas!

Unfortunately this PR  also adds code that was never in versions prior to Gnome 44 (and never intended to be in previous versions) - and was only implemented as a temporary workaround since keybind clash detection in Gnome 44 was broken at the time (since fixed).

The only line/reversion needed in this PR [is this line](https://github.com/Lythenas/PaperWM/blob/c07add165b586998d1972d08964c9a45fc03c2f8/settings.js#L245).

Closing this PR in favour of #539 since that one will implement a fix that is backwards compatible and works in Gnome 44, 43, 42.

_Note: I'm also planning on adding changes in #539 to develop branch - hopefully bringing us closer to `develop` branch working just as well in prior version of gnome._

PR #542, opened 2023-06-03
Updated README.md and combined gnome versions targeted for current support.

jtaala:
After some extended testing of the `develop` branch in both gnome 43 and gnome 42, we've updated README.md to state that gnome versions 42-44 should use `develop` branch.

This should simplify things for users of recent versions of Gnome, and make things easier to support (i.e. easier to support and focus testing on one branch rather than 3).

This PR updates `README.md` to reflect this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Clone the repo and check out the branch supporting the Gnome Shell version you'r
 - 3.28-3.38: https://github.com/paperwm/PaperWM/tree/gnome-3.38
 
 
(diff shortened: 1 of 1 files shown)


jtaala:
@Lythenas, can you check this out and see if the changes explain versions to use well enough?  Thanks.

Lythenas (approved):
I think the wording is fine. And since the develop branch is the default one anyway, I expect no problem for most people anyway.

PR #549, opened 2023-06-13
Cleanup: remove unneeded files

jtaala:
This PR is a bit of housekeeping - remove two unneeded files:

1. `.dir-locals.el`: emacs config file?
2. `yarn.lock`: yarn dependency management file

For (2) not sure why it's here (maybe it was an an attempt to package PaperWM up as an npm package?).  Not sure how that works given it's dependency on Gnome etc.

The other reason I want to remove `yarn.lock` is that our github repo gets flagged sometimes since some dependencies in `yarn.lock` are old and subject to vulnerabilities etc.


Lythenas (approved):
This seems fine to me since we are not really using yarn or npm. I'm guessing it might be there so you can use a JS LSP in your editor, but not sure.

Also, yes .dir-locals.el is for emacs.

PR #554, opened 2023-06-23
Preparation to move default branch from `develop` to `release` branch

jtaala:
I'm thinking it's about time to move from "an experimental extension"  and having the `develop` branch as default, to establishing a `release` branch and have that as the default branch in PaperWM.

A few reasons would be:

- we're starting to follow a more consistent release approach;
- establishing a release branch would provide users that just want a more stable/tested experience with PaperWM (users who want to be on the "bleeding edge" would use the `develop` branch);
- would allow more integration testing of multiple features before releases.  Currently we merge new features into `develop` quite often - we would do releases less often, and only after some longer testing period;

This PR the `README.md` to remove the "experimental extension" reference and recommend most(?) users use the `release` branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,11 +2,14 @@
 
 [![project chat](https://img.shields.io/badge/zulip-join_chat-brightgreen.svg)](https://paperwm.zulipchat.com)
 
+> ⚠️ **ATTENTION:** 
+> _we've recently set `release` branch as default.  This change aims to provide a more tested and stable experience for most users.  You can continue to use `develop` branch if you want the latest features and fixes as soon as they're available.  Changes will be merged into `release` branch after a week or so of testing in `develop` branch._
+
 PaperWM is a [Gnome Shell](https://wiki.gnome.org/Projects/GnomeShell) extension which provides scrollable tiling of windows and per monitor workspaces. It's inspired by paper notebooks and tiling window managers.
 
 Supports Gnome Shell from 3.28 to 44 on X11 and wayland.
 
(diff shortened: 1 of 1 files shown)


Lythenas:
Can you maybe add a "contributing" section (or something similar) that explicitly states that people should open PRs to the develop branch?

Otherwise I think this looks good.

jtaala:
> Can you maybe add a "contributing" section (or something similar) that explicitly states that people should open PRs to the develop branch?
> 
> Otherwise I think this looks good.

Oh good point, didn't think of that... I'm surprised GitHub doesn't allow setting a specific default branch for PRs (as opposed to a default branch for the repo).  

Hmmm - know of any way to do that?  Would this change of default branch cause more issues?  Indeed we would need to add a message re using the `develop` branch for PRs. 

I'm intrigued by your idea of changing the install script too - we could have two like `install-latest-release.sh` and an `install-this-branch.sh` (names may not be great but you get the idea) - with the latter symlinking?

Lythenas:
I believe changing the install script might also help with some cases where users do `git pull` and something does not work. E.g. we could check if the gnome version is compatible in the install script.

Lythenas:
> Hmmm - know of any way to do that? Would this change of default branch cause more issues? Indeed we would need to add a message re using the `develop` branch for PRs.

I don't know it would cause any problems. But everyone that already has a clone will of course continue to use the old branch until they change it manually.

I think it's fine as long as we document it properly.

PR #556, opened 2023-07-04
Bring PaperWM in line with extensions.gnome.org rules - preparing for submission

jtaala:
Combining #557 with this PR.

This PR (finally) brings PaperWM in line with extensions.gnome.org GJS rules & guidelines, which is necessary for submission of PaperWM to extensions.gnome.org.

See issues #205, #435.

Namely, it:

1. Changes PaperWM to use `enable()` instead of `init()`.  `init()` is executed when loaded (e.g. once when logging in).  From [GJS guide/rules](https://gjs.guide/extensions/review-guidelines/review-guidelines.html#rules):
>
> init() is called by GNOME Shell when your extension is loaded, not when it is enabled. Your extension MUST NOT create any objects, connect any signals, add any main loop sources or modify GNOME Shell here.
>
> It is permitted to create static objects and arrays in the global scope or init() (e.g. enumerations or string lists), but constructing class instances is not permitted. Objects intended to be constructed as instances should be created when enable() is called, and destroyed when disable() is called.

2. renames PaperWM UUID to `paperwm@paperwm.github.com`.  This is needed to avoid several issues related to [#489](https://github.com/paperwm/PaperWM/pull/489) (and also since it needs to be somwhere under the control of PaperWM admins).

3. removes the `metadata.json` version parameter.  This is used by extensions.gnome.org to iterate the version number on new releases.

4. simplifies the `install.sh` script!  Auto enabling has not worked for a long (long) time and just confuses users.  The new install script  now checks if  a non-symlinked PaperWM version exists (e.g. one from extensions.gnome.org... when it's available) and if not, applies the symlink with clear instructions to the user on how to enable it.
   - resolves #425. 

5. an installation from extensions.gnome.org means that users won't need to run the `install.sh` script (yay!)... but that means that the `set-recommended-gnome-shell-settings.sh` script wont' be run as well.  Given this, PaperWM now manages these setttings.  It does this by saving/backing-up (to PaperWM's settings/schema) user's settings for `workspaces-only-on-primary`, `edge-tiling`,  and `attach-modal-dialogs`.  
   - we save these settings to PaperWM's settings for safety (in case gnome terminates before PaperWM can restore etc.)
   - PaperWM will save these settings whenever it is enabled, and restore these setting whenever it is disabled;
   - PaperWM clears the backup'd settings after a successful restore, and will only save them on the next enable if they are clear --> this basically ensures that if gnome terminates (crashes etc.), the user's preferred settings won't be lost.
[...]

Issue #435, opened 2022-11-04
[enhancement] Publish PaperWM to extensions.gnome.org


bilboquet:
Hello,

Since I started to use PaperWM, I feel limited when I have to use a computer where it is not installed.
It is probably a lot of work but I think that bringing this wonderful extension to many others would be of great value. 
Hence I suggest it is published to extensions.gnome.org.
It will probably give the project more visibility and hopefully more support.

Many thanks for your work

gombosg:
[Review guidelines](https://gjs.guide/extensions/review-guidelines/review-guidelines.html)
I'll take a look at this. I don't know anything about GJS but good with JS in general so hopefully we could get this through.

bilboquet:
I don't know JS but I can offer some help for tests.

jtaala:
Duplicate of #205.  There may be some issues that go against GJS guidelines (not sure if they would go against https://wiki.gnome.org/Projects/GnomeShell/Extensions/Review).

Mainly things like connecting signals in `init` etc.

Issue #205, opened 2019-11-27
Release to extensions.gnome.org?


danmou:
Is there any reason PaperWM is not on [extensions.gnome.org](extensions.gnome.org) yet? And would it be possible to start using tagged releases, preferably with changelogs etc? I use PaperWM every day but I'm always hesitant to pull the latest changes because I never know what's new and there's always a chance something's gonna be broken.

hedning:
> Is there any reason PaperWM is not on extensions.gnome.org yet?

We'll probably have to change a few things
(https://wiki.gnome.org/Projects/GnomeShell/Extensions/Review), in particular
things are set up for
[gnome-shell-mode](https://github.com/paperwm/gnome-shell-mode/) use which
implies a few oddities. I've kinda just assumed it would be a lot of work, though looking at linked review help, it doesn't look that bad.

> And would it be possible to start using tagged releases, preferably with changelogs etc?

Yeah, we really should start doing that.

olejorgenb:
> And would it be possible to start using tagged releases, preferably with changelogs etc?

https://github.com/paperwm/PaperWM/releases

jtaala:
Closing as done!  https://extensions.gnome.org/extension/6099/paperwm/

PR #557, opened 2023-07-05
Addressing issues from submission 1 to extensions.gnome.org

jtaala:
### I've combined this PR with #556.

See #556, #435, #205.

We got the results back from our first submission to extensions.gnome.org.  Was a very in-depth review from the developer of [JustPerfection](https://extensions.gnome.org/extension/3843/just-perfection/)... (which extension I use and love btw).

Really good feedback (see feedback at end).  I'm going to be working through the issues raised and track my progress below.

- [x] **1 DONE**.   Good pickup and call here - was a bit of a mess.  Including a very interesting hard-coding of the `version` var (was written back before Gnome 40).  In any case, the first commit here was fixing that and removing old/deprecated code that was for gnome 3.xx.
- [x] **2 DONE**.  `sessionModes` removed from metadata.json (not needed in PaperWM).  Also improved the `safeCall` logging to only log if called method exists in module.
- [x] **3 DONE**.  I'm not sure of the original reason for setting `Extension` like it was doing.  Replaced with a much simpler `var Extension = imports.misc.extensionUtils.getCurrentExtension();`.
- [x] **4 DONE**.  Removed `convenience.js` as it duplicates current Gnome ExtensionUtils and refactored modules that used it .
- [x] **5 DONE**.  Removed unneeded check.
- [x] **6 DONE.**.  did full review of all `mainloop.timeout_add` instances - most were  actually (intended) as once off calls (e.g. destroy after first handler execution) - changed these to destroy after handler function execution.
- [x] **7 DONE**.  Went through and cleaned up `imports.xxx` references (and moving them to top where possible).
- [x] **8 DONE**. We were actually delegating the animation through `utils.tweener`... which was actually using `Widget.ease(...)`.  This was basically because it was easier to do this when we switched to `ease`.  However, this is quite confusing for users/reviews.  I've taken the opportunity now to rename `Tweener.addTween(...)` to `Ease.addEase(...)` etc.
- [x] **9 DONE** (finally!).  This was a very good point in the review.  Polyfilling in functions on global gnome objects isn't the best approach here.  I've now replaced the polyfills with backwards-compatible utility functions which is a cleaner/safer/better approach and should make this much easier to maintain going forward.
- [x] **10 DONE**.  Replaced the use of `print(...)` across PaperWM. 
- [x] **11 DONE**.  Moved global scope creation of instances to `enable` in `Kludges.sj`, `liveAltTab.js`, `prefs.js`, `tiling.js`.  Also cleaned up  and removed unneeded imports dependencies and improved formatting.
- [x] **12 DONE**. Hmm - we don't see an issue with importing Gtk and using accelerator_parse etc.  But we have cleaned up and removed references to Gtk where not needed.
- [x] **13 DONE**.  We've removed the hot corner overrding/disabling from PaperWM.  This was largely superfluous since hot corner works fine in PaperWM, and if users want to disable it they can do so via the usual gnome setting.  Having two functions that disable hot corner (one in gnome and one in PaperWM) seems superfluous.
  - so, on PaperWM `disable`, calling `layout._updateHotCorners()` resets hot-corners, effectively restoring the hot-corner user setting. 
- [x] **14 DONE**.  Removed unneeded import in `minimap.js`. 
- [x] **15 DONE**.  `destroy` should not call `dismissDispatcher` since `dismissDispatcher` calls `destroy`... leading to an infinite loop (recursion).
- [x] **16 DONE**.  All `log(...)` lines reviewed and remove superfluous logging; 
- [x] **17 DONE**. Lowercased PaperWM schema id values and recompiled schema.
- [x] **18 DONE**.  Refactored to  NOT call `setSchemas()` on Settings `init` (is not called on `enable`).
- [x] **19 DONE**.  nulled out vars (that we missed).  Good pickup! 
- [x] **20 DONE**. 
- [x]  **21 DONE**.  Refactored StackOverlay class - added proper destroy method and now removePreview on destroy.
- [x] **22 DONE**. Timeouts are now oneshot (remove/destroy timeout after function callback).  This should address this issue.
- [x] **23 DONE**.  Added missing semicolons and other cleanup in `virtTiling.js`.
- [x] **24 DONE**. Now using `add_style_class_name` and `remove_style_class_name` to not run over other extension styles.
- [x] **25 DONE**. Removed Gdk import from Tiling.js.
- [x] **26 DONE**.  Removed module self reference from `keybindings.js` and `tiling.js`. 
- [x] **27 DONE**.  Replaced `Gdk.Keymap` (which isn't available in Gnome 40+) with `Clutter.get_default_backend().get_default_seat().get_keymap()` and hence removed `Gdk` import from `keybindings.js`.  Haven't been able to remove `Gtk` since we need `accelerator_parse` from `Gtk`.
```
1. You mean this instead (line 437 kludges.js)?

    ```js
    if (version[0] >= 40) {
    //          ^
    ```
    
    Same issue for line 443 (kludges.js).

2. Remove `session-modes` or mention the reason in the disable comment section:
    https://gjs.guide/extensions/review-guidelines/review-guidelines.html#session-modes

3. What's the reason for not just using `ExtensionUtils.getCurrentExtension()` instead of lines 6-10 (app.js)?
    I see that in other files too.

4. Use initTranslations() and getSettings() from ExtensionUtils instead of creating your own custom functions (remove convenience.js after that):
    https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/js/misc/extensionUtils.js

5. Please remove line 3-6 (extension.js).
    That's unnecessary since you are using 42 and higher.

6. Remove the timeout in disable/destroy:

    - line 409 gestures.js
    - line 220, 289 navigator.js
    - line 54 settings.js
    - line 96, 233 stackoverlay.js
    - line 1686, 1752, 2787, 2798 tiling.js

    https://gjs.guide/extensions/review-guidelines/review-guidelines.html#remove-main-loop-sources

7. Import at top of the code instead of importing inside a function (for example, line 409 gestures.js).
    In GNOME Shell 45, you will have a hard time to maintain a code like that.

8. `Tweener` is a deprecated module. You can use `ease()` instead.
    For example:

    ```js
    widget.ease({    
        x: newX,
        y: 10,
        opacity: 100,
        duration: 2000,
        mode: Clutter.AnimationMode.EASE_OUT_BOUNCE,
        onComplete: () => {
            log('Animation is finished');
        }
    });
    ```

9. Your extension is 42 and higher, you don't need to use/check/modify:

    - line 263 keybindings.js.
    - line 47-222, 428, 434, 399, 607 kludges.js
    
    btw, the way you are modifying the globals is simply bad for extensions.
    because you cannot revert back or remove them on disable (for example line 219 kludges.js).
    
    Instead of doing that you should check for GNOME Shell version and run the proper code accordingly.

10. Don't use `print()` in your extension (for example, line 439 keybindings.js).
    If you want to log errors, use `console.error()` instead.

11. You cannot create instances in global scope which is the same as init:

    - line 29, 67, 800 kludges.js
    - line 23 liveAltTab.js
    - line 17 prefs.js
    - line 49-54 tiling.js

    https://gjs.guide/extensions/review-guidelines/review-guidelines.html#only-use-init-for-initialization

12. You shouldn't import `Gtk` in GNOME Shell:

    - line 65 kludges.js. (We have `St.Settings` though. so, the code 64-79 is unnecessary.)
    
    Indirectly imports `Gtk` inside extension.js:
    
    - line 21 (liveAltTab.js)
    - line 19 (minimap.js)
    - line 28-29 (topbar.js)
    - line 44 (tiling.js)

13. How do you revert back to original after using `disableHotcorners()` (line 285 kludges.js)?

14. Please remove `Lang` import (line 11 minimap.js).

15. `destroy` should also call `dismissDispatcher()` (line 238 navigator.js).

16. Remove unnecessary logs (for example, line 842 pref.js):
    https://gjs.guide/extensions/review-guidelines/review-guidelines.html#no-excessive-logging

17. Bad caps on schema id (line 4, 263, 270, 406, 407 schemas/org.gnome.shell.extensions.paperwm.gschema.xml).

18. Too much for global scope (line 89 settings.js):
    https://gjs.guide/extensions/review-guidelines/review-guidelines.html#only-use-init-for-initialization

19. You forgot to null out `schemaSource`, `workspaceList`, `conflictSettings` in disable (line 114 settings.js).

20. You should untrack on destroy (line 82 stackoverlay.js):

    ```js
    Main.layoutManager.untrackChrome(enterMonitor);
    ```

21. Cannot say if you call `removePreview()` on destroy (line 275 stackoverlay.js).
    
22. I understand that you call `removeBarrier()` in the destroy of another class (line 198 stackoverlay.js),
    but please remove the timeout in the same class,
    so we can track the timeout removal easier while reviewing (line 382 stackoverlay.js).

23. Please use semicolon since the other lines using it (line 8-105 virtTiling.js):
    https://gjs.guide/extensions/review-guidelines/review-guidelines.html#general-advice

24. Don't add or remove `style_class` like that to avoid removing class names that other extensions adding (line 736, 743).

    Use `add_style_class_name` and `remove_style_class_name` instead:
    
    https://gjs-docs.gnome.org/st12~12/st.widget#method-add_style_class_name
    https://gjs-docs.gnome.org/st12~12/st.widget#method-remove_style_class_name

25. Please remove `Gdk` import (line 29 tiling.js).

26. Remove current module import (line 8 keybindings.js).

27. It would be much better if you can find a way to drop `Gtk` and `Gdk` from lines 404, 413 and 447 (keybindings.js),
    so you can remove `Gtk` and `Gdk` import in that file.

If you need any help with your extension you can ask us on:
- [GNOME Matrix Channel](https://matrix.to/#/#extensions:gnome.org)
- IRC Bridge: irc://irc.gimpnet.org/shell-extensions

btw, for `Gdk.Keymap` replacement read this:
https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/4706#note_1292344
```


jtaala:
Closing since this has been completed.  Have combined this PR with #556.

PR #560, opened 2023-07-20
Address issues from submission 3 to extensions.gnome.org

jtaala:
Evening all,

We got our 3rd submission to extensions.gnome.org reviewed - still need some changes, but we're getting closer to compliance here.  This time there were 9 things that needs addressing (hey, it's better than 28 from our first review!).

See submission review at the bottom or here: https://extensions.gnome.org/review/43111

This PR addresses these issues and also:
- refactors module dependencies to simplify and remove module issues.  Refactored `utils.js` to extract out basic functions into `lib.js` that any other module can use (`lib.js` is has no dependencies).  This was needed since `settings.js` is shared between paperwm and `prefs.js` (prefs.js is run in separated gtk thread and doesn't have access to some gnome shell libraries)  - and it was dependent on `utils.js`.  Refactored other module function into more appropriate places.

- [x] **1 DONE**. Removed the previous enable/disable guard.  This isn't really needed now since we've moved to a model where PaperWM can cleanly disable/enable.
- [X] **2 DONE**.  Refactored imports.
- [X] **3 DONE**.  This warning method has been removed entirely as it's no longer needed.
- [X] **4 DONE**.  Init methods removed.  This means that user.js `init` (while still having legacy support) is no longer in the default `user.js` template.
- [X] **5 DONE**.  Explicit removal of main loop sources on disable have been added. 
- [X] **6 DONE**.  PaperWM gracefully can restore window and multimonitor positions when gnome locks and unlocks (which causes PaperWM to be disabled/enabled.  Have now removed `session-mode` `unlock-dialog`.
- [X] **7 DONE**.  Removed offending static settings create.
- [X] **8 DONE**.  Untracking on disable.
- [X] **9 DONE**.  Moved all `log` instance to `console.log`, `console.debug`, and `console.error`.  Note, to view debug messages now, you need to enable debug mode in GLib (e.g. from looking glass you can enable it by running `GLib.log_set_debug_enabled(true)` and then viewing `journalctl` as per normal. 

```
Good but theses issues still need to get fixed:

1. Selective disable isn't allowed (line 122 and 126 extension.js):
    https://gjs.guide/extensions/review-guidelines/review-guidelines.html#session-modes

4. Move import lines to top of the file:
    - line 140-143 extension.js
    - line 219-221 scratch.js

5. Does `warnAboutGnomeShellVersionCompatibility()` function exist because of legacy code (line 147 extension.js)?

6. Too much for init (line 90-91 extension.js):
    https://gjs.guide/extensions/review-guidelines/review-guidelines.html#only-use-init-for-initialization

7. You should remove the timeout in disable:

    - line 405 gestures.js.
    - line 208 navigator.js
    - line 44 settings.js
    - line 89, 235 stackoverlay.js
    - line 1729, 2742, 2756 tiling.js

    https://gjs.guide/extensions/review-guidelines/review-guidelines.html#remove-main-loop-sources

8. For `session-modes` you should add the reason to the disable function but
    I don't think moving windows to another place is a good reason for that.
    because in lock, your extension will be disabled too,
    despite the fact you have `unlock-dialog` in session mode.
    after disable, it will be re-enable because you have `unlock-dialog` to `session-modes`.
    
    IMO you shouldn't rely on that to avoid windows getting out order.
    Maybe saving the window positions and re-arranging them on enable is the better solution.
    
    Read about session modes here:
    https://discourse.gnome.org/t/extension-disable-called-on-session-mode-change-but-only-for-the-first-time/15832

9. You cannot create instances in global scope which is the same as init:

    - line 12 extension.js

    https://gjs.guide/extensions/review-guidelines/review-guidelines.html#only-use-init-for-initialization

10. You forgot to untrack (line 253 stackoverlay.js).

11. You are still using many `log()`s. Use `console.log()` instead:
    https://gjs.guide/extensions/review-guidelines/review-guidelines.html#no-excessive-logging

```


Lythenas:
>  6 WIP. This one might need some work here. Basically, we can't have unlock-dialog session mode (which keeps PaperWM "alive" during lockscreen. There's a couple of issues here, for example, I've noticed that in multimonitor setups, that on unlock the workspaces can move to a different monitor... Looking at a few options here, but basically we need to save workspaces to monitor and maybe selected window positions so we can restore them on unlock. I'm not planning on removing the unlock-dialog session mode just yet, but we will need to do this before PaperWM can be accepted on extensions.gnome.org.

I can also try to fiddle with this. But manually disabling and enabling the extension should produce the same behavior, right? And I haven't noticed any problems there. I.e. we already do save the window locations in our `Space` objects.

I will try to test this tomorrow and on the weekend.

module.js
```diff
@@ -0,0 +1,66 @@
+var ExtensionUtils = imports.misc.extensionUtils;
+var Extension = ExtensionUtils.getCurrentExtension();
+
+function ExtensionModule() {
+    return Extension.imports.extension;
+}
```
Lythenas: I understand the reason for doing it this way. But it feels very hacky.

Ideally I would like to reorganize the code to get rid of the dependency cycles, rather than doing it this way. But that can also be done later.

Also I have never run into the dependency cycle problem in a JS project before, although maybe I just got lucky^^

jtaala: Hey @Lythenas, yeah, you run into it when trying to refactor (especially to remove `init` methods) - e.g. previously we were statically defining some vars - which other modules dependended on being defined before they ran `enable`.  Problem was, when moving away fron `init` completely broke then... hence, I needed to play around with alot with ordering of `modules` const in `extension.js`.  For examples, `Tiling` had to be before `Settings.js`, but other modules needed `Settings.js` before their enable could be called...

In any case, it was pretty much `dependency hell` until I got it just right... only to be broken again when soon after by removing statically defined vars (object creations - which is the same as calling `init`).

The current approach solves these and means we no longer depend on the module `enable` method being called in specific order.

jtaala: It's pretty much the internal module pattern - modelled it after some ideas in here: https://medium.com/visual-development/how-to-fix-nasty-circular-dependency-issues-once-and-for-all-in-javascript-typescript-a04c987cf0de.

But, I'd happily accept a PR looking at other/cleaner approaches!

jtaala:
Sounds good!  Yeah, I just noticed it on a few occasions with multi-monitor.  Weirdly, when unlocking workspace 1 (which was on one monitor before locking) jumped to the other monitor... not sure what's going on there.

Yep, window order isn't too bad - I just notice that we `ensureViewport` on the selected one so the viewport position can move - but window position order is correct) - so that's not too much of an issue (although might be nice to save/restore viewport position - but that's just icing).

module.js
Lythenas: I would leave it like this for now. If we get time, we can refactor it in the future. But it might require a lot of refactoring^^

jtaala:
> so that's not too much of an issue (although might be nice to save/restore viewport position - but that's just icing).

Hey @Lythenas, I've just pushed some changes that makes disabling/enabling much more seemless in terms of window positions (basically restores viewport position on enable).  Also improved the moving to selected window from overview approach.

jtaala:
> > 6 WIP. This one might need some work here. Basically, we can't have unlock-dialog session mode (which keeps PaperWM "alive" during lockscreen. There's a couple of issues here, for example, I've noticed that in multimonitor setups, that on unlock the workspaces can move to a different monitor... Looking at a few options here, but basically we need to save workspaces to monitor and maybe selected window positions so we can restore them on unlock. I'm not planning on removing the unlock-dialog session mode just yet, but we will need to do this before PaperWM can be accepted on extensions.gnome.org.
> 
> I can also try to fiddle with this. But manually disabling and enabling the extension should produce the same behavior, right? And I haven't noticed any problems there. I.e. we already do save the window locations in our `Space` objects.
> 
> I will try to test this tomorrow and on the weekend.

Cool - to see thee behaviour, make sure to remove the `sessions-modes` json array from `metadata.json`.  Test with multiple monitors and see what happens when locking/unlocking (or enabling/disabling) PaperWM.  Strangely, this behaviour is quite different when `session-modes` is there (even if the logs show PaperWM disabling/enabling on lock/unlock).  It's nice an seamless without multi-monitors (without session-modes)... but multii-monitor setups seems to have issues.

jtaala:
> > I can also try to fiddle with this. But manually disabling and enabling the extension should produce the same behavior, right? And I haven't noticed any problems there. I.e. we already do save the window locations in our `Space` objects.
> > I will try to test this tomorrow and on the weekend.
> 
> Cool - to see thee behaviour, make sure to remove the `sessions-modes` json array from `metadata.json`. Test with multiple monitors and see what happens when locking/unlocking (or enabling/disabling) PaperWM. Strangely, this behaviour is quite different when `session-modes` is there (even if the logs show PaperWM disabling/enabling on lock/unlock). It's nice an seamless without multi-monitors (without session-modes)... but multii-monitor setups seems to have issues.

So... I managed to track the issue down on what I was seeing... that weird issue re workspaces.  It was actually caused by the `workspaces-only-on-primary` setting that I had turned on.  PaperWM manages this setting now and turns it off while PaperWM is enabled (if it's originally on) - then restores it on disable.  Problem is, if it gets turned back on on disable - then it messes up the workspaces and window layouts... (it's a gnome thing there) - which is what I was seeing.

jtaala:
Ok @Lythenas  - so I think I finally got [6] sorted.  Basically I differentiate between a user `disable` and a lock-screen `disable`.  In the latter case, I don't restore that user setting that (which messes up layouts etc. in multimonitors).  This setting however is restored is a user manually disables paperwm.

In the latest commit, I've now removed the `unlock-dialog` session mode.

Give this one a try.

gestures.js
```diff
@@ -1,4 +1,5 @@
 const Module = imports.misc.extensionUtils.getCurrentExtension().imports.module;
+const Settings = Module.Settings();
```
Lythenas: Does this then introduce potential dependency cycles again?

Should we maybe add a comment to `settings.js` that we should not import other modules there?

Lythenas: Also why was it needed to put this on the top level again?

Lythenas:
This seems to work fine now when locking/unlocking on a multi monitor setup. I also haven't encountered the situation where a non-focused window was centered in the workspace (until you focus that monitor) after unlocking.

gestures.js
jtaala: ah, settings is a bit different - we need more than the exported functions - we need the exported prefs var and the gsettings connect functionality (which then basically allow those modules to also become "listeners" to gsetting pref change signals).

Remember imports in gnome extensions basically just copies the exported functions & vars at the top level (in gnome extensions, top level functions and vars layer are exported).  You'll notice the modules that need this are the same ones that need to use `gsettings.connect(..., ....);` functions (e.g. they need to be notified when user prefs are changed).

If you don't do this, then when a user changes a setting (like say `show-window-position-bar`) `topbar.js` won't be informed of this change and do nothing (until PaperWM is restarted).  Same with the other modules.

All other `Module` usages basically just need access to  (independent) functions - hence we don't need to do a full import of other modules at the top (which can produce some side-effects, especially if that module imports other modules).

Anyways, this one tripped me up for a bit since I went to change a setting in prefs and PaperWM did nothing (didn't pick it up until it was restarted).

jtaala: Hope that makes sense.  So, basically, if you just want to use an (independent) function, then `Modules.BlahBlah().function(...)` if sufficient and clean (it's basically just like copy/pasting just that function in the module).  However, if you need more than that - e.g. if an implementation depends on creation of objects, gsetting connections, etc. (basicaly more than independent functions) - then it may need to be fully imported at the top.

The former use case (only just need an independent function in another module) then `Module` referencing is fine... which is like 95% of the use-case here.

jtaala: > Does this then introduce potential dependency cycles again?
> 
> Should we maybe add a comment to `settings.js` that we should not import other modules there?

No, not here, but just because `settings.js` (and also `utils.js`) are a little special and are `pure` modules in the sense that they don't import other modules (or only rely on Module referencing) - see comments above.

jtaala:
> This seems to work fine now when locking/unlocking on a multi monitor setup. I also haven't encountered the situation where a non-focused window was centered in the workspace (until you focus that monitor) after unlocking.

yep, looking good here too.  I made some changes to avoid `ensureViewport` (unless needed) on (re)enable and just use the last `cloneContainer` position.

gestures.js
Lythenas: Ok. If it works, it's fine.

ghost: If you do a function call like this in global scope it will be rejected on ego.

btw, importing modules inside a function is not a good practice and if you do that, you will have a hard time porting it to GNOME Shell 45.

jtaala: > If you do a function call like this in global scope it will be rejected on ego.
> 
> btw, importing modules inside a function is not a good practice and if you do that, you will have a hard time porting it to GNOME Shell 45.

@jrahmatzadeh!  fair enough - will update that.

 Given we've done a boatload of refactoring and changes, and better separating functions into more appropriate modules, we might be able to use imports for those modules (and clean up what gets exported).

Is there something in gnome 45 that impacts doing `imports` within functions?

ghost: Yes, 45 will move to ESM, so import should be at the top of the file. You should wait for the [port guide](https://gjs.guide/extensions/#upgrading) though. We're going to discuss about the imports on GUADEC. _Maybe_ we can come up with a solution for backward compatibility.

jtaala: Thanks so much for the heads-up @jrahmatzadeh!  Yep, good call then.  

Although I did think the `Module.js` approach was cool, at least it made it easier to figure out all the deps issues that were going on...

I've now removed it and put all imports at the top (and took the opportunity to clean up a bunch of stuff).

Hopefully things are looking much better for a new submission to ego.

jtaala: @Lythenas, check out the latest here - I removed `module.js` (I thought it was cool while it lasted...).  Hoping to merge this one info `develop`, and `release` soon!

jtaala: Refactored modules to be more appropriate for whether they're needed in gnome-shell or prefs.js.  Main issues was an inner dependency in settings.js on utils.js (and utils.js imported a bunch of stuff not available to `prefs.js`.  To work around that we were not importing utils.js at the top... but doing it in enable.  Anyways, this is now solved and all modules are imported at the top now.  It's also much cleaner.

Lythenas:
Works fine as far as I can tell.

I can't attest to if we fulfill the gnome guidelines. To be honest I haven't really looked at them yet. Thank you for all the effort @jtaala and also thank you for the feedback @jrahmatzadeh 

Merging this PR seems fine from my side. Unless there is something else you want to address first.

Lythenas:
I also found out that the official gnome extensions have eslint configured: https://gitlab.gnome.org/GNOME/gnome-shell-extensions That might also help to keep more in line with the gnome guidelines. If I have time I will take a look at configuring eslint for this repo. At the very least it should help with code quality/formatting.

jtaala:
> I also found out that the official gnome extensions have eslint configured: https://gitlab.gnome.org/GNOME/gnome-shell-extensions That might also help to keep more in line with the gnome guidelines. If I have time I will take a look at configuring eslint for this repo. At the very least it should help with code quality/formatting.

Thanks @Lythenas, yep, I was already using that .eslint (which was modified to better suit paperwm a little).  I've now pushed this.  It's been very useful in identifying non-used dependencies and other issues (including formatting).  I've been slowly going through and fixing up formatting issues etc.

PR #566, opened 2023-07-28
Full module dependency review and clean-up of exports

jtaala:
This PR is the result of a full module dependency review of PaperWM.

It adds explicit comments on which modules have an `enable` dependency which dictates the execution order of `enable`/`disable` in `extensions.js`.

It also clean up which properties are exported to other modules (in gnome extensions `var` properties are exported, while `let`, `const, etc. are not).  We now explicitly state which properties are exported, which makes paperwm modules much easier to maintain (i.e. we see what's actually needed by other modules).

This is part of an effort to clean-up & productionise paperwm.

### DIFF
--- a/app.js
+++ b/app.js
@@ -10,21 +10,21 @@ const GLib = imports.gi.GLib;
 const Gio = imports.gi.Gio;
 const Shell = imports.gi.Shell;
 
-var Tracker = Shell.WindowTracker.get_default();
-var CouldNotLaunch = Symbol();
+let Tracker = Shell.WindowTracker.get_default();
+let CouldNotLaunch = Symbol();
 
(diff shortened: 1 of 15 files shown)


Lythenas:
This seems fine to me (although I haven't actually run this).

PS: I still can't actually approve PRs...

jtaala:
> This seems fine to me (although I haven't actually run this).
> 
> PS: I still can't actually approve PRs...

Thanks @Lythenas  - okay, I've made some changes to the develop branch permissions that (hopefully) fix that.  Can you approve (e.g. do a review and approve) and see if you can then merge this PR? (or was the issue that you couldn't review and approve?).

PR #573, opened 2023-08-01
Fix "taken" (e.g. `Super+T`) windows should be "dropped" in correct order (order they were taken in)

jtaala:
This PR fixes (corrects) the "take" windows (e.g. `Super+T`) ordering on drop.

Currently, the ordering when dropping multiple windows is quasi-working, but not stable.  E.g., it can sometimes be correct, or part correct, but certain windows may be at the end (when they weren't originally "taken" last).

By "correct", I mean the ordering of dropped windows should be precisely the order that they were taken.  E.g. you if start at the first window and take them all, then they should be dropped exactly in that order (order they were "taken" in).

### DIFF
--- a/tiling.js
+++ b/tiling.js
@@ -114,17 +114,17 @@ var Space = class Space extends Array {
         this.showFocusModeIcon();
         this.unfocusXPosition = null; // init
 
-        let clip = new Clutter.Actor({name: "clip"});
+        let clip = new Clutter.Actor({ name: "clip" });
         this.clip = clip;
-        let actor = new Clutter.Actor({name: "space-actor"});
+        let actor = new Clutter.Actor({ name: "space-actor" });
 
         this._visible = true;
         this.hide(); // We keep the space actor hidden when inactive due to performance
 
         this.actor = actor;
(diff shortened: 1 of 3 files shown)


tiling.js
```diff
@@ -3953,18 +3953,27 @@ function takeWindow(metaWindow, space, {navigator}) {
...
-            });
+            let firstWindow = navigator._moving.find(v => v !== undefined);
+            if (firstWindow) {
+                Utils.later_add(Meta.LaterType.IDLE, () => {
+                    Main.activateWindow(firstWindow);
+                });
+            }
+
+            // clean up after move
+            navigator._moving = [];
```
jtaala, line 3976: Hey @Lythenas, this is the function that was fixed in this PR.

The rest of the changes are simply linter formatting changes (clean up formatting!).

jtaala: We'll apply linting across modules slowly - mainly to help/ease the burdern on @jrahmatzadeh  for his reviews.  ie. just linting `liveAltTab`, `navigator`, `tiling` for this PR, will lint other modules as we work on them.

Lythenas: Yes that seems fine to me.

PR #577, opened 2023-08-11
Fix multi-monitor persistence and improve PaperWM dynamic workspaces support

### LINKED ISSUE #571: Workspaces move between monitors upon replug
**Describe the bug**
I have a laptop with a built-in display and an external monitor plugged in. Upon replugging the external monitor (which happens often because it's a TV and it "replugs" itself every time I turn it on), sometimes, some workspaces switch monitors, which causes windows to end up on the wrong monitor. If unlucky enough, it can be quite annoying to fix (i.e. if the workspace that decided to move to the other monitor was in the middle of other workspaces with windows on them -- which requires a lot of rearranging to fix).

**To Reproduce**
I don't have a 100% reproducer, but more often than not, upon replugging an external monitor, two workspaces will switch monitors (so some workspace on primary will end up on secondary and vice versa). Which workspace switches monitors seems to be loosely based on which workspace is focused at the time of replug, but it's not 100% consistent.

**Expected behavior**
No workspaces switch monitors.

**System information:**

```
Example:
Distribution: Fedora Linux
GNOME Shell 44.3
Display server: Wayland
PaperWM branch/tag: fix-workspace-moving-breaks-workspace-switching
PaperWM commit: 11f9966c45259c8d04374cc6eb1af4ab7c6a296c
Enabled extensions:
- paperwm@paperwm.github.com
```

jtaala:
Fixes #571.

Multi-monitor workspace to monitor persistence was a bit broken (okay, quite broken).

This PR fixesworkspace-to-monitor persistence (e.g. PaperWM's ability to restore monitor-workspace layouts after a unlock or monitor change.

It also improves dynamic workspaces behaviour in PaperWM by:
- aligns some behaviours with gnome's dynamic workspace behaviour and always keep a workspace at the end;
- no longer takes into account the number of `fixed worspaces` when using dynamic workspaces.  Previously, even when using dynamic workspaces the minimum number of workspaces was taken from the set fixed workspace's value.  Now PaperWM dynamic workspace support aligns with gnome's.  Of course, if you want a certain number of workspaces you can use fixed workspaces (see Gnome `multitask` settings).

It now also attempts to remember and restore the previous multimonitor layout across where you had a multimonitor layout, switched to single monitor (during the session) and then reconnect to multimonitors.

The previous approach used a heuristic based on some properties of the monitors - this was unstable though as monitor ordering and (indexes) can change (it's generally stable but sometimes can change).

This PR replace the heuristic (guess) with more stable monitor connector ids (e.g. "eDP-1", "HDMI-1" etc.).  It does this by:

- implementing a new async dbus connector client for `DisplayConfig`.  This gets detailed monitor spec data (like `connector`, e.g. "eDP-1" etc, `vendor`, `serial`, etc.);
- now upgrades `Main.layoutManager.monitors` by adding the `connector` value to each monitor (also detects monitor changes so is up to date);
- we now persist a mapping between the monitor by `connector` (instead of index... which is just not stable) and workspace index (this allows PaperWM to restore a workspace that had been destroyed and then recreated, e.g. if using dynamic workspaces);

If you now check the `INFO_LEVEL` logs on unlock (or disable/enable PaperWM, monitor changes etc.), you'll now see lines like:
```
Aug 13 21:23:21 jay-blade gnome-shell[1599991]: #PaperWM enabled
Aug 13 21:23:21 jay-blade gnome-shell[1599991]: Workspace 6 restored to monitor DP-1
Aug 13 21:23:21 jay-blade gnome-shell[1599991]: Workspace 2 restored to monitor eDP-1
```
which shows a successful restoration of persisted data.

Summary of space-to-monitor restore logic:
- `new login`: primary monitor gets space 1, then monitor 2 gets space 2, etc.
- `lock / unlock`: each monitor's last used space gets restored on unlock;
~~- `multimonitor to single-monitor`: partly a gnome thing, but the space of the monitor where the mouse is will be what shows on the single monitor;~~
- `single-monitor to multi-monitor` : 
  - attempt to restore the last workspace that was on each monitor.  
  - if there's contention, e.g. both monitors last had the same workspace (which can happen, e.g. was on multimonitor and space 1 was on monitor 2, then monitor 2 got turned off, and then we switched space 1 to monitor 1)... then that workspace will be restored to primary monitor (primary monitor takes preference), and the next recently used workspace will go to monitor 2.


app.js
jtaala, line 1: Just formatting and linting changes for this file!

examples/keybindings.js
jtaala, line 1: Just formatting and linting changes for this file!

gestures.js
jtaala, line 1: Just formatting and linting changes for this file!

grab.js
jtaala, line 1: Add some convenience functions (mainly one to get monitor the pointer is currently on).  Other changes are linting and formatting.

utils.js
jtaala, line 1: Implemented dbus DisplayConfig proxy client (gets monitor information via dbus).

Also has some linting/formatting and making top-level imports consistent.

YaLTeR:
Okay, so I:
1. Relogged with second monitor disabled
2. Made some windows, so I had workspaces 1, 2, 3 and (empty) 4
3. Went to workspace 2
4. Enabled the second monitor in Settings

That resulted in Workspace 3 with its windows moving over to the second monitor, which is unwanted. So the main monitor ended up with workspaces 1, 2 and (empty) 4.

YaLTeR:
When that happened it's also a bit peculiar that the second monitor had no other workspaces, so I couldn't switch away and then do the Super+` trick from the main monitor.

YaLTeR:
Right now after some messing around I still have only 4 workspaces in total for some reason with no extra workspaces on either monitor to move windows to.

YaLTeR:
After putting a window on Workspace 4, I now have an empty Workspace 11 (???) which I can switch to from either the main screen or the second screen.

jtaala:
Thanks @YaLTeR, just so I can see what you're on, can you please confirm commit id?

jtaala:
> Okay, so I:
> 
> 1. Relogged with second monitor disabled
> 2. Made some windows, so I had workspaces 1, 2, 3 and (empty) 4
> 3. Went to workspace 2
> 4. Enabled the second monitor in Settings
> 
> That resulted in Workspace 3 with its windows moving over to the second monitor, which is unwanted. So the main monitor ended up with workspaces 1, 2 and (empty) 4.

Okay, let me try to reproduce this exactly and see what I get.

jtaala:
Let's sync up here.  Need precision here so I can reproduce:

- how many windows did you create on each workspace (I'll just use terminals);
- when you said relogged - you mean logout?  or just screen lock?

Also, for testing , please try disableing dynamic workspaces and see if behaviour is any different (that will help see if these issues only occur on dynamic workspaces).

jtaala:
Also, need to make sure we're on exactly the same starting point (e.g. commit id).

Lastly, to be sure can you please check that there's only one paperwm folder in:

`~/.local/share/gnome-shell/extensions`

Thanks!

Jay.

YaLTeR:
> can you please confirm commit id?

830318fafff2ed0028deabd7bbb7d40296201347

> how many windows did you create on each workspace (I'll just use terminals);

3 on first, 2 on second (one of which is the Settings window where I enabled the second monitor) and uhh, 3 on the third I think.

> when you said relogged - you mean logout?

Yep.

> please try disableing dynamic workspaces and see if behaviour is any different (that will help see if these issues only occur on dynamic workspaces).

I'll try a bit later.

> can you please check that there's only one paperwm folder in:

That is correct.

jtaala:
Awesome.  Okay, let me take a video of what I'm seeing - and let me know what I might be doing wrong here (to reproduce).

jtaala:
Hey @YaLTeR, here's what I'm seeing (short video) of trying to replicate what you were doing:

[video: multimonitor persistence attempt 0](https://drive.google.com/file/d/1MDRW5XGeOLBxQSnfHOcnzMMH_9w3mMST/view?usp=drive_link)

~~Ah, dang, stupid good drive, will try upload elsewhere.~~ now link is working apparently.

YaLTeR:
Thanks! Yeah, so this seems to match what happened for me. Before turning on the second monitor, I switched from Workspace 3 to 2, so 3, being the most recently used, showed up on the second monitor. However, I guess I just expect slightly different behavior here: since the second monitor wasn't plugged in at all before during this session, I do not expect any workspaces to move there, instead I expect it to start with a blank workspace. So it might be that Paper works as intended for this case?

jtaala:
> Thanks! Yeah, so this seems to match what happened for me. Before turning on the second monitor, I switched from Workspace 3 to 2, so 3, being the most recently used, showed up on the second monitor. However, I guess I just expect slightly different behavior here: since the second monitor wasn't plugged in at all before during this session, I do not expect any workspaces to move there, instead I expect it to start with a blank workspace. So it might be that Paper works as intended for this case?

Ah, okay - well, it's working as I coded it to.... but I'm not saying it's the best approach.  Was a mix of what I thought would be good/cool and some code constraints.

However, I'm open to suggestions on how it "should" work - given the scenario in the video linked - describe how you think it should work.  When you say a blank space, like one that hasn't been used previously? - or sequential from the current workspace?  Creating a new workspace wouldn't work with fixed workspaces.

Not sure we'd be able to implement it, but open to suggestions.  Also, I note that workspaces are a little different in PaperWM - pretty much all workspaces (except the ones that are currently shown on monitors) are alwasys available to all monitors (i.e. with workspace stacking switching etc.).  So, there's really _no_ hard concept of "workspace a, b belong to monitor 1, and workspaces c, d belong to monitor 2".

jtaala:
The "most recent workspace" tracking approach made sense to me as it seemed natural when switching between single and multi-monitor (e.g. if I'm working in multi-monitor with 2 workspaces => then if I switch to single monitors I have those same two as most recent and hence easiest to switch between).  That seemed more seemless to me (although might not be what other systems do).

The other upside is that it overcomes any issues re delay of monitors waking up after sleep etc.

YaLTeR:
> Creating a new workspace wouldn't work with fixed workspaces.

Right, for fixed workspaces you'll be the better judge; my brain very much works in dynamic workspaces mode. :)

> I'm open to suggestions on how it "should" work

I'm actually currently in the middle of thinking of a design on how dynamic workspaces should work regarding connecting and disconnecting monitors, but it'll be some time until I've had the chance to properly test it and see what actually works and what doesn't. However, some very rough notes are:

1. There's always a single empty workspace available at the end on each monitor (same idea as in vanilla Shell).
2. Connecting an entirely new monitor shouldn't move any workspaces around, so it should only have the one new empty workspace.
3. Disconnecting a monitor moves all its non-empty workspaces to the primary monitor (they have to go somewhere). However, reconnecting the same monitor should be able to restore the workspaces back onto it. So:
    1. All workspaces remember their "original monitor".
    2. When a monitor connects again, any workspaces for which it is the "original monitor" get moved back onto it.
    3. If a new window is created or moved onto a workspace, its "original monitor" is updated to its current monitor. This is to avoid the situation where the user works with two monitors, then unplugs one, keeps working with one monitor for a while (using workspaces without regard to their "original monitor"), then plugs it back, and suddenly seemingly random workspaces move.

This even works with Paper's ability to open an existing workspace on a different monitor: just update the workspace's "original monitor" to the new monitor in that case.

> The "most recent workspace" tracking approach made sense to me as it seemed natural when switching between single and multi-monitor (e.g. if I'm working in multi-monitor with 2 workspaces => then if I switch to single monitors I have those same two as most recent and hence easiest to switch between). That seemed more seemless to me (although might not be what other systems do).

I personally do most of my work on the primary monitor, while the second monitor has auxiliary windows. This is even more true with my current setup where my second monitor is a TV off to the side, so I only move things there when e.g. I want to watch a video on the TV or when I'm livestreaming or recording something so I put the OBS window there. Therefore it never makes sense to automatically move the MRU workspace there.

jtaala:
Cool - yes, the current implementation will at least solve the issue of restoring workspaces as they were, in the case of an external monitor connected.

The design above seems very interesting and can see some good thoughts there.  At face value, I can see some issues that would require quite a bit of refactoring and design changes, but that's to be expected.  

I guess it's personal use cases and preference as I don't use multimonitors like that (but each to their own!) - I often chop/change through workspaces and often switch them around depending on what I'm  doing.  I also very often disconnect and go sit in the lounge while working and then come back to plug in again, so this PR design makes some sense there.

Would definitely consider any PR when you get your design working in PaperWM for sure though.  I don't use dynamic workspaces but the changes in this PR went a long way in cleaning up and making multimonitor things a bit simpler and should provide a better base for working on.

Let me know how you go with the design on dynamic workspaces.

YaLTeR:
Is it reasonably easy to at least make sure there's always an empty workspace in the end of each monitor with dynamic workspaces with the current implementation?

jtaala:
> Is it reasonably easy to at least make sure there's always an empty workspace in the end of each monitor with dynamic workspaces with the current implementation?

Well, I guess `easy` is a relative term (and relatively easy is even more relaive!).  

However, have a look at [gnome shells' _checkWorkspacess](https://gitlab.gnome.org/GNOME/gnome-shell/-/blame/gnome-44/js/ui/windowManager.js#L216) method and compare to [PaperWM's version](https://github.com/paperwm/PaperWM/blob/2599bb015e701afd23912374390ab496f2db61a3/kludges.js#L478) which overrides that behaviour (and which modifies the behaviour around the extra space at end etc.)

jtaala:
> reasonably easy

It's referenced by [line 176](https://github.com/paperwm/PaperWM/blob/2599bb015e701afd23912374390ab496f2db61a3/kludges.js#L176) which you can remove/comment-out to restore default gnome behaviour with dynamic workspaces.  You'll likely have a bunch of side-effects though, but that should get you started.

YaLTeR:
Unfortunately, after a lock/unlock one of my workspaces swapped monitors again

jtaala:
> Unfortunately, after a lock/unlock one of my workspaces swapped monitors again

Oh poops, I've been hammering it and haven't seen that yet.

Can you reproduce it? (is it like you leave it for a while and come back or it just happens?).  Try to reproduce or get steps to reproduce.  I'll try as well but haven't had that yet today.

jtaala:
> I'll try as well but haven't had that yet today.

Okay, I just decided to hammer it and went through 32 consecutive lock-unlock cycles and workspaces were stable and didn't switch at all.  There must be something else going on in your setup.

Tell me, are you monitors the exact same resolution?

jtaala:
Ha, maybe it's multiple `monitor-changed` signals during unlock?  I've seen that referenced somewhere else.  I try a few guards (or delay time for restore) to see if that helps.

YaLTeR:
> Tell me, are you monitors the exact same resolution?

Nope, laptop is 2560x1600 and secondary is 1920x1080.

jtaala:
> > Tell me, are you monitors the exact same resolution?
> 
> Nope, laptop is 2560x1600 and secondary is 1920x1080.

Okay, do a pull now.  I've implemented a 100ms guard and delay after receiving a `monitor-changed` signal (and we ignore subsequent `monitor-changed` signals during the 100ms guard).  After guard delay we'll kick off an actual monitorChanged() method call to restore.

I've also remove the heuristic check on resolution checks for restore.

May be a long shot but give https://github.com/paperwm/PaperWM/commit/b8867ad3f90755199e803a6da6b732e69eb0108c a good work out.

jtaala:
Whoops, just a sec

jtaala:
> Whoops, just a sec

Okay, good to go.  Should be commit https://github.com/paperwm/PaperWM/commit/b8867ad3f90755199e803a6da6b732e69eb0108c.

jtaala:
A little cleanup just pushed.  See how you go. 

I've now done a total of 70 consecutive lock/unlock cycles and a similar number of "monitor yanks" (or just changing monitor setup between multimonitor and single) - haven't been able to reproduce a single workspace swap.  Not sure what else to try.

It's got to be some weird corner case.  Keep an eye out for any reproducible steps, as unless I can reproduce it, I'm not going to be able to isolate and fix it.

This PR fixes the monitor-swap I could reproduce (which was the monitor disconnect/reconnect one in the video I posted).  Whatever is happening with your setup we haven't been able to catch yet.

Sadly, you might have to get good at workspace stack switching to work around it until we've got a reproducible path.  

If it helps, the default key for `Switch to previously active workspace` in PaperWM makes this much easier ``Super+` `` (`Super + backtick`).  Do that on the workspace you want to move to the other monitor, move to the other monitor (with mouse or `Switch to right monitor` etc. shortcut) and then do `Super + backtick` again and you'll open the workspace on the new monitor).

If you're interested I can look at implementing a nice shortcut to swap monitor around (although I wonder how that will work with 3 monitors etc.).

jtaala:
I just realised we aren't saving the layout (for monitor restore) for dynamic workspaces when a new workspace is created, removed etc).  Now we do that when a workspace moves monitors, but forseeable, one could create a dynamic workspace and not move a workspace - then lock/unlock and restore could get a bit messed?

Does that sound like something that could happen in your workflow @YaLTeR?   Just an thought.  I don't use dynamic workspaces so that may have something to do with it?

There is something else we could try as well - and that's switching from our restore heuristic to dbus display connector ids (it's a bit of work to replace it but eventually we might need to try that).

YaLTeR:
> I just realised we aren't saving the layout (for monitor restore) for dynamic workspaces when a new workspace is created, removed etc). Now we do that when a workspace moves monitors, but forseeable, one could create a dynamic workspace and not move a workspace - then lock/unlock and restore could get a bit messed?

Hmm, well, yeah, I pretty much never move workspaces between monitors currently. Only to move it back when Paper unwantedly relocated it.

jtaala:
> Hmm, well, yeah, I pretty much never move workspaces between monitors currently. Only to move it back when Paper unwantedly relocated it.

well, cross that one out then.  Only things I've got left are something to do with dynamic workspaces (maybe that's why I haven't been able to reproduce on my system?) and monitor indexes being unstable (which is a _known_ thing in gnome).

Let's try implement actual connector id's as monitor identifiers (instead of relying on the monitor index order as reported by gnome `LayoutManger`).

On a sidenote, I've been playing with dynamic workspaces and PaperWM does seem to have some issues there (for example, doing a `<Super+Shift+pg_up>`, which is supposed to move the current window to a new dynamic workspace seems to mess things up in paperwm).  Anyways,  I reckon we give improving that a go and implement something like the design you mentioned above.

Can look at that in a separate PR though after we get some of the fixes in this PR in.

jtaala:
Hey all, converting this to draft as I'm implementing the async dbus proxy connector (which upgrades `Main.layout.monitors` to store the monitors `connector` value, e.g. "eDP-1" etc.).

Looking promising but does need a little bit of time to get right.

YaLTeR:
Uh, so I have no idea what I've done or if it's related to this PR, but after one of the unlocks I got a super glitched session:

Screenshot from 2023-08-13 15-00-21

This somehow cuts in the middle of the screenshot, but since my screens are not equal size (1920 and 2560 in that order), the cut is actually in the left part of my right screen.

I managed to kb-nav to an extensions window and disable Paper, then it showed this and the session didn't unglitch (I had to relogin to fix things):

Screenshot from 2023-08-13 15-01-06

I'm on 6230668498c9c8b8d5d5d1fc68f7802dc8811b3f.

jtaala:
Hey @YaLTeR - yeah, that commit turned out to be not a great approach.

I've just finished up, what I think is a winner!  Def pull the latest on this PR, it should be commit https://github.com/paperwm/PaperWM/commit/a2cdf3deaaa47dbbfaf0a435c6cf4718104e1d4a.

jtaala:
I'm just about ready to bring this one out of draft state.

So, here's what's different (pretty much a redesign of the restore approach):

- implements a new async dbus connector client for `DisplayConfig`.  This gets detailed monitor spec data (like `connector`, e.g. "eDP-1" etc, `vendor`, `serial`, etc.);
- upgrades `Main.layoutManager.monitors` by adding the `connector` value to each monitor (also detects monitor changes so is up to date);
- we now persist a mapping between the monitor by `connector` (instead of index... which is just not stable);
- if you now check the logs on unlock, you'll now see lines like:
```
Aug 13 21:23:21 jay-blade gnome-shell[1599991]: #PaperWM enabled
Aug 13 21:23:21 jay-blade gnome-shell[1599991]: Workspace 6 restored to monitor DP-1
Aug 13 21:23:21 jay-blade gnome-shell[1599991]: Workspace 2 restored to monitor eDP-1
```
which shows a successful restoration of persisted data.

jtaala:
> or if it's related to this PR,

Yeah, that was related to that commit state of this PR... turns out the issue was not what that commit assumed (thought might be multiple monitor changes signals) - adding the guard on there had a side-effect of messing some stuff up).

jtaala:
So, I'm thinking that one improvement would be that, with multimonitors and a fresh login - that workspace 1 should always be on the primary monitor?  Mainly for consistency.

Currently, if no saved previous layouts (e.g. new login) that may not be the case, at least on my setup (which may be a bit confusing).

jtaala:
> Currently, if no saved previous layouts (e.g. new login) that may not be the case

Found the reason, basically because primary monitor isn't guaranteed to be the first montor in `Main.layoutManager.monitors`, and for restore PaperWM iterates through that monitor array (and hence the first monitor in that array might not be primary).

YaLTeR:
Ah, cool, thanks

jtaala:
So, now that I've used this for a few days, and although it does fix the original issue, still not quite happy with some of the behaviour, especially when you switching between single/multiple monitors.

I think the approach where we try to keep/persist the last multi-monitor layout, even amidst switching to single monitor mode, and changing things up leads to some non-intuitive behaviour.

I'm going to change this slightly to just persisting the monitor layouts whenever:
- you lock (so on unlock prev state is restored);
- monitors change; so just save that layout (and don't try keep the last "multi-monitor" layout - just any layout as it was when changed);

This will simplify things and hopefully make the restore behaviour more intuitive.

jtaala:
My only concern with the simplified approach here (which may just be theoretical as I haven't proven it), is that if an external monitor takes a while to "wake up", then it may be treated as a new space.

For example, say we have space 3 on the primary monitor and space 6 on the external:

When we lock, we save that layout.... then external monitor sleeps/turns off etc.  Then we wake our computer and unlock... but the external monitor is off (or takes a while to wake up).  PaperWM on unlock will try to restore space 6 to the external monitor... uh oh, it's not there (since is off or waking up) --> PaperWM says "oh, well, not there, can't restore" and then restore space 3 to primary and will save this new layout (for the next restore).

Then, when the external monitor is woken up (or turned on) - PaperWM restores the "last" saved layout (which now doesn't have the last multimonitor layout saved)... and will assign another space to it.

This was what I was thinking when I chose to persist the last *multimonitor* layout.

At @YaLTeR - does this sound like a feasible use case to consider?  It's worth testing with this new simplified approach in any case to see how it goes.

jtaala:
Some testing - yes, this does happen if the monitor is "off" when we unlock PaperWM.  So it seems you actually might want to save the last space on a monitor and persist that, regardless of whether it's there on restore or not (which is fine btw) - so that when it does turn up again, we attempt to restore the last space that was on it.

YaLTeR:
Not sure off the bat, we'll see. FWIW the way I use the second monitor now, if it always starts with an empty workspace, would be better than if it sometimes gets a workspace from primary.

YaLTeR:
Not sure if it has anything to do with this PR but over the past two days I've noticed something aggressively trying to warp my cursor to the center of the laptop monitor every time I try to move the cursor to the second monitor. Only occasionally it lets me do that.

jtaala:
Cool, pull the latest - we now persist monitor connectors, and the last workspace that was on it (whether the monitor exists or not).  Means that, if primary monitor hasn't used that workspace (after external monitor) then it will be restored whenever that external monitor shows up.

jtaala:
> if it always starts with an empty workspace, would be better than if it sometimes gets a workspace from primary.

Stealing from primary shouldn't happen on latest (https://github.com/paperwm/PaperWM/pull/577/commits/ad5fc1a23f3633277403c478be1d6856b28e6c2a).  

We persist that monitor's last used workspace now.  So, unless the primary has used that workspace (after external) then it will be restore to external.

It now shouldn't steal the workspace from primary monitor.  This implicitly works since in restoration, primary monitor is checked first, then other monitors.

So basically should work like:
- `new login`: primary monitor gets space 1, then monitor 2 gets space 2, etc.
- `lock / unlock`: each monitor's last used space gets restored on unlock;
- `multimonitor to single-monitor`: partly a gnome thing, but the space of the monitor where the mouse is will be what shows on the single monitor;
- `single-monitor to multi-monitor` : 
  - attempt to restore the last workspace that was on each monitor.  
  - if there's contention, e.g. both monitors last had the same workspace (which can happen, e.g. was on multimonitor and space 1 was on monitor 2, then monitor 2 got turned off, and then we switched space 1 to monitor 1)... then that workspace will be restored to primary monitor (primary monitor takes preference), and the next workspace (e.g. workspace 2) will go to monitor 2.

jtaala:
> I've noticed something aggressively trying to warp my cursor to the center of the laptop monitor

Let me know if you get steps to reproduce this.  Only thing I've seen that has ever warped pointer to the center is the switch monitor keybinds for `Switch to the right monitor`, `Switch to the left monitor`.

jtaala:
> would be better than if it sometimes gets a workspace from primary

Just to clarify, in PaperWM, only the spaces currently shown on monitors "belong" to that monitor - all other spaces belong in the void! (for lack of a better word) and can be "opened" on any monitor (through PaperWM workspace switching etc.).

But yes, on the latest, we shouldn't get a space that is currently shown on a primary monitor "yanked" to another monitor (unless there's a bug or some corner case I haven't thought of!).

jtaala:
> unless there's a bug or some corner case I haven't thought of!)

Okay, now it shouldn't ever steal a space from a primary monitor! https://github.com/paperwm/PaperWM/pull/577/commits/935760462c503b33674cc0cda5e4da023b3db465.

jtaala:
See latest - we now persist and restore viewport position (e.g. tiling position) for each space as well - for a seamless restore (i.e. will look just like before disabling).

This PR turned into a much larger PR than first thought, but in general I'm now happy with the behaviour of restores and am planning on merging this in soon (next couple of day or so).

YaLTeR:
> Let me know if you get steps to reproduce this. Only thing I've seen that has ever warped pointer to the center is the switch monitor keybinds for Switch to the right monitor, Switch to the left monitor.

Updated to latest PR commit, still seeing this. Pretty much most of the time when I try to move my mouse to my second monitor, be it by moving the mouse or with the "Switch to the left monitor" bind, a splitsecond after the cursor enters the monitor it gets warped back to the center of my main monitor. I frequently need to spam "Switch to the left monitor" multiple times to finally get it over. Whenever this happens the focused window indicator in the top bar does show the second monitor window for one frame, so it is only after the pointer is on the second monitor that it is warped back.

jtaala:
> Updated to latest PR commit, still seeing this

It sounds ClickOverlay related.  After a new login, how easy is it to reproduce this?

If you can always get into this state, can you try bisecting this?  Especially if you're only seeing this since the last few days.

@Lythenas have you seen this?

I haven't seen this so I wonder if it's something to do with monitors with different resolutions?

jtaala:
Give me a quick rundown of your setup @YaLTeR, so I can try replicate.  Even better, if your able to take a video/photo (on phone or something) showing your setup and that issue.

- two monitors, external monitor to the left?
- external monitor resolution 1080p and laptop monitor 1440p?
- which monitor is the primary monitor?
- dynamic workspaces?
- you don't use any setting like "focus follows mouse" do you?

jtaala:
hmm - found something that might be related.  Scratch windows - in my testing if there's a scratch window on a monitor, the the switching monitor shortcuts seem to fail.

Will do some digging.

jtaala:
Hey @YaLTeR - I think 

> Updated to latest PR commit, still seeing this

I think(?) I found it @YaLTeR.  https://github.com/paperwm/PaperWM/pull/577/commits/1b1606f63b597a6115bc5f417f62f9fd93984e66 should fix this.  Was caused by incorrect reference in `Scratch` to a new get function in `Tiling.spaces`.

Pull the latest and test that one out.

Luckily @YaLTeR uses scratch windows (purportedly) since it threw an error on console - but since I don't often use scratch windows didn't pick it up.  Nice find @YaLTeR!

P.S. I'm pretty sure this was the issue but let us know @YaLTeR if it still happens!

jtaala:
> hmm - found something that might be related. Scratch windows - in my testing if there's a scratch window on a monitor, the the switching monitor shortcuts seem to fail.

Interesting, it seems scratch windows maintain focus even when moving mouse to another workspace.  This differs from normal PaperWM behaviour.  @YaLTeR - can you recall if you had scratch or floating windows on a monitor when the warping happened?

Will have a look at this tonight.

YaLTeR:
> Luckily @YaLTeR uses scratch windows

I do use floats occasionally, but I don't think I had any when the warping occurred.

> Give me a quick rundown of your setup

image

Funnily enough now that I'm taking this screenshot the issue does not seem to reproduce.

> dynamic workspaces?

Yep.

> you don't use any setting like "focus follows mouse" do you?

I don't.

Lythenas:
@jtaala I have occasionally seen the mouse cursor warping (also before this PR) but not recently. And it has never really been an issue. But I will try the latest changes today.

I do use the scratch layer daily. E.g. the password manager and VPN window is there. Also when running the Android Emulator I usually put it on the scratch layer because that seems to cause less trouble.

jtaala:
Thanks @YaLTeR and @Lythenas for all the testing.

Another day another fix(es)!  See latest (https://github.com/paperwm/PaperWM/pull/577/commits/638484142e168045692dd3dab564880a0c70f759) for changes which fix the scratch-windows-across-monitors issues.

I think (hope) we're nearly there with this PR.

jtaala:
> I have occasionally seen the mouse cursor warping (also before this PR) but not recently

So, I finally found that stupid pointer warp issue.  Could reproduce it somewhat (basically noticed that when clicking the overview workspace thumbnail - pointer would warp to other monitor and was like "ha?").

Tracked it down to the warp monitor code being in `switchWorkspace` which is called whenever any workspace is activated.  It was tripping on some weird corner cases.

I've now moved the warp pointer code out of there and only in `switchMonitor` - **which means it is only ever called when using the `switch monitor to left/right/etc.`** (which is how it should be).

_P.S. is scratch / floating window was another issue... which is good because we've also fixed that._

jtaala:
Okay, let's try this again... pretty confident that found and fixed the pointer warp code... planning on merging this into `develop` in a day or so (maybe saturday).  Appreciate any last minute tests or feedback you have on the latest (currently https://github.com/paperwm/PaperWM/pull/577/commits/7330509886e8c415a077ec4a713f0e6289ad5a09).

Thanks all.

jtaala:
@YaLTeR & @Lythenas,

So, there's a reason why myself and @Lythenas haven't been able to reproduce what you're seeing @YaLTeR: *dynamic workspaces*.

Dynamic workspaces is the issue with replug/restores.  Basically, PaperWM overrides gnome's `_checkWorkspaces` and manages dynamic workspaces a little difrerently, and stops things like removing ~~empty workspaces~~ PaperWM's visible spaces.

When PaperWM is disabled we undo the PaperWM's `_checkWorkspaces` method.  Now gnome's `_checkWorkspaces` is in play, and this method removes empty workspaces (since PaperWM's version isn't active).

That means that any change in monitors will trigger gnome's version and remove empty workspaces etc.  Problem is PaperWM treats workspaces differently, and hence on restore, workspaces have been removed, are different then when PaperWM left... and it's now left with a bit of a mess trying to restore spaces (which are intrinsically linked to workspaces) against a changed gnome workspace landscape (without know what happened).

Anyways, not sure what to do here.  I'll look at restoring empty workspaces to what PaperWM saw when unlocking before dynamic workspaces changed things (without PaperWM's knowledge).

So basically, this PR fixes a ton of issues with multimonitor in PaperWM and are good changes... for fixed worksapces... but dynamic workspaces presents a few extra challenges here.

jtaala:
As I feared, dynamic workspaces leads to some very indeterminate behaviour during restore.  Weird stuff that doesn't happen with fixed workspaces.  Easily reproducible bugs when an external monitor is "yanked" out, or turns off / timeouts during a lock state.  Even in lock, gnome code will then remove/modify workspaces and leave PaperWM in a state where it can't restore correctly.  This is essentially why @YaLTeR was seeing workspace switching and other issues, when I wasn't.

Furthermore, I've found some pretty weird bugs around using dynamic workspace functions (like shortcuts to create a dynamic workspace, and dragging a window in overview to create an "in-between" workspace.  It quickly messes up the workspace names/ordering etc. and leads to some other weird behaviour.

I don't see a way around this, other than rewriting dynamic workspace support in PaperWM - which will likely require quite some refactoring and redesign on the spaces, monitor code.  This would provide a good opportunity to do a better approach for dynamic workspaces.

Hence, in order to provide a stable experience (e.g. workspaces/monitors restore properly etc.) we pretty much need to make `dynamic-workspaces` a controlled setting in PaperWM (like other settings that don't work well in PaperWM, e.g. `attach-modal-dialogs`, `edge-tiling`, `workspaces-on-on-primary`).

_This means that PaperWM will now disable `dynamic workspaces` when it is enabled - and restore the original setting when PaperWM is disabled._

Sadly, currently you can't have both consistent and stable restores AND dynamic workspaces.  There's too many corner cases and issues currently with PaperWM and dynamic workspaces.

@YaLTeR, best way forward I can see is doing this, and then working on proper/better dynamic workspace support - as you outlined previously.

YaLTeR:
I see, understandable.

> This means that PaperWM will now disable dynamic workspaces when it is enabled - and restore the original setting when PaperWM is disabled.

Could it let you enable it with a warning or something? Because it works "well enough" at least for what I'm doing; I can deal with the occasional workspace moving around, and know to avoid the overview workspace thumbnails with Paper by now.

jtaala:
> Could it let you enable it with a warning or something?

Yeah, I can add a persistent user config setting (prob will call it something like `allow-dynamic-workspaces` that you can set to true in dconf etc.).

smichel17:
What about, as a workaround, a way to move a workspace to a different monitor? Or swap two workspaces between monitors. So for now we don't worry about whether we sometimes get mixed up and put the wrong workspaces on the wrong monitors, because the user can fairly easily get them back where they belong. Maybe a shortcut, with no default keybinding.

(note: I haven't read the whole thread. I'll test this out next week)

jtaala:
> What about, as a workaround, a way to move a workspace to a different monitor? Or swap two workspaces between monitors. So for now we don't worry about whether we sometimes get mixed up and put the wrong workspaces on the wrong monitors, because the user can fairly easily get them back where they belong. Maybe a shortcut, with no default keybinding.

Well, this would help but not workaround all the issues.  With dynamic workspaces, you not only have workspaces being incorrectly restored - you also have instances where windows of workspaces are incorrectly restored to the wrong workspace (e.g. windows were on workspace 3 but on restore they all get restored to workspace 2).

Plus, it wouldn't help with the workspace names & paperwm space settings getting impacted/messed up for dynamic workspaces.

In other words there's deeper issues, other than just workspaces being restore to an incorrect monitor.

jtaala:
> a way to move a workspace to a different monitor? Or swap two workspaces between monitors.

But I do think this is a good idea and will be super useful (currently to swap spaces on monitors I need to stack switch a space away, then move to the other monitor, and stack switch that space to on the monitor, then do the same on the original monitor).  A quick shortcut to do that would be very useful.

jtaala:
@YaLTeR  Just added a setting to stop PaperWM controlling dynamic-workspaces.  Run this from a terminal window (once you pull latest on this PR and logout/login):

```
dconf write /org/gnome/shell/extensions/paperwm/allow-dynamic-workspaces true
```
then PaperWM won't touch dynamic workspace settings.

Note: once you run that you can disable/enable PaperWM (which will restore dynamic workspace and then when PaperWM is enabled again won't touch it).

jtaala:
Ha... that was unexpected.

I had an idea to use our `space.uuid` (each space has a unique id we use for saving settings etc.) for restore instead of the workspace index... theory being that with dynamic workspaces the workspace index is not stable.

Turns out - I'm now getting a stable restore with dynamic workspaces...

As such, I've now disabled the enforcement of fixed workspaces - as it may not be needed now.

I'll continue to test the latest out - @YaLTeR, if you can also start testing this one out, would be appreciated.

Might still be some corner cases since I've just implemented this but it looks promising.

jtaala:
All, please see latest changes here.  Namely this:

> It also improves dynamic workspaces behaviour in PaperWM by:
> - aligns some behaviours with gnome's dynamic workspace behaviour and always keep a workspace at the end;
> - no longer takes into account the number of `fixed worspaces` when using dynamic workspaces.  Previously, even when using dynamic workspaces the minimum number of workspaces was taken from the set fixed workspace's value.  Now PaperWM dynamic workspace support aligns with gnome's.  Of course, if you want a certain number of workspaces you can use fixed workspaces (see Gnome `multitask` settings).

Dynamic workspaces seems to be performing well with latest changes and restoring stably (that's a word right?).  I'm quite liking it.  It's seems intuitive and allows to create workspaces when needed, and destroys the _end ones_ when empty.

A couple of things to note now:
- in overview, dragging a window to workspace thumbnails to create an "in-between" workspace works fine... but note that the workspace is injected / moved (this is gnome behaviour).  e.g.
  - dragging a window in between workspace 1 and 2, will move workspace 2 (and windows) to workspace 3 and inject the moved window to "workspace 2";
  - this makes sense from a technical pov but might be surprising to users (unless they're used to  this behaviour in dynamic workspaces);
- there is a use-case that can restore unstably: 
  - when you lock and then "yank" out the external monitor cable (which is a destructive monitor change) AND then unlock and eventually reconnect;
  - note that this differs from the external monitor turning off or sleeping - which restores fine;
  - not quite sure what to do here, it's a pretty "destructive" use case but can happen.

YaLTeR:
Thanks! I'm going to give this a try. Sounds great!

> unless they're used to this behaviour in dynamic workspaces

Yeah, this is what I expect (I don't even look at workspace numbers tbh)

YaLTeR:
So after rebooting some of my autostarted windows are glitched out, and there's also this in the journal:

```
авг 21 08:22:03 gnome-shell[1488]: JS ERROR: Exception in callback for signal: showing: TypeError: Tiling.spaces.monitors is undefined
                                      fixTopBar@/var/home/yalter/.local/share/gnome-shell/extensions/paperwm@paperwm.github.com/topbar.js:709:17
                                      _callHandlers@resource:///org/gnome/gjs/modules/core/_signals.js:130:42
                                      _emit@resource:///org/gnome/gjs/modules/core/_signals.js:119:10
                                      _changeShownState@resource:///org/gnome/shell/ui/overview.js:300:14
                                      runStartupAnimation@resource:///org/gnome/shell/ui/overview.js:683:14
                                      _startupAnimationSession@resource:///org/gnome/shell/ui/layout.js:778:27
                                      _startupAnimation@resource:///org/gnome/shell/ui/layout.js:763:18
                                      _prepareStartupAnimation@resource:///org/gnome/shell/ui/layout.js:754:14
                                      async*_loadBackground/signalId</id<@resource:///org/gnome/shell/ui/layout.js:681:26
```

jtaala:
> So after rebooting some of my autostarted windows are glitched out, and there's also this in the journal:
> 
> ```
> авг 21 08:22:03 gnome-shell[1488]: JS ERROR: Exception in callback for signal: showing: TypeError: Tiling.spaces.monitors is undefined
>                                       fixTopBar@/var/home/yalter/.local/share/gnome-shell/extensions/paperwm@paperwm.github.com/topbar.js:709:17
>                                       _callHandlers@resource:///org/gnome/gjs/modules/core/_signals.js:130:42
>                                       _emit@resource:///org/gnome/gjs/modules/core/_signals.js:119:10
>                                       _changeShownState@resource:///org/gnome/shell/ui/overview.js:300:14
>                                       runStartupAnimation@resource:///org/gnome/shell/ui/overview.js:683:14
>                                       _startupAnimationSession@resource:///org/gnome/shell/ui/layout.js:778:27
>                                       _startupAnimation@resource:///org/gnome/shell/ui/layout.js:763:18
>                                       _prepareStartupAnimation@resource:///org/gnome/shell/ui/layout.js:754:14
>                                       async*_loadBackground/signalId</id<@resource:///org/gnome/shell/ui/layout.js:681:26
> ```

Ewww - does that happen every start?

jtaala:
> autostarted windows

Hmmm autostarted windows!  What are they and how are you autostarting them?  Weirdly, they seem to start before PaperWM has finished initing monitors etc.

This the first time you've noticed it?

YaLTeR:
> Hmmm autostarted windows! What are they and how are you autostarting them? Weirdly, they seem to start before PaperWM has finished initing monitors etc.

`~/.config/autostart`, also available from GNOME Tweaks

> This the first time you've noticed it?

Yeah, it's the first time it happened. Restarting the extension fixed it fwiw

jtaala:
> Yeah, it's the first time it happened. Restarting the extension fixed it fwiw

Cool, it's worth putting an optional chain on it anyways.

Apart from that - is dynamic workspaces working somewhat now?

YaLTeR:
> Apart from that - is dynamic workspaces working somewhat now?

Still giving it a go, seems to work so far. Some very minor things:

- on an empty monitor there should be only one workspace I think, right now there are two empty ones
- I seem to have quite a few workspaces named "Workspace 11"

jtaala:
> * I seem to have quite a few workspaces named "Workspace 11"

Yeah, that's prob from your previous use of dynamic workspaces - which wasn't handled as nicely, and had some funny stuff going on (from old restore etc.)

Disable PaperWM, open up `dconf`, find paperwm folder (just search), right-click on `workspaces` folder (under paperwm) and select `reset recursively` and the apply.

Then enable PaperWM again.

There will be a terminal command to reset workspaces that we should prob document.

jtaala:
> * on an empty monitor

Hmm - you're prob on multimonitors?  So you have one extra (always have an extra).

E.g. 1 space for built-in monitor, 1 space for external monitor and once at the end.

Consistent with single monitor setup (e.g. 1 for monitor and one at end).

YaLTeR:
Ah, so like the empty workspace is still "shared" between monitors?

jtaala:
> Ah, so like the empty workspace is still "shared" between monitors?

Yes.  In PaperWM ALL workspaces (not shown on current monitor/s) are shared between any monitor.

E.g. if you're on wayland and touchpad, do a 3-finger swipe down and you should see other spaces.  Can do that on either monitor (and see the same spaces).

YaLTeR:
Right, I keep forgetting that

YaLTeR:
I'm right now on 3903938e041d46bbbfceae33c2ea0fd203198e73 after a resume from sleep with my two monitors, and there's no empty workspace at the bottom of the main monitor. I have 4 workspaces in total with the only empty workspace sitting on the second monitor. What I did to get this was to Super+` the non-empty workspace from the second monitor into the main monitor.

This is the same as PaperWM used to work before, but not sure if intended with the latest changes.

jtaala:
> This is the same as PaperWM used to work before, but not sure if intended with the latest changes.

Hey Ivan, yes this is intended.

You seem to be struggling with workspace concepts in PaperWM - it is different from other window managers but probably one of my favourite things about PaperWM.

So, there's no such thing as a workspaces "belonging" to a monitor - there's simply a single "line" of workspaces available to any monitor, and we simply "show" a workspace on a monitor - but once it's moved away from any monitor - it can be "shown" on another monitor.

Hence

> and there's no empty workspace at the bottom of the main monitor

doesn't make any sense in PaperWM, since no workspace can be "at the bottom" of a monitor.  Once it's not shown on a monitor, it's back in the "line" or available workspaces (which are available to any monitor).

> What I did to get this was to Super+` the non-empty workspace from the second monitor into the main monitor.

This is exactly the function of Super+backtick  - it shows the last used workspace on any monitor - and this allows us to quickly move a workspace to another monitor (which is another one of my favourite features) - e.g. "unshow" (basically switch to another workspace on monitor 1) and then on another monitor hit Super+backtick and you've just "moved" a that workspace to another monitor.

YaLTeR:
> You seem to be struggling with workspace concepts in PaperWM

Nah, I'm just never sure how close "improved dynamic workspaces support" got to how dynamic workspaces work in GNOME :) It's okay though

> it is different from other window managers

Not really, i3 and sway also work this way, seems to be quite a popular implementation actually. Funnily enough when I used sway for a few years, I set up a static set of workspaces for each monitor, because even then I found it more convenient to have per-monitor workspaces.

jtaala:
> Nah, I'm just never sure how close "improved dynamic workspaces support" got to how dynamic workspaces work in GNOME :) It's okay though

Gotcha.  Improved support basically means PaperWM isn't (or shouldn't) be borking workspace names when using standard dynamic workspace features (like dragging windows to overview thumbnails to create a new workspace in between other workspaces) and other issues which caused a bunch of duplicate workspace names etc. and always keeping an empty workspace for use at the end.

> Not really, i3 and sway also work this way.

Yes, I also come from i3, so this seemed very natural to me - I assumed from many of your references about workspaces and monitors (e.g. "Right, I keep forgetting that") that it may have been a foreign concept - but good to hear it's not!

jtaala:
> I'm just never sure how close "improved dynamic workspaces support" got to how dynamic workspaces work in GNOME :)

Fair enough - for what it's worth, I agree that many gnome approaches (for dynamic workspaces or otherwise) isn't what I consider best - but given PaperWM is a gnome extension, "improved dynamic workspaces support" means improved support for (gnome's) dynamic workspaces.

jtaala:
After using this with dynamic workspaces for a few days now, I think it really needs a shortcut to inject/insert a new workspace to the right of the current workspace (basically do what happens when you drag a window in overview between two of the workspace thumbnails in overview).

jtaala:
Merging into `develop` to get this more widely tested.

Issue #580, opened 2023-08-18
resizeWInc() and friends don’t work with GNOME terminal


gustavcedersjo:
The resize functions that increase or decrease the window size does not work with GNOME terminal.

The resize functions gets the current window size, calculates how high/wide the window is, rounded downwards to tenths of the maximum size and adds/subtracts a tenth to the rounded size. The problem is that GNOME terminal also resizes its windows to whole rows and columns. If the user resizes a window to 30%, say, and GNOME terminal decreases that size to 29,9%, the next call to resizeWInc() will increase the size to 30%, and GNOME terminal will decrease it back to 29,9%.

Solution: replace the appropriate Math.floor() with a Math.round() in the resizeWInc() functions and friends.

PR #581, opened 2023-08-18
Fix rounding errors in resizeWInc() and similar

gustavcedersjo:
Some applications change their window size to certain boundries, such as terminal emulators to fit a whole number of text columns. When resizing such windows with resizeWInc(), the window should grow in width to the next 10% of the total width. However, if the terminal emulator reduces that 10% to 9.9% to hit a whole number of text columns, the next call to resizeWInc() will just try to increase that last 0.1% and the terminal emulator will immediately take it back to hit a whole number of columns.

This patch changes the behaviour of resizeWInd() and friends to increase the width with something between 5% and 15%, whatever hits a tenth of the full width, instead of something between 0% and 10%.

### DIFF
--- a/tiling.js
+++ b/tiling.js
@@ -3494,7 +3494,7 @@ function resizeHInc(metaWindow) {
 
     let maxHeight = workArea.height - Settings.prefs.horizontal_margin * 2 - Settings.prefs.window_gap;
     let step = Math.floor(maxHeight * 0.1);
-    let currentHeight = Math.floor(frame.height / step) * step;
+    let currentHeight = Math.round(frame.height / step) * step;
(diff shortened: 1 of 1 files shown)


gustavcedersjo:
Solves #580

jtaala:
Nice find and fix @gustavcedersjo!

Interestingly, I've just released that I don't use these keybinds (I only really use the `Useful window widths` etc.).

Many thanks.  Fixes look good to me.

Merging.

jtaala (approved):
Tested and confirm changes fix issue.  I'm surprised I don't use these keybinds (got too lazy with custom useful window widths I setup).

jtaala:
Note: these changes will get merged in in our next release (which should be within the next week or so).

PR #582, opened 2023-08-22
Improvements to multi-monitor behaviour

jtaala:
This PR is a continuation of #577.

It is intended to address issues from #577 that may result from wider testing in `develop` branch.  It's also for making small improvements to some of the implementations there, including looking at adding several shortcuts to make things a little easier when working with multimonitors.

Previous commenters on #577 include @Lythenas, @YaLTeR, @smichel17.

Referencing a few comments here (hoping to address these in this branch/PR):

- https://github.com/paperwm/PaperWM/pull/577#issuecomment-1684968422
  - > What about, as a workaround, a way to move a workspace to a different monitor? Or swap two workspaces between monitors. So for now we don't worry about whether we sometimes get mixed up and put the wrong workspaces on the wrong monitors, because the user can fairly easily get them back where they belong. Maybe a shortcut, with no default keybinding.
- https://github.com/paperwm/PaperWM/pull/577#issuecomment-1688046344
  - > After using this with dynamic workspaces for a few days now, I think it really needs a shortcut to inject/insert a new workspace to the right of the current workspace (basically do what happens when you drag a window in overview between two of the workspace thumbnails in overview).


jtaala:
@YaLTeR, let me know if you're still getting lots of workspace 11's (or another number).

Some commands to reset workspace settings:

```
dconf reset /org/gnome/desktop/wm/preferences/workspace-names
dconf reset -f /org/gnome/shell/extensions/paperwm/workspaces/
```

That first ones interesting, it list workspace names that gnome will use - default is empty and gnome then uses the `Workspace x`.

YaLTeR:
Those dconf resets fixed the workspace 11 I had after a reboot, will see what happens next.

Btw the autostarted apps glitching out has been happening every single reboot after whichever commit started it.

YaLTeR:
Hmm, got quite a few glitched workspaces after a sleep/resume on this branch. No errors in the journal.

jtaala:
> Btw the autostarted apps glitching out has been happening every single reboot after whichever commit started it.

Hmmm - can you bisect this?

jtaala:
> Hmm, got quite a few glitched workspaces after a sleep/resume on this branch. No errors in the journal.

What does glitched workspaces mean?  Def let us know if you can get some reproduce steps.  

Anyone else seeing this?

YaLTeR:
> What does glitched workspaces mean?

The workspaces were empty in the overview animation, and when actually activated they had glitched windows behind everything else, similar to the autostart issue.

jtaala:
> similar to the autostart issue.

Thanks @YaLTeR.  Sounds related.  Give me an app to try in autostart (terminal?) and I'll work on reproducing to isolate.

YaLTeR:
Try terminal I guess. For me it's Telegram, Discord, Firefox and Syncthing, all from Flathub.

jtaala:
Cheers, just used brave (browser).  Definitely starts before tiling init and put the window in scratch (floating).  Had to tile it manually (default shortcut `Ctrl+Super+Escape`).  Is that what you're seeing?  Or something more glitchy?

YaLTeR:
Kind of, but not really, I was unable to successfully navigate to those windows as if they were floats.

jtaala:
Thanks @YaLTeR  - found it and fix pushed.

PR #592, opened 2023-08-30
feature: Cycle Backwards on the available widths and heights

kassick:
Adds new shortcuts that allow cycling backwards in the list of widths and heights

PR #598, opened 2023-09-05
Feat/cycle backwards

kassick:
Same as https://github.com/paperwm/PaperWM/pull/592 -- recreating after a force push closed the PR :/


kassick:
@jtaala I think this resolves the issues you pointed in the previous MR -- I had to recreate the PR as I did a force push to my fork and the original PR was auto-closed as a result :/

jtaala:
Hey @kassick, getting a few errors on this for the cycle heights:

```
Sep 06 09:30:03 jay-blade gnome-shell[31845]: JS ERROR: TypeError: findFn is not a function
                                              calcTargetHeight@/home/jay/.local/share/gnome-shell/extensions/paperwm@paperwm.github.com/tiling.js:3841:33
                                              allocate@/home/jay/.local/share/gnome-shell/extensions/paperwm@paperwm.github.com/tiling.js:3854:48
                                              layout@/home/jay/.local/share/gnome-shell/extensions/paperwm@paperwm.github.com/tiling.js:507:46
                                              cycleWindowHeightDirection@/home/jay/.local/share/gnome-shell/extensions/paperwm@paperwm.github.com/tiling.js:3865:19
                                              cycleWindowHeightBackwards@/home/jay/.local/share/gnome-shell/extensions/paperwm@paperwm.github.com/tiling.js:3826:12
                                              dynamic_function_ref/<@/home/jay/.local/share/gnome-shell/extensions/paperwm@paperwm.github.com/utils.js:106:33
                                              asKeyHandler/<@/home/jay/.local/share/gnome-shell/extensions/paperwm@paperwm.github.com/keybindings.js:256:51
```

I'll dig into it later and get some reproduce steps for you.

Thanks!

jtaala:
Ah, got it, it's a typo - should be `findPrev` not `FindPrev`.

tiling.js
```diff
@@ -3797,20 +3818,30 @@ function cycleWindowWidth(metaWindow) {
...
+}
+
+function cycleWindowHeightDirection(metaWindow, direction) {
     let steps = Settings.prefs.cycle_height_steps;
     let frame = metaWindow.get_frame_rect();
 
     let space = spaces.spaceOfWindow(metaWindow);
     let i = space.indexOf(metaWindow);
 
+    let findFn = direction === CycleWindowSizesDirection.FORWARD ? Lib.findNext : Lib.FindPrev;
```
jtaala: typo, should be:

```
    let findFn = direction === CycleWindowSizesDirection.FORWARD ? Lib.findNext : Lib.findPrev;
```

Notice the non-capitalised `Lib.findPrev`

jtaala (changes requested):

tiling.js
kassick: ouch!

I've realized I had only updated the height cycling function -- not only was it wrong, but incomplete ...

Good thing you tested it again!

kassick: 27e5c927318a8dfd4c7a0534d6051b7bb03ef1e6 should fix it

(I'll test as soon as I can restart my session)

kassick: Done

jtaala: Cool, will give it a another test later on today and then merge into develop (and do a release with both your PRs).

Cheers.

schemas/org.gnome.shell.extensions.paperwm.gschema.xml
```diff
@@ -217,11 +217,21 @@
       <summary>Cycle through useful window widths</summary>
     </key>
 
+    <key type="as" name="cycle-width-backwards">
+      <default><![CDATA[['<Super><Alt>r']]]></default>
```
smichel17, line 221: I think shift might be a better modifier for reversing direction, for consistency with other keybinds

jtaala: Yeah, interestingly, in my tests I changed mine to shift+alt+e (to match my alt+e for widths).

kassick: I agree -- I intended to use shift as direction modifier, but did not want to change the defaults. Currently, Super+Shift+R is the default keybindings for cycle heights...

What you folks think, should we change these defaults and use shift as backwards and alt for height?

jtaala: Ah good point!

Changing well-established defaults (which users might be used to now and haven't changed) is always difficult.

My thoughts are that we want to keep default keybinds (especially if they've been there a long time).  So, I suggest leaving as you originally done @kassick - users can change the new keybinds to their preference.

Thoughts @smichel17?

We can change a default keybind - I guess if we just document/record why we changed it.  If we changed the default for cycle heights - any ideas of what to change the new one to?

jtaala: Ah - I reckon let's change that sucker (cycle heights), it should probably get it's own key anyways.

I like consistency, and I use (not that it matters here) `alt+e` for cyle width and `alt+q` for cycle heights, and the backwards versions are just appended by `shift`.

Any thoughts on what to change height to?  `alt+t`?

kassick: I would avoid using `Alt+$LETTER` as default for anything to avoid clashing with app-specific keybindings (not that I use `transpose-words` that often in emacs, but I'd rather not lose arbitrary keybindings when default change ;) )

I had initially though of
- `<Super>+r` and `<Super>+<Shift>+r` for cycle width
- `<Super>+<Alt>+r and <Super>+<Alt>+<Shift>+r` for cycle heights

On using `r` and `t` because they are side-by-side: Currently, `<Super>+t` is `take window`, so `<Super>+R` and `<Super>+T` for the cycle would require changing the default for take -- and IMHO the current default is good.

On using `w` and `h` for _width_ and _height_: `<Super>+w` would clash with the Switcher extension and `<Super>+h` would annoy weird people like me who like to use vim-like `<Super>+hjkl` movements.

jtaala: > Any thoughts on what to change height to? `alt+t`?

Oops - meant to say `<Super>+t` - but as you said that's taken.

jtaala: > and IMHO the current default is good.

Agree - unless there's a good alternative to cycle heights, happy to leave it as it currently is.

jtaala (approved):

jtaala:
Merged into develop - thanks for your work on this @kassick.

PR #610, opened 2023-09-30
FIX: `Super+Tab` window previews now scale minimap-scale preference

jtaala:
This PR sets LiveAltTab windows previews (default keybind in PaperWM is `Super+Tab`) to scale according to minimap-scale user preference - i.e. they better match that size preference.

Also fixes the case where a user might set the `minimap-scale` size to 0 (which disables minimap).  Currently this breaks the stackoverlay edge window previews (i.e. they don't show).  In this case, previews and alttab will still show (with default sizing) when minimap is disabled by setting it to 0.

Hence, this PR allows users who want a bigger window preview to have it... while also allowing disabling the minimap and keeping the current/default liveAltTab window previews.


Lythenas (commented):
This makes it so `0` no longer turns off the minimap (as the descriptions in the settings says).

So either we should still hide the alt tab preview when set to `0` or we should remove that comment. But some users might want to disable the preview.

Otherwise this seems fine.

Lythenas:
Also maybe we just want a separate setting for the edge preview and for the alt tab preview.

jtaala:
> This makes it so `0` no longer turns off the minimap (as the descriptions in the settings says).
> 
> So either we should still hide the alt tab preview when set to `0` or we should remove that comment. But some users might want to disable the preview.
> 
> Otherwise this seems fine.

No, 0 still turns off minimap (it should).

jtaala:
> This makes it so `0` no longer turns off the minimap (as the descriptions in the settings says).

Just tested again.  0 does indeed turn off minimap (are you seeing otherwise)?

jtaala:
Super+tab preview always worked previously also when 0 was set for minimap, e.g. previously there was no relationship between minimap size and super-tab preview.

Now it just also can be resized to match the preferred minimap size (previously was static size and couldn't be changed).

Lythenas:
> No, 0 still turns off minimap (it should).

Ah I think I misunderstood. The minimap is what shows up when during navigation. That is indeed hidden when set to 0. But the alt tab preview and the edge preview both still show.

Then this might be fine (not sure if could be some cases where the alt tab preview or the edge preview get strange sizes). But I think we should at least update the description in the preferences to say that it also affects the scale of the alt tab preview and the edge preview.

jtaala:
> So either we should still hide the alt tab preview when set to `0` or we should remove that comment. But some users might want to disable the preview.

Ah, just to resolve any confusion - minimap shows up whenever using keybinds for `Switch to the .... window`, where `Super+Tab` switching (which this PR improves) is for `Switch to previously active window` and `Switch to previously active window, backward order`.

jtaala:
> Ah I think I misunderstood. The minimap is what shows up when during navigation. That is indeed hidden when set to 0. But the alt tab preview and the edge preview both still show.

Yeah, that's it.  Previously `super`+`tab` just did it's thing and the size couldn't be changed - so was always there regardless of minimap size (0 or otherwise).  Also previously, setting minimap to 0 also had the side-effect of breaking the edge previews.

jtaala:
> Also maybe we just want a separate setting for the edge preview and for the alt tab preview.

This is another option - could add two more options in prefs (or keep them hidden prefs?) to set (including disabling via `0`).  Just trying to avoid too many granular options, but would make it quite flexible - what do you reckon?

Lythenas:
> This is another option - could add two more options in prefs (or keep them hidden prefs?) to set (including disabling via `0`). Just trying to avoid to many granular options, but would make it quite flexible - what do you reckon?

Hidden options would be fine. And I think adding options for things like this is fine. It doesn't really make the code more complicated you are just reading a pref for the specific thing instead of one that applies to multiple things.

PR #611, opened 2023-09-30
Add preference for edge window preview size and window-switcher preview size

jtaala:
This PR adds several more user preferences which allow users to change the size of:
 - tiling edge window previews;
 - window-switcher (default keybind `Super+Tab`) previews.

image


jtaala:
Hey @Lythenas,

Have  looksy at this one.  Yes, it's much nicer having individual options here for users to set to their liking (code is cleaner too).

Lythenas:
Looks good. If it is possible I would maybe also add add that minimap is shown during window navigation (or something like that) and the window switcher preview is shown during alt tab (or super tab).

Not sure if that will fit into the UI, so maybe as a tooltip would be fine.

jtaala:
> Not sure if that will fit into the UI, so maybe as a tooltip would be fine.

Yeah, won't fit in the UI but added tooltips.

Lythenas (approved):
Looks perfect.

PR #621, opened 2023-10-08
Allow user to change the percent of workspace width `Super+F` (maximize window takes)

### LINKED ISSUE #620: allow maximize width (super f) to take up less than 100% width
i like to know where i am among my windows, but when i maximize their width i can no longer see anything of the surrounding windows

it would be nice if the user could configure the max width to allow the user some freedom in this regard

- settle for not knowing what's around on max width
- super-f-minus

**Additional context**
cf. seemingly opposite request #585

jtaala:
Fix #620.

This PR adds a user preference that allows users to set a value other than `100%` for the width the `Maximise the width of the active window` keybind uses (`Super+F`).

See #620 for a use case where this might functionality would be useful.

This allows setting the width to "maximize" to: `0.5` <= _number_ <= `1.0`, for example, you could set it to 0.95 by running the following from a terminal (or in dconf editor):

```
dconf write /org/gnome/shell/extensions/paperwm/maximize-width-percent 0.95
```


Lythenas (approved):
Tested in VM seems to work fine.

PR #643, opened 2023-10-24
Added touch scroll gestures support

### LINKED ISSUE #363: [Feature request] Touch gestures for scrolling through workspace?
First off, awesome extension, I use it everyday and have grown addicted to its workflow! I'm using a convertible laptop, and was wondering if it was possible to add a way to scroll through windows in a workspace with touch events?

The way I'd envision it would be to handle scrolling on the topbar, or along the top and bottom edges of the screen (I use Dash to Panel).

I'm not well-versed enough in GJS to really help, sorry

Thesola10:
Rebase of the previous PR. Now with GNOME 45's new gesture handling, swiping and leaving the panel no longer freezes workspace scrolling :D

### DIFF
--- a/tiling.js
+++ b/tiling.js
@@ -24,8 +24,6 @@ export let spaces;
 
 let borderWidth = 8;
 
-var prefs = Settings.prefs;
-
 // Mutter prevints windows from being placed further off the screen than 75 pixels.
 export const stack_margin = 75;
 
@@ -1893,9 +1891,10 @@ export const Spaces = class Spaces extends Map {
 
         // Initialize spaces _after_ monitors are set up
         this.forEach(space => space.init());
+
         // Bind to visible workspace when starting up
         signals.disconnect(Main.panel);
-        signals.connect(Main.panel, "captured-event", Gestures.horizontalTouchScroll.bind(this.get(workspaceManager.get_active_workspace())));
+        signals.connect(Main.panel, "captured-event", Gestures.horizontalTouchScroll.bind(this.activeSpace));
 
         this.stack = this.mru();
     }
--- a/tiling.js
+++ b/tiling.js
@@ -1893,7 +1893,7 @@ export const Spaces = class Spaces extends Map {
         this.forEach(space => space.init());
 
         // Bind to visible workspace when starting up
-        signals.connect(Main.panel, "captured-event", Gestures.horizontalTouchScroll.bind(this.activeSpace));
+        this.touchSignal = signals.connect(Main.panel, "captured-event", Gestures.horizontalTouchScroll.bind(this.activeSpace));
 
         this.stack = this.mru();
     }
@@ -2260,8 +2260,8 @@ export const Spaces = class Spaces extends Map {
         }
 
         // Update panel to handle target workspace
-        signals.disconnect(Main.panel);
-        signals.connect(Main.panel, "captured-event", Gestures.horizontalTouchScroll.bind(toSpace));
+        signals.disconnect(Main.panel, this.touchSignal);
+        this.touchSignal = signals.connect(Main.panel, "captured-event", Gestures.horizontalTouchScroll.bind(toSpace));
 
         inPreview = PreviewMode.NONE;
     }
--- a/tiling.js
+++ b/tiling.js
@@ -1893,7 +1893,6 @@ export const Spaces = class Spaces extends Map {
         this.forEach(space => space.init());
 
         // Bind to visible workspace when starting up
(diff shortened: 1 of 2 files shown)


jtaala:
Hey @Thesola10, I created a PR to your repo with some (small) changes (mostly cleanup and formatting) for this.  See https://github.com/Thesola10/PaperWM/pull/1

Thesola10:
Thanks, will check and merge asap

~~I'd try and tweak gesture cancel as it interferes with Media Controls (and Media Controls alone, maybe because of nested buttons?)~~ but yeah, should be ready to merge

okay, the issue was that Media Controls listens to `TOUCH_BEGIN` and not `TOUCH_END`, I should know, I'm the one who wrote the touch support PR over there

jtaala:
Hey @Thesola10 - can you change the PR to merge in your `touch-scroll` branch instead of your `develop` branch?(which is behind your `touch-scroll` branch).

Thanks!

Thesola10:
> Hey @Thesola10 - can you change the PR to merge in your `touch-scroll` branch instead of your `develop` branch?(which is behind your `touch-scroll` branch).
> 
> Thanks!

I wish, otherwise I wouldn't have had to recreate this PR

I'll push `develop` to match `touch-scroll` when I can.

Thesola10:
and, done!

btw, I checked "allow edits from maintainers", apparently that allows you to directly push to my `develop` branch

tiling.js
```diff
@@ -1872,6 +1872,11 @@ export const Spaces = class Spaces extends Map {
 
         // Initialize spaces _after_ monitors are set up
         this.forEach(space => space.init());
+
+        // Bind to visible workspace when starting up
+        signals.disconnect(Main.panel);
```
jtaala: We don't need to disconnect signals here.  `init` method is called on PaperWM startup - and all signals are destroyed when PaperWM is disabled (including during lockscreen).

tiling.js
```diff
@@ -2233,6 +2238,10 @@ export const Spaces = class Spaces extends Map {
             monitor.clickOverlay.activate();
         }
 
+        // Update panel to handle target workspace
+        signals.disconnect(Main.panel);
```
jtaala: We don't need to disconnect signals here.  `init` method is called on PaperWM startup - and all signals are destroyed when PaperWM is disabled (including during lockscreen).

jtaala (changes requested):
Hey @Thesola10,

Looks okay.  One concern here is that this touchscreen code (which intercepts all captured-events) will be run regardless of whether users have a touchscreen capable monitor or not.

In `gestures.js` for example, we don't enable gestures for X11 (since we can't use gestures there).  Similarly, we have some code that gets enabled only is there is multiple monitors (it tracks mouse cursor movements to see which monitor it's on).

Can you see a way to do something likewise?  E.g. some check to see whether there is a touchscreen involved, and only then enable this touchscreen gesture support?

jtaala:
> ~~Can you see a way to do something likewise? E.g. some check to see whether there is a touchscreen involved, and only then enable this touchscreen gesture support?~~

Actually, looking at it - I'm **NOT** too concerned with this in the end - you've got a switch/case check for event types, that's sufficient enough.

jtaala:
> I'm  too concerned

That' was supposed to be:

>  **NOT** too concerned.

tiling.js
Thesola10: This disconnect is required because it changes the affected workspace when scrolling. The comment above I do agree with however

Thesola10:
Yeah, my signal handler only catches **touch** signals, so far I've never had issues with it on non-touch systems

jtaala:
It's a bit late now (in Aus) - but I'll give this one last test tomorrow and then I plan on approving this.

tiling.js
jtaala: I see, makes sense.  Could you instead replace it with:

```js
        // Update panel to handle target workspace
        if (this.touchId) {
            signals.disconnect(Main.panel, this.touchId);
        }
        this.touchId = signals.connect(Main.panel, "captured-event", Gestures.horizontalTouchScroll.bind(toSpace));
```
Give the the above a test.  It's more specific and will disconnect by id.  This way, if we ever want to connect another signal to `Main.panel` it won't also get disconnected.

jtaala (changes requested):
Looking fine.

Request to just make the `signal.disconnect` and little more targeted (see comment).

jtaala (approved):

Thesola10:
all green, merge when ready, and goodnight :)

Issue #638, opened 2023-10-18
Full screen Wayland apps show black screen and top left window corner near center of screen


CapSel:
**Describe the bug**
I tried to watch a movie in one of streaming services and I pressed F11 in firefox to go to full screen. I saw black screen and window's top left corner was near the middle of screen. Similar behaviour was when I pressed win+shift+f. Same behaviour with kitty (terminal emilator) and other apps.

The other strange behaviour - it's happening only to windows that are last on the right. For second (third...) to last windows maximization shows 1 frame with problem, but then it fixes itself. My Signal app with ozone/wayland enabled is not counted.

**To Reproduce**
1. Put window to maximize as the last window
2. Press win+shift+f to maximize window to full screen

**Expected behavior**
Full screen windows should not have top left corners near middle of the screen.

I use Arch Linux and newest PaperWM from extensions.gnome.org

jtaala:
Assuming Gnome 45?

Can you please take a video (even on a phone) or screenshot?

Following your steps, I haven't been able to reproduce this.

jtaala:
P.S. I'm also on Arch.

jtaala:
> 1. Put window to maximize as the last window

Can you elaborate here what do you mean maximize the window (as in PaperWM maximize horizontally?  or just like double-clicking the firefox topbar?).

CapSel:
> > 1. Put window to maximize as the last window
> 
> Can you elaborate here what do you mean maximize the window (as in PaperWM maximize horizontally? or just like double-clicking the firefox topbar?).

Firefox calls it "full screen" - available from View -> Full Screen.

![Screenshot from 2023-10-18 13-49-03](https://github.com/paperwm/PaperWM/assets/3890413/0d4c23a5-085c-4e08-82bf-ccf50de5d063)

CapSel:
I think gnome terminology for keyboard shortcut that does same thing is super+shift+f

jtaala:
Gotcha.  Hmm - I'm not seeing this.

To rule out other extensions (if you have any others), can you disable other extensions and see if this is still happening?

This is what I see (let me know if I'm doing something wrong):

https://github.com/paperwm/PaperWM/assets/30424662/6c077cc6-a521-40cc-8d0f-b518a0f08cb6

CapSel:
I've disabled all extensions and left only PaperWM - still happens.

jtaala:
> > > 1. Put window to maximize as the last window
> > 
> > 
[...]

PR #644, opened 2023-10-25
Fix 638 fullscreen animation window resize bug backport (for gnome 44)

jtaala:
#641 is the gnome-45 version that was merged recently.  This PR is a backport of this fix for gnome-44.

See https://github.com/paperwm/PaperWM/issues/638.

This PR fixes an animation race-condition issue that can cause a fullscreened window to be shown incorrect (only part of it is shown).

I could reproduce the issue by setting the animation-time to 0 or a high number. This PR fixes this case of the issue.

This also adds safe boundaries on animation-time to avoid several issues related to three-finger swiping the tile.

It's improved the situation for [@Capsel](https://github.com/CapSel) - but a (related) issue can still occur (which we haven't been able to reproduce).

### DIFF
--- a/Settings.ui
+++ b/Settings.ui
@@ -46,8 +46,8 @@
     <property name="page_increment">10</property>
   </object>
     <object class="GtkAdjustment" id="animation_time_adjustment">
-    <property name="upper">9.99</property>
-    <property name="lower">0.03</property>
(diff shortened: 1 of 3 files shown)


jtaala:
Closing as now part of #650

PR #645, opened 2023-10-25
Add keybindings to switch to ws from all monitors

### LINKED ISSUE #635: Switch to a workspace sticked to another monitor
I have 2 monitors setup and somehow all workspaces are moved to my right monitor. And there are no workspaces on the left monitor at all.
Is there a way to move (not swap) workspaces from one monitor to another?

Have a shortcut to **move** a workspace to a specific monitor.
Or make optional sticking a workspace to a monitor.

It is also interesting in what way the workspaces distributed between monitors by default (say if I have 5 workspaces and 2 monitors)?
I see that when a workspace has no windows I can switch to it from any monitor but once there is at least one window in a workspace the workspace sticks to one specific monitor, is this expected behavior? 
Would it possible to make this optional maybe?

Lythenas:
Fixes #635 

Adds two new keybindings:

- Switch to workspace below (ws from all monitors)
- Switch to workspace above (ws from all monitors)

by default they have no keys bound.

They work the same as "Switch to workspace below/above (ws only from current monitor)" (renamed from just "Switch to workspace below/above"). Except the user can switch to all workspaces (except the once currently active on another monitor).

Please test this.


Lythenas:
The animation at the beginning (i.e. when starting the switching) is a tiny bit strange if there are workspaces that were previously on other monitors. I think they ease into position from the other monitor. But it was a bit hard to tell in the VM. I will test tomorrow on my real device and create a recording.

jtaala:
Sounds good.  For animation stuff, it's often helpful to set the `animation-time` to something big (like 5.00 or 6.00) - makes it much easier to see what's going on.

jtaala:
Hey @Lythenas - this looks good to me!  Tested and didn't see/notice any weird animation stuff too.

Good job!

tiling.js
```diff
@@ -2252,8 +2252,25 @@ export const Spaces = class Spaces extends Map {
...
+        }
+        return out;
+    }
+
+    _getOrderedSpacesFromAllMonitors(monitor) {
+        let nWorkspaces = workspaceManager.n_workspaces;
+        let out = [];
+        for (let i = 0; i < nWorkspaces; i++) {
+            let space = this.spaceOf(workspaceManager.get_workspace_by_index(i));
+            if (this.monitors.get(space.monitor) !== space || space.monitor == monitor) {
```
jtaala: Only thing I can suggest (which is super minor) but using stricter equality for `space.monitor === montor`:

```
            if (this.monitors.get(space.monitor) !== space || space.monitor === monitor) {
```

But that's mainly so my linter doesn't complain about it.

Lythenas:
I slowed down the animation and to me it looks like if the workspace was previously on that monitor or is empty. The workspace is positioned below the current workspace, when the animation begins. If the workspace was previously on another monitor it is positioned kind of behind the current workspace. But it is also moved to the correct position and this is barely noticeable on normal animation speed.

To be honest I'm not sure how to fix that. And I'm fine with the current state.

jtaala (approved):
Hey @Lythenas,

I just made a couple of tweaks.  

Main one is not actually setting the space.monitor until the space is actually selected.  In other words, just switching (in sequence switching view) shouldn't actually set the monitor to that space - in case users have both sets of keybinds set, and might get confused why their space "moved" to another monitor when they didn't actually select the space on that monitor.  That's the theory anyway - let me know if that makes sense and what you think.

I also see the animation - I think it looks cool, and in the very least is a small visual signal that this space was on the previous monitor.

I don't mind it at all.  We could probably dig deep to try change that but I don't think it's worth it tbh.

Approving this.

jtaala:
Oh, now for the (maybe not fun?) part.  Can you throw a small section in the README.md re this stuff if you're happy with the code?

Lythenas:
Makes sense. I also noticed this but thought people *probably* won't use both keybindings. But this better.

Ah totally forgot about the readme. I will write something.

Lythenas:
I added something to the "The workspace stack & monitors" section in the readme. I hope it is understandable.

@jtaala feel free to merge if this is ok. It says it needs an additional approval, but I can't approve my own PR.

jtaala (approved):

jtaala:
Hey @Lythenas,

Just approved.  Looks good to me.  Can you give merging a go? (you should be able to merge - unless I messed the settings up).

Jay.

PR #654, opened 2023-10-28
Removed wrap property from keybinds list items

Thesola10:
This caused keybinds and window rules descriptions to wrap too early, making them unreadable.

### DIFF
--- a/KeybindingsRow.ui
+++ b/KeybindingsRow.ui
@@ -28,8 +28,7 @@
                 <property name="ellipsize">end</property>
                 <property name="halign">start</property>
                 <property name="lines">1</property>
-                <property name="wrap">True</property>
-                <property name="wrap-mode">word-char</property>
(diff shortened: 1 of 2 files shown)


Thesola10:
**Before** | **After**
-----------|-----------
image | image

jtaala:
I believe that wrapping issue can occur for font scaling factors != 1.00 (and maybe fractional scaling?).

Lythenas:
I'm a bit confused why one string is still shown very long while the others are cut off early. But removing the wrapping does not change anything on the long string so I think this is fine. But not sure if there is a *proper* fix for the issue.

jtaala:
Just tested this with font scale factor at 0.90 (in gnome tweaks) and this change does indeed this the truncating issue from scale factors.  Accepting.

jtaala (approved):

PR #676, opened 2023-11-08
FIX for issue #674: moving window to workspace stopped working

### LINKED ISSUE #674: Shortcut "Move window one workspace up/down" doesn't work
**Describe the bug**

When I assign a key and focus on a window, upon activation of the shortcut I just switch from the wall view to the up/down workspace, without moving the actual window.

**To Reproduce**
1. Set the shortcut
2. Trigger shortcut

**Expected behavior**

To be able to move the active window to the up/down workspace

**Screenshots**

**System information:**

I have the gnome extension

**Additional context**

Bug is present in both Xorg and Wayland, albeit in wayland I found a lot of things breaking, so I am trying to get this working on Xorg mainly.

jtaala:
This PR fixes #674 (in gnome 45).

### DIFF
--- a/tiling.js
+++ b/tiling.js
@@ -2388,7 +2388,7 @@ export const Spaces = class Spaces extends Map {
 
         if (move && this.selectedSpace.selectedWindow) {
             const navigator = Navigator.getNavigator();
-            if (navigator._moving === null ||
+            if (navigator._moving == null ||
                 (Array.isArray(navigator._moving) && navigator._moving.length === 0)) {
                 takeWindow(this.selectedSpace.selectedWindow,
                     this.selectedSpace,


tiling.js
```diff
@@ -2388,7 +2388,7 @@ export const Spaces = class Spaces extends Map {
 
         if (move && this.selectedSpace.selectedWindow) {
             const navigator = Navigator.getNavigator();
-            if (navigator._moving === null ||
+            if (navigator._moving == null ||
```
Lythenas, line 2391: Can you also add a comment here that this needs to be `==` instead of `===`? Otherwise we might break it again in the future on accident.

jtaala: Yeah, good idea - I'll stick a comment in develop.

jtaala: Actually, a much better (cleaner, simpler to avoid confusion) is just to use

```
            if (!navigator._moving ||

```

like we do here: https://github.com/paperwm/PaperWM/blob/4ee4d466047972636d7a7dbb3b943f6598fa4370/tiling.js#L4321

jtaala: Line 4321 is from the `takeWindow` function which (which the function in question calls) and actually does this same check (to create the `_moving` array).  So it makes sense to align these two and to avoid confusion here.

I'll make this change and re-test and get it in develop to avoid future occurrences of this issue.

PR #679, opened 2023-11-10
Feature 526 open new window position gnome45

### LINKED ISSUE #526: open new window to left of focused window
I find that often I want to open a new window to the left of the focused window. Since text goes left to right, the left side of a window like a browser or a terminal often has more useful information in it than the right side does. When the two windows can't fit on the screen together, opening a new window to the right will prevent referring back to the original window as effectively as if the new window were on the left. 

Currently I have to open the window and then move it to the left. That sounds easy, but it's kinda not. Really the process is: Press key that will open the new window. Wait some amount of time for it to appear. Screen shifts right. Multifinger keypress to move it to the left. Screen shifts left. The variable delay and screen movement make this significantly more taxing than if the window could be opened to the left. It's a small papercut, but one that happens repeatedly thruought my day.

I'd be ok with an option that changes all windows to default to open to the left rather than the right.

Since super-I also pulls from the right, any change would also need to affect it presumably.

jtaala:
Completes #526 by adding options to set where to insert new windows (e.g. to the `left` / `right` of windows, or at `start` / `end` of tiling):

image


tiling.js
```diff
@@ -38,6 +38,9 @@ export const FocusModes = { DEFAULT: 0, CENTER: 1 }; // export
 
 export const CycleWindowSizesDirection = { FORWARD: 0, BACKWARDS: 1 };
 
+// position to open window at (e.g. to the right of current window)
+export const OpenWindowPositions = { RIGHT: 0, LEFT: 1, START: 2, END: 3 };
```
Lythenas: Since you have this as basically an enum it would be ideal to refer to this everywhere, where you currently just write 0, 1, 2, 3.

But this is just a nitpick and I don't know if that would actually be convenient/possible.

Lythenas (approved):
I tried this out and it works as expected.

tiling.js
jtaala: This is a good call.

The reason why I couldn't use in `prefs.js` is because it's was defined in `tiling.js` and we can't import `tiling.js` into `prefs.js` (it has too many deps that depend on gnome-shell thread - note pref.js doesn't run on gnome thread).

In any case, could actually do it if define the enum in `settings.js` (this module is lightweight and doesn't have strong deps so can be used in extensions.js and prefs.js.

I think it's worth doing.

jtaala:
@Lythenas what do think about this?

> Since super-I also pulls from the right, any change would also need to affect it presumably.

from https://github.com/paperwm/PaperWM/issues/526#issue-1716193077

Should we also change the side that we pull windows from (e.g. make it pull windows from the left?).  I'm thinking probably should for consistency?

Lythenas:
> Should we also change the side that we pull windows from (e.g. make it pull windows from the left?). I'm thinking probably should for consistency?

Would it then also pull from start or end? Not sure how much sense that makes. But I don't really use columns, so not sure.

jtaala:
> Would it then also pull from start or end? Not sure how much sense that makes. But I don't really use columns, so not sure.

Yeah, that does seem a bit weird - in the case of start/end I think it would make sense to just pull from left/right respectively.

tiling.js
jtaala: See latest commit.

Lythenas:
Then we should maybe just add one new keybinding for it instead of using this setting, since it does not really match up with the 4 options anymore.

jtaala:
> Then we should maybe just add one new keybinding for it instead of using this setting, since it does not really match up with the 4 options anymore.

Yeah, I agree - that would make more sense here.

Something like a `Consume the windows to the _left_ into the active column`.

jtaala:
I'll merge this one in and we can create an enhancement issue to add a consuming from left keybind.

Issue #682, opened 2023-11-11
Can't scroll when switching workspaces


vvvvvvitor:
**Describe the bug**
After switching workspaces, I can't scroll anymore.

**To Reproduce**
1. Open the GNOME overview
2. Switch workspace
3. Go back to the previous workspace
4. Can't scroll

**Expected behavior**
For the scrolling to work normally.

**Screenshots**
Couldn't capture any footage as it's really inconsistent.

**System information:**
```
Distribution: Fedora Linux
GNOME Shell 45.1
PaperWM branch/tag: v45.3.3
Enabled extensions:
- paperwm@paperwm.github.com
```

jtaala:
Hey @vvvvvvitor.  

I can't reproduce this and haven't seen this.  Let's put a tag on this to see if anyone else is seeing this also.

jtaala:
Also,

1. is PaperWM the only extension you have enable?
2. are you on X11 or Wayland?

jtaala:
Hey @Lythenas and @Thesola10,

You ever seen this?

Lythenas:
No I don't think I've seen this before.

@vvvvvvitor with "scrolling" do you mean normal mouse scrolling in a window or do you mean navigating the windows?

vvvvvvitor:
> Also,
> 
> 1. is PaperWM the only extension you have enable?
> 2. are you on X11 or Wayland?

Yes and I'm on Wayland.

vvvvvvitor:
I'll

vvvvvvitor:
Accidentally closed this, sorry. I'll try to get a video of it happening, although I've only managed to make it happen 2 times, basically what happens is that instead of scrolling the workspace, it switches the workspace.

vvvvvvitor:
> No I don't think I've seen this before.
> 
> @vvvvvvitor with "scrolling" do you mean normal mouse scrolling in a window or do you mean navigating the windows?

Navigating windows.

vvvvvvitor:
Managed to make it happen! File size ended up being too big so I had to upload it to youtube.
https://youtu.be/_R0Ymp7M4w0

jtaala:
Ah (@Lythenas) - that's why, we're talking about _gnome_ workspace switching! (as opposed to PaperWM space/workspace switching).  I don't use gnome's workspace switching (PaperWM space switching was intended to used instead of gnome workspace switching in PaperWM btw).

[...]

PR #698, opened 2023-11-15
Fix 682 tiling gesture can break on gnome workspace switching (for Gnome 44)

jtaala:
Fixes https://github.com/paperwm/PaperWM/issues/682 (for Gnome 44).

Review and refactor of gesture code.


jtaala:
Is there anyone on Gnome 42-44 (Wayland)  that can test this PR?

I have a virtual machine for Gnome 44 tests - but touchpad gestures don't work there (events get sent to my host machine).  Would appreciate it if some can help test this fix port!

```
git clone https://github.com/paperwm/PaperWM.git
cd PaperWM
git checkout fix-682-tiling-gesture-can-break-on-gnome-workspace-switching
./install.sh
```
then logout/login, and test touchpads gestures are working correctly.

Rutherther:
Hello, I am using Gnome 44, and I tried testing this. I hope I did everything correctly - I don't know of way to check that I am indeed using code of this branch, I can check that `~/.local/gnome-shell/extensions/paperwm@paperwm.github.com` is indeed on this branch, and that I have relogged, but from that point on its just based on trust to Gnome actually using that one instead of older one.

And yes, it's working fine, I am not using PaperWM for long time, but I've tried gestures before and then in this branch, both seem to be fine, and the same.

I unfortunately don't know how would I go about reproducing the issue that is being solved, as I probably don't understand what they were doing to trigger it. So I haven't tested it actually fixed it in Gnome 44.

jtaala:
Thanks @Rutherther - sounds fine.  Was just wanting a sanity check and that this branch would work okay.

PR #705, opened 2023-11-20
Add advanced settings tab in preferences (with other PaperWM settings)

jtaala:
This PR adds an `Advanced` tab to preferences and exposes nearly all other PaperWM settings.

Recently we've been receiving lots of questions about how to do things like "disable topbar styling" and other things.  I got sick of continually pointing people to the README.md.

It's just easier to expose these settings in preferences so they can see available settings.

I've also moved settings around to simplify the preference pages, see below:

image
image
image


prefs.js
```diff
@@ -74,32 +75,55 @@ class SettingsWidget {
...
+                this._settings.set_boolean(key, !state);
+            });
+        };
 
-        let hMargin = this.builder.get_object('hmargin_spinner');
-        hMargin.set_value(this._settings.get_int('horizontal-margin'));
-        hMargin.connect('value-changed', () => {
-            this._settings.set_int('horizontal-margin', hMargin.get_value());
-        });
+        const setValue = (builderKey, settingKey) => {
```
Lythenas: Maybe change this to `setValueInt` since we also have string values.

Also `setValue*` is a bit misleading for these functions since they don't actually set the value. Maybe call them similar to the boolean functions. E.g. `intValueChanged`, `doubleValueChanged`, `percentValueChanged`.

prefs.js
```diff
@@ -380,6 +374,65 @@ class SettingsWidget {
...
+            switch (setting) {
+            case 3:
+                builder.set_active_id('three-fingers');
+                break;
+            case 4:
+                builder.set_active_id('four-fingers');
+                break;
+            default:
+                builder.set_active_id('fingers-disabled');
+            }
```
Lythenas: I think this is slightly better using a lookup table. E.g.

```suggestion
            const lookupTable = {
                3: 'three-finters',
                4: 'four-fingers'
            };
            builder.set_active_id(lookupTable[setting] ?? 'finger-disabled');
```

You can also generate the inverse lookup table with e.g.

```
Object.fromEntries(Object.entries(obj).map(a => a.reverse()))
```

or write it by hand (which might be slightly less magic^^).

Same goes for the other switch statements.

Lythenas (commented):
Tested and works fine.

I added some comments which are mostly nitpicks. I'm also fine with merging this as is if you disagree.

Thesola10 (approved):
What can I say, it doesn't explode, LGTM

but more seriously, maybe move `Swipe Sensitivity` settings to Advanced? most users would be okay with the defaults and those nondescript float values could be overwhelming.

Also, I'd move the remaining `Touchpad Gestures` _above_ `Gaps and margins`, seeing as the default window size currently makes the touchpad section virtually invisible.

prefs.js
jtaala: don.e

prefs.js
jtaala: Done - yeah, is much more succinct.  Thanks!

jtaala:
> but more seriously, maybe move `Swipe Sensitivity` settings to Advanced? most users would be okay with the defaults and those nondescript float values could be overwhelming.

Yeah, I've never actually changed those (actually, I think I did accidentally once and then just reset them via dconf :-)

PR #710, opened 2023-11-25
Warp pointer when switching to workspace

### LINKED ISSUE #704: Switch to already visible workpace
When I switch to workspace that is already on my screen, but on a different monitor, it does nothing. I am using gnome keybinding to switch to different workspace. When I switch to workspace that is not currently visible, it correctly brings me to the workspace on the monitor I have focus on.

I would love to have the focus switched to the monitor with the workspace when workspace is visible on different monitor. Alternatively, switching workspace would work as well.

As gnome doesn't have independent workspaces (so it's not expected to have more workspaces visible) out of the box, I didn't have much luck finding any extension that would do this.

Rutherther:
Attempt to implement/fix #704 

When switching to workspace that is either already visible on non-focused monitor, or is restored from previous state to non-focused monitor, cursor stays in the old workspace.

This warps the cursor to the monitor of workspace the user is switching to.

I actually stumbled upon a code that warped the cursor in switchWorkspace, as I accidentally first went to the master branch, and it has code for warping cursor inside of switchWorkspace. So PaperWM did have this functionality before, I am not sure why it was removed.

### DIFF
--- a/tiling.js
+++ b/tiling.js
@@ -2276,6 +2276,8 @@ export const Spaces = class Spaces extends Map {
         signals.disconnect(Main.panel, this.touchSignal);
         this.touchSignal = signals.connect(Main.panel, "captured-event", Gestures.horizontalTouchScroll.bind(toSpace));
 
+        Utils.warpPointerToMonitor(monitor);
+
         inPreview = PreviewMode.NONE;
     }
 


Rutherther:
Sorry I accidentally opened the PR for release branch.

PR #711, opened 2023-11-25
Warp pointer when switching to workspace

Rutherther:
Attempt to implement/fix https://github.com/paperwm/PaperWM/issues/704

When switching to workspace that is either already visible on non-focused monitor, or is restored from previous state to non-focused monitor, cursor stays in the old workspace.

This warps the cursor to the monitor of workspace the user is switching to.

I actually stumbled upon a code that warped the cursor in switchWorkspace, as I accidentally first went to the master branch, and it has code for warping cursor inside of switchWorkspace. So PaperWM did have this functionality before, I am not sure why it was removed.

### DIFF
--- a/tiling.js
+++ b/tiling.js
@@ -2276,6 +2276,8 @@ export const Spaces = class Spaces extends Map {
         signals.disconnect(Main.panel, this.touchSignal);
         this.touchSignal = signals.connect(Main.panel, "captured-event", Gestures.horizontalTouchScroll.bind(toSpace));
 
+        Utils.warpPointerToMonitor(monitor);
+
         inPreview = PreviewMode.NONE;
     }
 


jtaala:
Thanks @Rutherther!

Good attempt! and you're thinking is right here.

Although, unfortunately this adds a bug/regression where if you go into Gnome overview and click the second monitor, it will cause the cursor to be warped to the first monitor.

jtaala (changes requested):
Thanks @Rutherther!

Good attempt! and you're thinking is right here.

Although, unfortunately this adds a bug/regression where if you go into Gnome overview and click the second monitor, it will cause the cursor to be warped to the first monitor.

As well as warping behaviour by simply move the mouse to another monitor (which is not wanted).

In short, although warping mouse pointer is what you want to do - it wouldn't go in `switchWorkspace` method (which is called whenever workspace focusing changes) - it should go into a higher level calling function (e.g. when you use your gnome keybinds) to isolate warping only under the circumstance you want it to occur.

jtaala:
It would be great if this change was all that was needed here - but it's a bit more involved.

Feel free to keep experimenting and getting familiar with the codebase to try this here.

I'll look at this when I can - although I don't use those gnome-keybinds and am working on a few other things atm.

jtaala:
I'll switch this to a draft in the meantime.

Rutherther:
> Although, unfortunately this adds a bug/regression where if you go into Gnome overview and click the second monitor, it will cause the cursor to be warped to the first monitor.

Ah, right, gnome treats that as the same workspace, that didn't occur to me. Alright, I will try to look into that soon.

Rutherther:
So sorry for taking so long, I have a lot of stuff going on right now, I think I will be able to look into it at the end of December.

Rutherther:
It's not just warping when clicked on monitor in the overview. I also found another bug. When I connect Wacom tablet, and map it to single monitor, I cannot get my cursor out of the monitor with mouse. It just warps back to the same monitor.

After I hopefully get around to this, this is something to test after it's reimplemented.

Rutherther:
I am sorry, I won't be able to continue with this in any foreseeable future. Closing this PR.

ankostis:
Do you have any more feedback on the Wacom tablet issue?

Rutherther:
I am not sure. This issue was introduced only in this PR that I was unable to fix. If you are having this issue without using code from this PR that's likely different issue. I am not using PaperWM anymore, so I am unable to help.

jtaala:
> I am not sure. This issue was introduced only in this PR that I was unable to fix. If you are having this issue without using code from this PR that's likely different issue. I am not using PaperWM anymore, so I am unable to help.

Thanks @Rutherther for the reply!

I don't have a wacom tablet so can't progress or test this one.  If anyone has a wacom tablet and wants to jump in a get familiar with the codebase let us know!

Jay.

schrieveslaach:
@Rutherther, @jtaala, I'm using paperwm with the patch of this PR and I can report that the issue with the wacom tablet is still present on Gnome 43 (Debian based distro). However, on Arch with the most recent Gnome version (48), the patch works fine with and without Wacom tablet.

schrieveslaach:
@jtaala, it seems to me that @Rutherther isn't active on GitHub anymore when I look into its profile. What do you think is a good way to fix #704?

Rutherther:
I am still somehow active on GitHub, but my interest in PaperWM has vanished, so I can't really help here anyway. :(

jtaala:
> @Rutherther, @jtaala, I'm using paperwm with the patch of this PR and I can report that the issue with the wacom tablet is still present on Gnome 43 (Debian based distro). However, on Arch with the most recent Gnome version (48), the patch works fine with and without Wacom tablet.

Hey @schrieveslaach.  Unfortunately, I'm no longer actively developing / working on PaperWM (see this issue https://github.com/paperwm/PaperWM/issues/980).

schrieveslaach:
@Rutherther, @jtaala, thanks for the heads up and sorry for bothering too much.

FYI: with #1045 I found a solution that fixes the setup (without extending PaperWM).

Issue #652, opened 2023-10-28
Cleanly fade out tiled windows when opening Overview


Thesola10:
**Is your feature request related to a problem? Please describe.**
Currently when opening the full screen GNOME overview, as soon as the pan-out animation starts, the tiled windows flash out of view, if the `Show scratch windows in overview` option is set to `Only`

**Describe the solution you'd like**
Fading out tiled windows to make opening the overview easier on the eyes.

**Describe alternatives you've considered**
Setting `Show scratch windows in overview` to another value: The overview gets cluttered and messy when tiled windows are included and many windows are displayed.

Not using the Overview: While I do use Arc Menu in desktop mode, touch functionality is much easier with the Overview active, and it can also help with switching workspaces.

jtaala:
Wait, do you mean non-tiled window? (e.g. scratch windows).

Tiled windows shouldn't "flash out of view", e.g:

https://github.com/paperwm/PaperWM/assets/30424662/c12b56eb-295c-4dcd-84ff-ddaddfabfa05

jtaala:
I had assumed you meant scratch windows when they are set to `Show scratch windows in overview` = `Never` in paperwm settings.

Thesola10:
I actually should have specified, I'm talking about the specific "Show scratch windows in overview" = "Only" setting

jtaala:
Oh, okay, I understand now.

PR #715, opened 2023-11-26
Enhancement #672: allow users to change overview exit transition animation

### LINKED ISSUE #672: Disable Overview transition animation on window selection
Sometimes selecting a window from the overview is much faster than navigating using the keyboard. But each time I select a window from the overview when using PaperWM, the overview translation animation runs which gets disorienting after multiple uses.

Instead of animating the translation to the window selected from the overview, there can be an option to just fade in the window.

- Speeding up the animations for the GNOME shell
- Disabling animations
- Speeding up animations in PaperWM

**Additional context**
https://github.com/paperwm/PaperWM/assets/26873863/9778a130-a13d-4d54-b51b-ad9fb07d22f1

jtaala:
This PR fixes #672.

Namely, it now allows users to select from PaperWM settings the animation used to ensure the selected window (after exiting Gnome Overview) is brought into view:

image


Lythenas:
Is the fade supposed to show the workspace background and then fade in the window? That seems a bit strange to me and not really better than no animation.

jtaala:
> Is the fade supposed to show the workspace background and then fade in the window? That seems a bit strange to me and not really better than no animation.

No, it's pretty much like no animation but a little softer (it just fades window out on exiting overview, and fades it back in).  This was the preferred one by the original submitter:

https://github.com/paperwm/PaperWM/issues/672#issuecomment-1801508556

Lythenas (approved):
Then I'm fine with this.

Settings.ui
```diff
@@ -1114,6 +1114,48 @@
...
                       </object>
                     </child>
+                    <child>
+                      <object class="GtkListBoxRow">
+                        <property name="activatable">False</property>
+                        <property name="focusable">False</property>
+                        <child>
+                          <object class="GtkGrid">
+                            <property name="focusable">False</property>
+                            <property name="tooltip_text" translatable="yes">Sets the animation when exiting Gnome overview and ensuring selected window is in view</property>
```
Thesola10:
```suggestion
                            <property name="tooltip_text" translatable="yes">Sets the animation when exiting GNOME overview and ensuring selected window is in view</property>
```

schemas/org.gnome.shell.extensions.paperwm.gschema.xml
```diff
@@ -449,6 +449,11 @@
         <summary>Duration of animations in seconds</summary>
     </key>
 
+    <key type="i" name="overview-ensure-viewport-animation">
+        <default>1</default>
+        <summary>Sets the animation when exiting Gnome overview and ensuring selected window is in view. 0:NONE, 1:TRANSLATE, 2:FADE</summary>
```
Thesola10:
```suggestion
        <summary>Sets the animation when exiting GNOME overview and ensuring selected window is in view. 0:NONE, 1:TRANSLATE, 2:FADE</summary>
```

Thesola10 (approved):
oh, this only applies if tiled windows are shown in Overview

could this be used to address #652 as well?

jtaala:
> oh, this only applies if tiled windows are shown in Overview
> 
> could this be used to address #652 as well?

No, unfortunately that would be a separate issue.  These changes don't effect the actual overview animations - they happen after the overview animation has finished (when bring the selected window into tiled view).

jtaala:
> > could this be used to address #652 as well?

The issue with #652 is that it's the default gnome behaviour (to just dissappear windows that normally wouldn't display in overview).  E.g. here's with PaperWM disabled (note I've disabled attaching modal dialogs to windows):

https://github.com/paperwm/PaperWM/assets/30424662/b8d01bf1-271f-417e-aa1c-827272804f71

So to change #652 we'd need to override or modify gnome overview animation (entry and exits).

Thesola10 (approved):

Issue #700, opened 2023-11-16
multiple monitors: strange behavior


stefangweichinger:
Fedora 39 system, with 2 monitors attached.

I have several questions and still try to work out things.

* when I press `Super` I see the same windows 2 times on both monitors. I expect to see the windows of each monitor only on each monitor ... 
* It seems I don't fully understand the "Scratch layer" (yes, I read the README): do windows get moved to the scratch layer as soon as I drag them by mouse?
* might be related to Scratch layer also: I see weird behavior with Chrome windows: for example I have browser windows on each monitor and if I move mouse from one monitor to the other, Chrome windows "hop behind" other windows. Very annoying if I have to cut and paste stuff from one browser window to another for example.

I still have to figure out how to exactly reproduce this.

I turned off other extensions related to top bar or similar elements, just to debug this.

Thanks for any help, be patient with a newbie ... and thanks for PaperWM, works great on my laptops!

jtaala:
Thanks @stefangweichinger.

Not sure I'm understanding what you're describing.  Can you please take a screen capture/video (even if on your phone etc.). and post it here?

Re Scratch and dragging: it depends when you release the dragged window - if you're by a tiling boundary, there will be visual indicators where you can drop the window (and it will be tiled):

https://github.com/paperwm/PaperWM/assets/30424662/910b77e3-c4b5-4412-8cb1-64dd44e3851f

However, if you drop NOT in a tiling boundary the window will be added to the Scratch layer:

https://github.com/paperwm/PaperWM/assets/30424662/74a2de57-2dcc-4036-963c-8d2f64360890

stefangweichinger:
@jtaala thanks for the infos and the videos, understood. I will try to come up with a video asap.

stefangweichinger:
@jtaala question: are tiled windows and "scratch windows" somehow displayed differently? Could I have different edges, for example? I will check later on my desktop ... it's not always clear to me if a window is tiled or a scratch window (which makes it hard to understand that "flipping" behavior I see). But maybe I can reproduce and provide the video ... and you can make sense of it.

jtaala:
> tiled windows and "scratch windows" somehow displayed differently?

It should be  very apparent - tiled windows are tiled and take the full vertical height.  Scratch windows are floating and look like normal Gnome windows (when PaperWM is not activated).  

Note: generally scratched windows ~~are~~ aren't used very much (I hardly used them at all - maybe when watching some video I might pop one out so its' above tiled windows)..

[...]

Issue #680, opened 2023-11-11
Better touchpad support


vvvvvvitor:
Currently touchpad support is rather lackluster, while there are basic gestures like scrolling with 3 fingers, others are completely missing even though they are pretty important.

Here are some that I noticed while using. (May update.)
- Scrolling while moving a window
- Tiling
- Switching workspaces

jtaala:
Please see the touchpad section on configuring gestures:

https://github.com/paperwm/PaperWM#touchpad-gestures

jtaala:
Hey @vvvvvvitor,

Check that link out.  For workspace switching, we provide 3-finger swipe down (which uses PaperWM's concept of spaces... including nicer workspace previewing).  You can also pull it down part-way and then 3-finger swipe on the part shown workspace to move through the tiling.

Of course you can also disable that and use the default gnome workspace switching gestures (and then switch tiling swiping to 4-fingers) - see https://github.com/paperwm/PaperWM#touchpad-gestures for how to do that.

See video below of using PaperWM's 3-finger swipe down to peek at other workspace windows (or click to select that window):

https://github.com/paperwm/PaperWM/assets/30424662/3aba19f5-9574-4890-b925-9b71196a76fa

jtaala:
I don't know what you mean by these two, can you elaborate?

- Scrolling while moving a window
- Tiling

I.e. what would gestures for above look like? (and what would they do?).

jtaala:
Finally, we're happy with the current gestures and configurability.  However, we'd definitely consider any PR you might provide here!

vvvvvvitor:
> I don't know what you mean by these two, can you elaborate?
> 
> * Scrolling while moving a window
> * Tiling
> 
> I.e. what would gestures for above look like? (and what would they do?).

You can't scroll the workspace while holding a window, sometimes it even bugs making the windows freeform.

[Gravação de tela de 2023-11-11 13-06-32.webm](https://github.com/paperwm/PaperWM/assets/119557151/c28bc016-2a0f-402e-bbd3-2182b2cff387)


As for tiling, it ins't a feature on default GNOME but it would be extremely useful to have the ability to tile the windows with gestures.

[Gravação de tela de 2023-11-11 13-09-57.webm](https://github.com/paperwm/PaperWM/assets/119557151/4bfba207-6935-4667-9cff-3d583fa55cb6)

vvvvvvitor:
> > I don't know what you mean by these two, can you elaborate?
> > 
[...]

PR #718, opened 2023-11-26
Improve scratch window behaviours (for mouse-dragging windows and new windows when a scratch window is selected)

jtaala:
This PR changes several widow behaviours related to scratch windows.  It seeks to make scratch windows more predictable (in how one scratches a window) and try to minimise instances of "accidentally/unexpectedly" scratching windows.

Users (especially new users) seem to struggle with some of the behaviours for scratch windows.

See some example comments here:

- https://github.com/paperwm/PaperWM/issues/680#issuecomment-1806929052
- https://github.com/paperwm/PaperWM/issues/700#issue-1997608259
- https://github.com/paperwm/PaperWM/issues/700#issuecomment-1824202077

On review, a couple of behaviours are rather unintuitive with scratch windows:

1. when a scratch window is selected, any new window opened will also be scratched.  I understand the thought here, but this is rarely what's wanted (for me anyway) and often leads to confusion.  This PR undoes this.  New windows will be tiled (regardless of whether a scratch window is currently selected).

2. Dragging a window (with mouse) and NOT dropping on a target now re-tiles that window (previously it scratched that window).  I've had two user reports in the last few weeks of this (reported as a bug).  In any case, it seems the more intuitive option of for that window to be re-tiled:

https://github.com/paperwm/PaperWM/assets/30424662/9c4a20d6-52e3-408f-9c0a-62ee0ceb6244


Lythenas:
Is it possible to somehow keep the highlight where the window will be tiled when you let it go? I think that would make it more obvious what is going to happen.

jtaala:
Sorry for all the review requests @Lythenas!

jtaala:
> Is it possible to somehow keep the highlight where the window will be tiled when you let it go? I think that would make it more obvious what is going to happen.

Hmm - I don't think so (not without a large rewrite I think).  Currently when a user let's go we're simply `unmakeScratchWindow` - so there's no guarantee it would get tiled in the last highlight area.  It's kind of like just unscratching an already scratched window (it'll jump back into some tiling spot).

jtaala:
I'll have a bit more of a play around - maybe I can keep the last highlight active (until another one is selected).

Lythenas:
Alternatively maybe making the highlight "triggers" bigger, so there is always a highlight would be better. But not sure how exactly that would interact between vertical and horizontal highlights. I rarely ever use the columns feature. So it might be a bit tricky to get it to feel like it does what you want it to do.

jtaala:
@Lythenas, give the latest a go.  We keep the last drop target active now.  Feels visually more intuitive.

Lythenas:
Yes that feels much better and is much more intuitive I think. I'm fine with merging this.

Lythenas (approved):

PR #733, opened 2023-12-10
Improve default PaperWM settings (for new users)

jtaala:
This PR changes the default setting for wallpapers in PaperWM.

Feedback received from several new users is that the coloured backgrounds (which have been default) is quite jarring / unexpected when first starting PaperWM.

Although this setting can be changed it's not immediately apparent on how to "go back to my nice wallpaper".


Lythenas:
Can you also update the readme? Although there I only found `default‑background` and not `use-default‑background`. Is that the same setting?

Thesola10 (approved):
Had a similar onboarding experience with a relative, this is absolutely a welcome change

jtaala:
> Can you also update the readme? Although there I only found `default‑background` and not `use-default‑background`. Is that the same setting?

Hey @Lythenas, no, it's a different setting.  The settings in the README.md are for settings that _aren't_ exposed in the UI.  This setting has always been exposed in the settings/prefs UI so was never in the README.md.  That hasn't changed (it's still in the prefs UI) - just the default setting has changed.

PR #736, opened 2023-12-10
Fix for touch window drag-and-drop

### LINKED ISSUE #735: Window drag-and-drop does not work when started from touchscreen
**Describe the bug**
~~When I try to move a window by its titlebar with the touch screen, the entire tiling view becomes stuck, and going back to the mouse reveals the window "stuck" to the mouse.~~

As of GNOME 46, drag-and-drop doesn't move the tiling hint and the window doesn't land where it should.

Finishing the drag and drop with the mouse then breaks the three-finger trackpad gesture and I have to use the touch scroll gesture to restore normal behavior

**To Reproduce**
1. Tap and hold the titlebar of a tiled window with a touchscreen
2. Move the mouse cursor using a mouse or touchpad

**Expected behavior**
The window follows my finger just as it would the mouse.

**Screenshots**

**Additional context**
Scratch windows do not exhibit this behavior whatsoever.

Thesola10:
This fixes #735

### DIFF
--- a/grab.js
+++ b/grab.js
@@ -85,6 +85,12 @@ export class MoveGrab {
         center && clone.set_pivot_point(0, 0);
 
         this.signals.connect(this.actor, "button-release-event", this.end.bind(this));
+        this.signals.connect(this.actor, "touch-event", (evt) => {
+            if (evt.type() == Clutter.EventType.TOUCH_END)
+                this.end();
+            else if (evt.type() == Clutter.EventType.TOUCH_UPDATE)
+                this.motion(this.actor, evt);
+        });
         this.signals.connect(this.actor, "motion-event", this.motion.bind(this));
         this.signals.connect(global.display, "window-entered-monitor",
             this.beginDnD.bind(this)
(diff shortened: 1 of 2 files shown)


Thesola10:
Right now the window dragging works with this PR, but tiling scroll still gets stuck afterwards. Not sure what causes that.

Thesola10:
Okay, this is weird. In my nested Shell, touch event coordinates match the global pointer exactly, but on my host Wayland session, it doesn't, which breaks everything.

Thesola10:
Okay, we may need to tell the stage to intercept all touch events to extract a facsimile of `global.get_pointer()`, and somehow do so right as `MoveGrab` is instantiated, even though it's an inherently async opeartion.

Most likely this means creating a broker object between `global.get_pointer()` and our stage-based approach, which is itself involved in (or passed to) the `MoveGrab` instance.

Thesola10:
Pending merge of #751 to rebase before implementing touch handlers

Thesola10:
Actually, disregard the last commit message, any one event also "unlocks" the drag-and-drop event, so this is a reasonable fix -- at least it removes the softlock situation.

jtaala (commented):
Hey @Thesola10,

Looks good.  I made a couple of small changes - mainly just a little linting/cleanup, and switched to using PaperWM's `warpPointer` util method (and improved it to made ripple animation showing optional).

Happy to approve this if you're happy with the changes.

Thesola10:
yep, looks good to me!

PR #737, opened 2023-12-13
Use <Alt>Tab shortcut for the "live-alt-tab" action

xalt7x:
By default GNOME is configured to switch applications (rather than windows). Both `<Super>Tab` and `<Alt>Tab` shortcuts are bound to the "switch-applications" action. "switch-windows" is unset by default. User can manually set `<Alt>Tab` to "switch-windows" and get Windows-like switcher. Some Linux distributions (e.g. Ubuntu) do this override by default.   

Currently PaperWM binds "live-alt-tab" (Window switcher with live previews) to `<Super>Tab` only.  In result:

- On vanilla GNOME + PaperWM `<Alt>Tab` shortcut doesn't do anything
- On Ubuntu + PaperWM `<Alt>Tab`  uses standard GNOME "switch-windows" action (which compared to the "live-alt-tab" has smaller thumbnails without live preview).

This commit adds `<Alt>Tab` binding to the "live-alt-tab" action.


Lythenas:
I'm fine with this change, but will wait for other comments.

jtaala:
Yeah, I'm okay with adding this as another default keybind.  Alt tab is more than standard for this as well.

Given we'll have two keybinds doing the same things here - is it worth just having `alt`+`tab` instead? (p.s. I have always change this to `alt`+`tab` and use `super`+`tab` for something else).

Although I guess this would impact users who haven't custom set this keybind (e.g. by default `super`+`tab` would now not do anything.

xalt7x:
> Although I guess this would impact users who haven't custom set this keybind (e.g. by default `super`+`tab` would now not do anything.

Well, on most GNOME distributions `super`+`tab` includes applications from all workspaces and raises all application windows on a current workspace with a focus on the last active one. Surprisingly this switcher works even with PaperWM.

On a related note, I've noticed that "switch-group" (switch windows of the same application) shortcuts (`alt`+`above_tab` , `super`+`above_tab`) are disabled for PaperWM. I understand that `super`+`above_tab` is overriden by "Switch to previously active workspace" PaperWM action. For `alt`+`above_tab` I'm not sure,  "switch-group" action sound useful for PaperWM users...

jtaala:
Hey @xalt7x,

PaperWM disables gnome keybinds that clash with it's current keybinds (if you disable PaperWM it will actually restore any keybinds that it disabled).

Anyways, we don't specifically disable `switch-group` - it's just because it's keybind clashes I would think.  I actually don't know what `switch-group` does... that keybind doesn't exist on my distro).  

You can change the `switch-group` keybind (or PaperWM's keybind for it?).

xalt7x:
@jtaala 

> I actually don't know what `switch-group` does...  

It switches windows of currently selected application. Please see the screenshot below

image

> that keybind doesn't exist on my distro).

I believe it's default GNOME keybind.  

When PaperWM is disabled or `/org/gnome/shell/extensions/paperwm/keybindings/previous-workspace` is unset
 `dconf read /org/gnome/desktop/wm/keybindings/switch-group` returns:
> ['\<Super\>Above_Tab', '\<Alt\>Above_Tab']

P.S. PaperWM doesn't disable "switch-group" on `alt`+`above_tab` if I exclude `super`+`above_tab` keybinding on "switch-group" prior to PaperWM activation.
`dconf write /org/gnome/desktop/wm/keybindings/switch-group " ['<Alt>Above_Tab']"`

jtaala:
> @jtaala
> 
> 
> When PaperWM is disabled or `/org/gnome/shell/extensions/paperwm/keybindings/previous-workspace` is unset `dconf read /org/gnome/desktop/wm/keybindings/switch-group` returns:
> 
> > ['<Super>Above_Tab', '<Alt>Above_Tab']

Yes, that sounds  right - assuming those are the default keybinds.

> P.S. PaperWM doesn't disable "switch-group" on `alt`+`above_tab` if I exclude `super`+`above_tab` keybinding on "switch-group" prior to PaperWM activation. `dconf write /org/gnome/desktop/wm/keybindings/switch-group " ['<Alt>Above_Tab']"`

Hmm - not sure if there's a prob there - PaperWM should only disable keybinds for default gnome shortcuts if there's a clash with keybind set in PaperWM keybinds (e.g. in prefs/settings).

When you enable PaperWM it checks for any clashes, and saves the original gnome keybind for restoring when PaperWM is disabled.

jtaala (approved):

Issue #389, opened 2021-08-11
current windows lost focus when mouse move across monitor


Checkmate95:
when mouse move to another monitor, current windows from previous monitor lost focus to window on current monitor but it work normally for window in same monitor is there any solution? I tried to disabled "Make topbar follow monitor focus" but it doesn't help

olejorgenb:
No, I don't think we support that atm. (ie.: the monitor which contain the mouse pointer will always contain the focused window) (correct me if I'm wrong hedin)

jtaala:
I'm not sure I'm understanding this one right - you want to have two windows focused (selected?) at the same time?

Need some more information here.

Planning on closing this in a month or so unless it's still an issue or you can provide more information.

Lythenas:
No I think the OP just want to turn off the "auto follow focus" (i.e. when you move the mouse to the other monitor, the focus changes to the last active window on that monitor).

I think it is feasible to add a setting for that.

jtaala:
> No I think the OP just want to turn off the "auto follow focus" (i.e. when you move the mouse to the other monitor, the focus changes to the last active window on that monitor).

ah, that makes sense.  Yes, probably should have a setting for this.

tflori:
I also would like to disable that. The reason are dialogs that you move to another monitor are pushed in the background what makes it complicated to go back to them:

1. open a window on your second monitor
2. open a dialog from an app on your first screen (for example settings of inkscape) 
3. move the dialog on your second screen

Immediately the window loses its focus and the dialog is pushed in the behind the active window on your second screen.

Lythenas:
I think adding an option to not have "focus follow mouse" would be fine.

But I also noticed this behavior. Unfortunately in Gnome dialogs are tied to the "parent" window. So if you move a dialog to another monitor than its parent, it gets very hard to control it. Normally in Gnome you can't even interact with the parent window and the dialog is fixed to the center of the parent. But we turn off that behavior because I think it causes some problems for PaperWM.

I don't know if or how we can fix that. It might get pretty complicated because we have to work against some Gnome behavior.

TLouf:
Just wanted to chip in with saying that as a user who primarily navigates between windows and workspaces with the keyboard, adding this option would be extremely valuable to me, regardless of the dialogs issue.

stefangweichinger:
new user here, with 2 monitors (with different sizes, if that matters somehow): I *think* I also see this issue on my system.
[...]

Issue #655, opened 2023-10-28
Cannot perform touch events on non-focused monitor


Thesola10:
**Describe the bug**
GNOME is capable of handling multiple monitors, each with their own touch surface. When PaperWM is active, touch screen events are discarded for all apps on displays other than the currently focused one, and I need to move the mouse to the other display to use the touch screen on it.

**To Reproduce**
1. Connect two touch capable displays
2. Open windows on both displays with PaperWM active
3. Move the mouse to one display
4. Try to touch the other display
5. Nothing happens

**Expected behavior**
The other window reacts properly to touch events

**System information:**

```
Distribution: Arch Linux
GNOME Shell 45.0
Display server: Wayland
PaperWM branch/tag: develop
PaperWM commit: c0ce6ab65bbf72bfc59579eb24a119060d68b581
Enabled extensions:
- drive-menu@gnome-shell-extensions.gcampax.github.com
- trayIconsReloaded@selfmade.pl
- arcmenu@arcmenu.com
- nightthemeswitcher@romainvigier.fr
- paperwm@paperwm.github.com
- dash-to-panel@jderose9.github.com
- quick-settings-tweaks@qwreey
- expandable-notifications@kaan.g.inam.org
- quick-settings-avatar@d-go
- caffeine@patapon.info
- reboottouefi@ubaygd.com
- guillotine@fopdoodle.net
- quicksettings-audio-devices-renamer@marcinjahn.com
- gsconnect@andyholmes.github.io
- vertical-workspaces@G-dH.github.com
- mediacontrols@cliffniff.github.com
```

**Additional context**
Native GNOME without PaperWM lets touch events through to unfocused monitors.

Thesola10:
This might also occur with a single touch monitor and a non-touch monitor, if the mouse is in the non-touch monitor.

jtaala:
This is likely because of the StackOverlay actor that gets enabled on monitors where the mouse isn't.

See this thread where I talk a bit about this https://github.com/peterfajdiga/karousel/issues/3

It's basically how we do a "per monitor workspace".

jtaala:
Uncomment these lines to see the overlay (that's most likely blocking your touch events):

https://github.com/paperwm/PaperWM/blob/release/stackoverlay.js#L282-L283

Thesola10:
So basically, hooking `StackOverlay`'s `TOUCH_BEGIN` to "switch monitors as if the mouse entered it" could work?

[...]

PR #751, opened 2024-01-14
Replace ClickOverlay `enterMonitor` reactive actor approach with much simpler implementation

jtaala:
@Lythenas, @Thesola10, I've been thinking about this one for a while:

PaperWM uses a very curious approach to implement a "one space per monitor" paradigm (see [my response](https://github.com/peterfajdiga/karousel/issues/3#issuecomment-1705214515) to a question of how PaperWM does this).

Unfortunately this approach is quite complex and exhibits many side-effects that have required many workarounds (which often causes further issues).

This PR replaces this approach with a simpler `PointerWatcher` implementation (a Gnome supported implementation to monitor mouse pointer movements).  Moving to this approach provides the following advantages:

- switch to a Gnome supported feature to detect when mouse pointer moves to another monitor (instead of an unsupported, custom, error-prone implementation);
- significantly simplifies the codebase for PaperWM's multimonitor support;
- replaces around ~164 lines of (now) unneeded code with ~31 lines;
- removes a complex part of multimonitor support and would allow us to add simple behaviours (based on user preference) like https://github.com/paperwm/PaperWM/issues/389;

_Note: we actually already use the `PointerWatcher` approach for multi-monitor drag/drop support - so this PR is essentially removing the other approach and adding a "activate space" method to the `PointerWatcher`._

@Lythenas and @Thesola10 - could you give this PR branch a test?  I'll set it as a draft atm given it's quite a large change and I'm sure there might be some side-effects from this.

@Thesola10 - I note that this may impact some of your touch signals, although, we no longer have the ClickOverlay element blocking actions so it may turn out better (just a hypothesis).

### DIFF
--- a/stackoverlay.js
+++ b/stackoverlay.js
@@ -69,9 +69,31 @@ export function multimonitorDragDropSupport() {
 export function enableMultimonitorDragDropSupport() {
     pointerWatch = PointerWatcher.getPointerWatcher().addWatch(100,
         () => {
-            Tiling.spaces?.clickOverlays?.forEach(c => {
-                c.monitorActiveCheck();
-            });
+            // check if in the midst of a window resize action
+            if (Tiling.inGrab && Tiling.inGrab instanceof Grab.ResizeGrab) {
+                const window = global.display?.focus_window;
+                if (window) {
+                    Scratch.makeScratch(window);
+                }
+            }
+
+            /**
+             * stop navigation before activating workspace. Avoids an issue
+             * in multimonitors where workspaces can get snapped to another monitor.
+             */
+            Navigator.finishDispatching();
+            Navigator.finishNavigation(true);
+
+            const monitor = Utils.monitorAtCurrentPoint();
+            const space = Tiling.spaces.monitors.get(monitor);
+
+            // if space is already active, do nothing
+            if (Tiling.spaces.isActiveSpace(space)) {
+                return;
+            }
+
+            const selected = space?.selectedWindow;
+            space?.activateWithFocus(selected, false, false);
         });
     console.debug('paperwm multimonitor drag/drop support is ENABLED');
 }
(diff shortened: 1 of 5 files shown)


Thesola10:
okay, the touch workspace switching is working a lot better. No issues to report.

Does not seem to impact #735 whatsoever.

stackoverlay.js
```diff
@@ -69,9 +70,34 @@ export function multimonitorDragDropSupport() {
 export function enableMultimonitorDragDropSupport() {
     pointerWatch = PointerWatcher.getPointerWatcher().addWatch(100,
```
Lythenas, line 71: Is this 100ms delay noticeable?

I found this in gnome shell where they set it to `1000 / 60` i.e. 60 times per second: https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/js/ui/magnifier.js#L229

Lythenas (commented):
The code looks good. To be honest it seems so simple compared to the click overlay...

I will use this for the next couple of days to see if I notice any issues.

stackoverlay.js
jtaala: Given we're only using this for ~~drag/drop across monitors, or~~ detecting when mouse moves to a new monitor (to activate it), 60 times per second would be overkill (that makes sense if you're interacting with a interface element that you can see, like a magnifying/zoom glass).

I chose to check mouse pointer position every 100ms to ease computational burden.  Note, PointerWatcher is quite clever and only fires if mouse moves (and has some approaches to minimize the number of times it fires depending on movement).

jtaala: > for drag/drop across monitors

It just occurred to me that drag/drop support is no longer an issue with this approach - i.e. it was only an issue since the `enterMonitor` element was blocking it (and we've now removed the `enterMonitor` element entirely).

So this approach is only to activate the space when mouse enters a new monitor, which is even simpler.

Thesola10 (approved):

Lythenas (approved):
Seems to work fine. I didn't notice any problems.

Thesola10:
I may have failed to properly test this PR. As it turns out, this seems to introduce a regression of #655.

jtaala:
> I may have failed to properly test this PR. As it turns out, this seems to introduce a regression of #655.

Thanks @Thesola10,

My guess is that we may need to add a `touch-event` listener on `space.background`, similar to what was on the `enterMonitor` element.

I've added this on PR #755 (see commit https://github.com/paperwm/PaperWM/pull/755/commits/21efa47d0443064ccd044bcde7190a1fc1f44fa9).  

Note this may not do anything (and I don't have a touchscreen to test this on).  Can you test https://github.com/paperwm/PaperWM/pull/755/commits/21efa47d0443064ccd044bcde7190a1fc1f44fa9 and see if it responds to touch events on another monitor (it should activate that space on touch with this change).  It may help in finding out more info here (or do nothing).

PR #766, opened 2024-01-29
Add global touch coords broker

### LINKED ISSUE #764: Window touch drag-and-drop starts at pointer, not at finger
**Describe the bug**
This is a technical addendum to #736. The way touch-based window dragging works, the initial coordinates when grabbing begins is used to determine the relative pointer position to the window using the global pointer. However, just as in #735, the global pointer cannot be used as a coordinates source for touch events.

This bug exists mainly as a self reminder, though anyone is free to pick it up.

**To Reproduce**
1. Tap and hold a window titlebar with a touch screen
2. Drag the window

**Expected behavior**
Same behavior as mouse-based drag-and-drop

**Screenshots**
The window jumps relative to where the pointer was when the gesture started

**System information:**
Irrelevant -- trust me :)

Thesola10:
This closes #764 by implementing a broker function, `Utils.getPointerCoords()`, which selects from `global.get_cursor()` and a memorized set of coordinates from the last touch event on Stage.

Hopefully this approach will also correct future touchscreen misbehavior.

**Edit** 2024-03-09:

This PR has been ~~repurposed as a continuation for~~ accidentally superseded by the broker partly introduced in #791.

### DIFF
--- a/grab.js
+++ b/grab.js
@@ -63,7 +63,7 @@ export class MoveGrab {
 
         this.initialY = clone.targetY;
         Easer.removeEase(clone);
-        let [gx, gy, $] = global.get_pointer();
+        let [gx, gy, $] = Utils.getPointerCoords();
 
         let px = (gx - actor.x) / actor.width;
         let py = (gy - actor.y) / actor.height;
@@ -86,7 +86,8 @@ export class MoveGrab {
 
         this.signals.connect(this.actor, "button-release-event", this.end.bind(this));
         this.signals.connect(this.actor, "touch-event", (act, evt) => {
-            if (evt.type() === Clutter.EventType.TOUCH_END) {
+            if (evt.type() === Clutter.EventType.TOUCH_END
+                || evt.type() === Clutter.EventType.TOUCH_CANCEL) {
                 this.end();
             }
             else {
@@ -121,7 +122,7 @@ export class MoveGrab {
         let clone = metaWindow.clone;
         let space = this.initialSpace;
 
-        let [gx, gy, $] = global.get_pointer();
+        let [gx, gy, $] = Utils.getPointerCoords();
         let point = {};
         if (center) {
             point = space.cloneContainer.apply_relative_transform_to_point(
@@ -177,7 +178,7 @@ export class MoveGrab {
     }
 
     spaceMotion(space, background, event) {
-        let [gx, gy, $] = global.get_pointer();
+        let [gx, gy, $] = Utils.getPointerCoords();
         let [sx, sy] = space.globalToScroll(gx, gy, { useTarget: true });
         this.selectDndZone(space, sx, sy);
     }
@@ -308,12 +309,7 @@ export class MoveGrab {
     motion(actor, event) {
         let metaWindow = this.window;
         // let [gx, gy] = event.get_coords();
-        let [gx, gy, $] = global.get_pointer();
-        if (event.type() === Clutter.EventType.TOUCH_UPDATE) {
-            [gx, gy] = event.get_coords();
-            // We update global pointer to match touch event
-            Utils.warpPointer(gx, gy, false);
-        }
+        let [gx, gy, $] = Utils.getPointerCoords();
         let [dx, dy] = this.pointerOffset;
         let clone = metaWindow.clone;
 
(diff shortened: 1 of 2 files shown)


Thesola10:
Currently does not work as coordinates do not seem to update. The way I understand it, the Stage is somehow not the global vantage point I thought it would be events-wise. What I need is an Actor that can intercept every single touch-event in the entirety of GNOME Shell...

grab.js
```diff
@@ -498,7 +494,7 @@ export class MoveGrab {
         Utils.later_add(Meta.LaterType.IDLE, () => {
             if (!global.display.end_grab_op && this.wasTiled) {
                 // move to current cursor position
-                let [x, y, _mods] = global.get_pointer();
+                let [x, y, _mods] = Utils.getPointerCoords();
```
Thesola10, line 497: Likely unwarranted -- if we want touch coords to remain separate from a UX standpoint, this just won't cut it. Then again, I'm not even sure how that may relate to drag-and-drop softlock, so I'll wait until I get my code to work before I undo this change.

Thesola10:
Another proposal, which fixes the immediate issue of centering touch window movement, is to discard changes to `grab.js` in this PR, and instead apply this patch:

```diff
diff --git a/grab.js b/grab.js
index 5aee3ec..9956e81 100644
--- a/grab.js
+++ b/grab.js
@@ -63,7 +63,7 @@ export class MoveGrab {
 
         this.initialY = clone.targetY;
         Easer.removeEase(clone);
-        let [gx, gy, $] = global.get_pointer();
+        let [gx, gy, $] = Utils.getPointerCoords();
 
         let px = (gx - actor.x) / actor.width;
         let py = (gy - actor.y) / actor.height;
```

However this is much less elegant and may cause confusion as to the role of `Utils.getPointerCoords()`, as well as keeping the fugly warp hack from #736.

Thesola10:
Might have to use [`Clutter.GestureAction`](https://gjs-docs.gnome.org/clutter13~13-gestureaction/) but that involves building an entire class just for that. Would've loved to use `SwipeTracker.TouchSwipeGesture`, but that's private.

Thesola10:
Pushed #791 as an alternate version of this fix, implementing [this solution](https://github.com/paperwm/PaperWM/pull/766#issuecomment-1915727311).

utils.js
```diff
@@ -16,15 +16,37 @@ const Display = global.display;
...
+        case Clutter.EventType.TOUCH_BEGIN:
+        case Clutter.EventType.TOUCH_UPDATE:
+            inTouch = true;
+            break;
+        case Clutter.EventType.TOUCH_END:
+        case Clutter.EventType.TOUCH_CANCEL:
+            inTouch = false;
+            break;
+        }
+        touchCoords = event.get_coords();
```
jtaala, line 43: This will be called for other events (not just touch events) which is likely not the intent here.  Add a `default` case that simply calls `return Clutter.EVENT_PROPAGATE;`, e.g.

```
    touchSignal = global.stage.connect("captured-event", (actor, event) => {
        switch (event.type()) {
        case Clutter.EventType.TOUCH_BEGIN:
        case Clutter.EventType.TOUCH_UPDATE:
            inTouch = true;
            break;
        case Clutter.EventType.TOUCH_END:
        case Clutter.EventType.TOUCH_CANCEL:
            inTouch = false;
            break;
        default:
            return Clutter.EVENT_PROPAGATE;
        }

        // was one of our touch events
        touchCoords = event.get_coords();
        return Clutter.EVENT_PROPAGATE;
    });
```

The `captured-event` is rather heavy, and captures all events (clicks, swipes, activations, etc.) so it's important that it's as efficient as possible.

Thesola10:
I wonder if it's because `CLUTTER_EVENT_PROPAGATE` causes Clutter to omit followups? Behavior is very inconsistent, it works on the panel so why not the root Stage?

Thesola10:
huh. this is annoying
I had intended to turn this PR into a `develop` + revert `fbca4` for bookkeeping and stuff, but since i accidentally also pushed the merge commit from #791 this PR is marked as merged as well. 

headdesk

jtaala:
> huh. this is annoying I had intended to turn this PR into a `develop` + revert `fbca4` for bookkeeping and stuff, but since i accidentally also pushed the merge commit from #791 this PR is marked as merged as well.
> 

I noticed that - do you need me to revert anything?

Thesola10:
nah, the branch is still there in the correct state so you can delete it if you want, as long as the comment and commit message are there it's fine

PR #770, opened 2024-01-31
Fixes 769 - guard against recursive call on position changed

### LINKED ISSUE #769: Infinte Recursion on the `position-changed` window signal
**Describe the bug**

The `position-changed` triggers another `position-changed` signal. If for some unknown reason the target window resists the computed x,y coordinates, the function causes an infinite recursion.

```typescript
        this.signals.connect(metaWindow, 'position-changed', w => {
            if (inGrab)
                return;
            let f = w.get_frame_rect();
            let clone = w.clone;
            let x = this.visibleX(w);
            let y = this.monitor.y + clone.targetY;
            x = Math.min(this.width - stack_margin, Math.max(stack_margin - f.width, x));
            x += this.monitor.x;
            if (f.x !== x || f.y !== y) {
                try {
                    w.move_frame(true, x, y);
                }
                catch (ex) {

                }
            }
        });
```

```
jan 31 15:43:10 thnkpd gnome-shell[40060]: JS ERROR: too much recursion
                                           addWindow/<@file:///home/kassick/.local/share/gnome-shell/extensions/paperwm@paperwm.github.com/tiling.js:857:23
                                           addWindow/<@file:///home/kassick/.local/share/gnome-shell/extensions/paperwm@paperwm.github.com/tiling.js:857:23
                                           addWindow/<@file:///home/kassick/.local/share/gnome-shell/extensions/paperwm@paperwm.github.com/tiling.js:857:23
                                           addWindow/<@file:///home/kassick/.local/share/gnome-shell/extensions/paperwm@paperwm.github.com/tiling.js:857:23
                                           addWindow/<@file:///home/kassick/.local/share/gnome-shell/extensions/paperwm@paperwm.github.com/tiling.js:857:23
                                           addWindow/<@file:///home/kassick/.local/share/gnome-shell/extensions/paperwm@paperwm.github.com/tiling.js:857:23
```

**To Reproduce**

I don't know what triggered this situation -- I was in slack (the desktop app) and had just clicked a thread to open it in the side view, not as a new window. Maybe some popover hint implemented as a window or something totally unrelated to slack.

**System information:**

```
Please include this information in your bug report on GitHub!
Distribution: Fedora Linux
GNOME Shell 45.3
Display server: Wayland
PaperWM branch/tag: develop
PaperWM commit: 336ac5058a7f49b36c38023bd03437bb3934eebe
Enabled extensions:
- paperwm@paperwm.github.com
- custom-hot-corners-extended@G-dH.github.com
- dash-to-dock@micxgx.gmail.com
- switcher@landau.fi
- bluetooth-quick-connect@bjarosze.gmail.com
- caffeine@patapon.info
```

**Additional context**

jtaala:
Resovles #769.

See #769.  Adds a guard on recursively  calling `position-changed`.  This purportedly can happen on windows that resist / do not respect `move_frame`, which causes a recursion and can lead to session freeze.

### DIFF
--- a/tiling.js
+++ b/tiling.js
@@ -846,6 +846,15 @@ export class Space extends Array {
         this.signals.connect(metaWindow, 'position-changed', w => {
             if (inGrab)
                 return;
+
+            // guard against recursively calling this method
+            // see https://github.com/paperwm/PaperWM/issues/769
+            if (w.pos_change_time &&
+                w.pos_change_time === global.get_current_time()) {
+                return;
+            }
+            delete w.pos_change_time;
+
             let f = w.get_frame_rect();
             let clone = w.clone;
             let x = this.visibleX(w);
@@ -854,6 +863,7 @@ export class Space extends Array {
             x += this.monitor.x;
             if (f.x !== x || f.y !== y) {
                 try {
+                    w.pos_change_time = global.get_current_time();
                     w.move_frame(true, x, y);
                 }
                 catch (ex) {
--- a/tiling.js
+++ b/tiling.js
@@ -847,6 +847,20 @@ export class Space extends Array {
             if (inGrab)
                 return;
 
+            let f = w.get_frame_rect();
+            let clone = w.clone;
+            let x = this.visibleX(w);
+            let y = this.monitor.y + clone.targetY;
+            x = Math.min(this.width - stack_margin, Math.max(stack_margin - f.width, x));
+            x += this.monitor.x;
+
+            // check if mismatch tracking needed, otherwise leave
+            if (f.x === x && f.y === y) {
+                // delete any mismatch counter (e.g. from previous attempt)
+                delete w.pos_mismatch_count;
+                return;
+            }
+
             // guard against recursively calling this method
             // see https://github.com/paperwm/PaperWM/issues/769
             if (w.pos_mismatch_count &&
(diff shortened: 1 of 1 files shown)


Lythenas:
From my point of view this looks good. Although I never ran into the recursion case.

Maybe @kassick can run it for a couple of days and check if he sees the warning in the logs. Otherwise I can approve in a couple of days.

jtaala:
Thanks @Lythenas - note I've changed the check approach here, since I wasn't happy with basing the check on a clutter timestamp (the last event time).

The new approach should be safer (checks for position mismatch and tracks this in subsequent checks) - should be more robust for windows that don't change frame position.

jtaala:
@kassick - can you checkout and test this PR when you can.  I tried to reproduce the issue with slack but haven't succeeded thus far.  These changes will stop recursive calls though.

kassick:
I've checked out the changes and will run with it through next week to validate.

>  I tried to reproduce the issue with slack but haven't succeeded 

I've been using gnome-45 for a couple of months now; paperwm for the past year. The issue happen only once...  it appears to be a once-in-a-blue-moon kind of bug :/

Since the session froze, it's not clear if gnome even managed to paint the offending window -- maybe slack was just a window that was visible mid-paint when the recursion started, caused by something else...

I'll keep an eye in the logs to see if the warning appears.

Thanks for the fix!

jtaala:
well... you won't see the any warnings in the logs unless you enable `warnings` via environment variables for your gnome shell session:

https://gjs.guide/extensions/development/debugging.html#environment-variables

jtaala:
We could change it to `console.log` while you're testing if you want?  Should make it a bit easier to see it in normal gnome-shell log.

But tbh, given it's not known how to reproduce this issue, and a guard against continual recursive calls on this signal won't hurt - I'm happy to merge this in whenever.

tiling.js
```diff
@@ -839,26 +839,49 @@ export class Space extends Array {
...
+            if (f.x === x && f.y === y) {
+                // delete any mismatch counter (e.g. from previous attempt)
+                delete w.pos_mismatch_count;
+                return;
+            }
+
+            // guard against recursively calling this method
+            // see https://github.com/paperwm/PaperWM/issues/769
+            if (w.pos_mismatch_count &&
+                w.pos_mismatch_count > 1) {
```
Lythenas, line 867: Not sure if it matters, but `> 1` here means it aborts on the fourth recursive call. First call sets to 0, second call increments to 1, third call increments to 2, on fourth call `count > 1`.

Lythenas:
Yes, I think merging this is fine since it's hard to reproduce. But we can maybe leave the issue open for a while.

tiling.js
jtaala: Yeah, I wanted to give it a few chances to call `move_frame`- so this means that currently we'd give it  a total three chances for `move_frame` to succeed (note: I wouldn't count the first pass through as recursive as it's the original call - so original plus two chances - then abort).

Lythenas (approved):

PR #776, opened 2024-02-08
add new keybinding: activate-under-mouse

timbertson:
Started using PaperWM recently, and I'm very impressed, amazing work!

One thing I had to do was to turn off focus-follows-mouse, as it caused windows to flutter around with every wayward motion. I really don't like the click-through behaviour where to activate a window (and nothing else) by clicking you have to find a hit area which doesn't do anything.

So I added a `focus-under-mouse` keybinding - pretty straightforward, it just activates whatever window is under the mouse. It's probably reasonable to have this unbound by default, I made the modification as I'd like to use it but I'll buy that it's not a super common request.


jtaala:
> by clicking you have to find a hit area which doesn't do anything.

Yes, I first thought "why not just click?" - but this is a good point.

> It's probably reasonable to have this unbound by default

I think so, given it's a valid (but not a common?) use I'd think.

jtaala:
I'd accept this - I'll move the function to `Utils` (where other mouse-position methods are) but looks good.  Thanks for the contribution!

jtaala:
Made a few slight changes (mainly unbound keybind by default, fixed so works with multiple monitors, renamed main method, linting, and a little simplification of main method to fit in with other PaperWM methods).

Also, see https://github.com/paperwm/PaperWM/discussions/723#discussioncomment-8290949

Given your contribution, if you have a donation link to accept donations (e.g. buymeacoffee.com etc) let me know and I'll add for the next release.

jtaala (approved):
Some changes implemented:

- unbound keybind by default;
- space pointer position translation (so works on all monitors);
- space activation has already occurred, so activateWithFocus is overkill - ensuring viewport should be sufficient.

timbertson:
> if you have a donation link to accept donations (e.g. buymeacoffee.com etc) let me know and I'll add for the next release

Nah all good, thanks for such a quick PR response 😀

PR #786, opened 2024-02-19
tiling: cycle through windows when switching beyond the first or last windows

davvid:
The "switch-next" and "switch-prev" actions currently do nothing once they reach the first or last or windows. Make these actions more useful by making them cycle around instead.

Cycle to the first window when using "switch-next" on the last window.

Cycle to the last window when using "switch-prev" on the first window.


davvid:
Resubmitted against the develop branch: https://github.com/paperwm/PaperWM/pull/787

Issue #278, opened 2020-03-26
Wrap around the window switcher?


indigoviolet:
Is it possible to wrap around when the end of the window sequence is reached in either direction?

Thank you for this extension, it's great!

olejorgenb:
Glad you like it :) Input is always appreciated.

We have decided against wrap-around.

- The wrap-around is (subjective) disturbing when happening by accident (a perfect animation might help)
- Holding down left/right wouldn't work as a shorthand[1] for first/last if wrap-around was enabled.
- Nailing the wrap-around animation is not trivial (the wrap-around should be noticeable without being too disturbing) The default animation didn't work very well for wrap-around IIRC.

If you want to try the default animation experience apply this patch and restart gnome-shell:

```diff
diff --git a/tiling.js b/tiling.js
index a0c2542..5f380de 100644
--- a/tiling.js
+++ b/tiling.js
@@ -730,11 +730,11 @@ class Space extends Array {
         let row = space[index].indexOf(space.selectedWindow);
         switch (direction) {
         case Meta.MotionDirection.RIGHT:
-            index++;
+            index = (index+1) % space.length;
             row = -1;
             break;;
         case Meta.MotionDirection.LEFT:
-            index--;
+            index = (index-1 + space.length) % space.length;
             row = -1;
         }
         if (index < 0 || index >= space.length)
```

(only works for super-left/right, not super-,/.)

[1] Best enjoyed with reduced key-repeat delay though.

indigoviolet:
I appreciate the quick response, the explanation and the patch. Cheers!

jtaala:
Closing as question was answered.

PR #787, opened 2024-02-19
actions: add "loop" variants of "switch-{next,previous,left,right,up,down}"

davvid:
The "switch-next",  "switch-prev" and related actions currently do
nothing once they reach the last window in the action's direction.
Add "loop" variants of these actions that loop around when the last
window is reached.
    
The new actions are "switch-previous-loop", "switch-next-loop",
"switch-left-loop", "switch-right-loop", "switch-up-loop" and
"switch-down-loop".


jtaala:
Thanks @davvid!

This would need to be an option though (e.g. in preferences) to enable/disable this behaviour.

It doesn't "feel" right to me, although maybe it's just because I prefer, and am used to the original behaviour.  Others might like this though, so would consider accepting this if it's made a user option (with default behaviour as it is currently).

Cheers,

Jay.

jtaala:
Reach out if you need a hand in adding this to preferences!

jtaala:
Just adding this here for context as to why wrap-around was not implemented previously: https://github.com/paperwm/PaperWM/issues/278

davvid:
Alrighty, I added a preference for the loop-around so that it's opt-in. Thanks for the context.

I updated everything except for `schemas/gschemas.compiled`. Do you want me to run `make -C schemas` to regenerate it and commit that as well? I did regenerate the compiled schemas to get things working (on a separate `cycle-windows/gnome-44` branch because I'm currently on gnome 44.8) but I didn't commit it since it's a binary file and it seemed worth holding off on that in case there's review notes that might affect the final schema.

Thanks for being open to adding this feature. cheers!

Lythenas:
This works fine as is, so from my side would be fine to merge (after generating the schema).

smichel17:
Would it make sense to add this as a different action (no default keybinding) instead of a setting?

On February 21, 2024 5:21:52 PM UTC, Matthias Seiffert ***@***.***> wrote:
>This works fine as is, so from my side would be fine to merge (after generating the schema).
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/paperwm/PaperWM/pull/787#issuecomment-1957378969
>You are receiving this because you are subscribed to this thread.
>
>Message ID: ***@***.***>
-- 
To respect your time, I try to write short, functional emails.

jtaala:
> This works fine as is, so from my side would be fine to merge (after generating the schema).

I thought I was doing something wrong here as I couldn't get this to work.  Then I realised this works for `switch-next`, `switch-previous` but NOT for the `switch-left`, `switch-right` (which I use - also realised I don't even use switch next/previous).

From the description I would expect this to also work for `switch-left`, `switch-right` - thoughts on implementing this for `switch-left`, `switch-right` as well (if the option is enabled)?

I guess it comes down to if this behaviour (if enabled) only makes sense for `switch-next`/`switch-previous` and not the `switch` left/right keybinds.

jtaala:
> > This works fine as is, so from my side would be fine to merge (after generating the schema).
> 
> I thought I was doing something wrong here as I couldn't get this to work. Then I realised this works for `switch-next`, `switch-previous` but NOT for the `switch-left`, `switch-right` (which I use - also realised I don't even use switch next/previous).
> 
> From the description I would expect this to also work for `switch-left`, `switch-right` - thoughts on implementing this for `switch-left`, `switch-right` as well?

This is another option, and would have been much easier to implement (sorry @davvid!).  I guess keybinds would allow uses to assign keybinds for both use cases (or just one by replacing the defaults if you always wanted wrap around).

See comment https://github.com/paperwm/PaperWM/pull/787#issuecomment-1957842112 - keybinds would make more sense here (keybinds for `switch-next-loop`, `switch-previous-loop` AND  `switch-left-loop`, `switch-right-loop`) since this would allow users to choose their preferred use case (maybe they only want loop for left/right and not next/previous..., or vice-versa,... or both).  I.e. this would be more customisable than a single option for both (and less confusing that adding another option for left/right).

davvid:
Good stuff, yeah these do feel like distinct actions. I'll take a stab at that when I get a chance later this or next week. Another nice thing is we won't need to change the schemas or add more options to the UI.

davvid:
I've split these out to separate actions and regenerated the schemas. Let me know what you think.

jtaala:
Thanks @davvid - just reordered and simplified keybind descriptions.

Planning on merging this one into develop and doing a release for this soon.

jtaala (approved):

reinier:
Has this ever been merged into a release? I want to have the option to loop the selection of active windows and have not found a way to do it. This thread seems promising.

Thesola10:
It should be, develop gets periodically merged into release. Maybe a more recent change broke it? Do open a ticket if that's the case.

reinier:
I see it now, the windows wrap around feature. Is there also a thing for this on the monitor and workspace side of things?

PR #789, opened 2024-02-20
Added `live-alt-tab-scratch` and `live-alt-tab-scratch-backward` keybinds (alt-tab just for scratch windows)

### LINKED ISSUE #783: How to easily scroll through all windows in scratch layer
When I send multiple windows to scratch, I'd like to be able to stack them in a i3-like tabbed-stack and scroll through them using Next/Prev workspace shortcuts. 

I do this in i3 by sending an entire stack to the scratch layer. I can cycle through all scratch windows easily by just navigating left/right through the stack. New browser windows created in that stack in i3 continue to remain in that stack, allowing me to create new browser windows in scratch without having to send them to scratch.

My current workaround in PaperVM is to use shortcuts for 'switch to previously active window' and 'switch to previous active window (backwards)' to cycle through the scratch layer windows. It works fairly well but if I overshoot, it mixes the current active window (in non-scratch layer) into the sequence.

Can we have a new shortcut to cycle through only the windows in the scratch layer. This would be like having a window stack in scratch layer.

Detect that if scratch layer is showing, then 'switch to previously active window' and 'switch to previously active window (backwards)' only cycle through the scratch layer windows (eg apply a filter to the list of windows if scratch layer is showing).

Thanks for this extension. I just discovered it and it's making Gnome so much easier to use for my workflow now.

jtaala:
Addresses #783.

This adds keybinds ~~(with no defaults added)~~ for selecting next/previous windows on scratch layer only.

Intended to address the issues mentioned in #783, and make it easier to work with multiple scratch windows.


Lythenas (approved):
Works fine.

PR #791, opened 2024-03-05
Add global coordinates broker but keep global cursor transform

### LINKED ISSUE #764: Window touch drag-and-drop starts at pointer, not at finger

Thesola10:
This PR functionally fixes #764 but is non-ideal from a code quality standpoint. The cleaner version and discussion around it is and should remain in #766, this PR is a hotfix.

### DIFF
--- a/utils.js
+++ b/utils.js
@@ -191,10 +191,11 @@ export function isInRect(x, y, r) {
  * Retrieves global pointer coordinates taking into account touch screen events.
  */
 export function getPointerCoords() {
-    if (inTouch)
+    if (inTouch) {
         return touchCoords;
-    else
+    } else {
         return global.get_pointer();
+    }
 }
 
 /**
--- a/utils.js
+++ b/utils.js
@@ -17,15 +17,15 @@ export let version = Config.PACKAGE_VERSION.split('.').map(Number);
 
 let warpRipple;
 
-let touchSignal = null;
+let signals, touchCoords;
 let inTouch = false;
-let touchCoords = null;
 
 export function enable() {
     warpRipple = new Ripples.Ripples(0.5, 0.5, 'ripple-pointer-location');
     warpRipple.addTo(Main.uiGroup);
 
-    touchSignal = global.stage.connect("captured-event", (actor, event) => {
+    signals = new Signals();
+    signals.connect(global.stage, "captured-event", (actor, event) => {
         switch (event.type()) {
         case Clutter.EventType.TOUCH_BEGIN:
         case Clutter.EventType.TOUCH_UPDATE:
@@ -50,7 +50,8 @@ export function disable() {
     warpRipple = null;
     markNewClonesSignalId = null;
 
-    global.stage.disconnect(touchSignal);
+    signals.destroy();
+    signals = null;
 }
 
 export function assert(condition, message, ...dump) {
@@ -189,6 +190,7 @@ export function isInRect(x, y, r) {
 
 /**
  * Retrieves global pointer coordinates taking into account touch screen events.
+ * May not work for continuous tracking, see #766.
  */
 export function getPointerCoords() {
     if (inTouch) {
--- a/utils.js
+++ b/utils.js
@@ -16,15 +16,37 @@ const Display = global.display;
 export let version = Config.PACKAGE_VERSION.split('.').map(Number);
 
 let warpRipple;
+
+let touchSignal = null;
+let inTouch = false;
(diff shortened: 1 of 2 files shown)


utils.js
```diff
@@ -16,15 +16,37 @@ const Display = global.display;
...
+        case Clutter.EventType.TOUCH_BEGIN:
+        case Clutter.EventType.TOUCH_UPDATE:
+            inTouch = true;
+            break;
+        case Clutter.EventType.TOUCH_END:
+        case Clutter.EventType.TOUCH_CANCEL:
+            inTouch = false;
+            break;
+        }
+        touchCoords = event.get_coords();
```
jtaala, line 43: This will be called for other events (not just touch events) which is likely not the intent here.  Add a `default` case that simply calls `return Clutter.EVENT_PROPAGATE;`, e.g.

```
    touchSignal = global.stage.connect("captured-event", (actor, event) => {
        switch (event.type()) {
        case Clutter.EventType.TOUCH_BEGIN:
        case Clutter.EventType.TOUCH_UPDATE:
            inTouch = true;
            break;
        case Clutter.EventType.TOUCH_END:
        case Clutter.EventType.TOUCH_CANCEL:
            inTouch = false;
            break;
        default:
            return Clutter.EVENT_PROPAGATE;
        }

        // was one of our touch events
        touchCoords = event.get_coords();
        return Clutter.EVENT_PROPAGATE;
    });
```

The `captured-event` is rather heavy, and captures all events (clicks, swipes, activations, etc.) so it's important that it's as efficient as possible.

jtaala:
@Thesola10 let me know if you when you want me to test/review this one.

Thesola10:
@jtaala go ahead!

jtaala (approved):
Looks good.  I switched to PaperWM's signal approach (for consistency and it's also easier to destroy/cleanup - for when/if we add more signals to utils.js.

Give  it a final test and I'll merge and do a new release.

utils.js
```diff
@@ -10,22 +10,22 @@ import * as Main from 'resource:///org/gnome/shell/ui/main.js';
 import * as Ripples from 'resource:///org/gnome/shell/ui/ripples.js';
 import * as Config from 'resource:///org/gnome/shell/misc/config.js';
 
-import { Lib } from './imports.js';
+import { Utils, Lib } from './imports.js';
```
Thesola10: wait you can do that??? just let the file circular-import itself??

Thesola10: and it works flawlessly. Wow uh, JS I guess :sweat_smile:

Thesola10:
image

just filed a commit instead

utils.js
jtaala: Oops, yeah, you can do that fine... although I didn't even realise it's utils.js itself (lol) - so don't need to import.  I fixup.

Thesola10: I think you do need to import, when I tried using the class on top it didn't work so there's still an order scope a la C/C++

Otherwise I would have absolutely used the Signals class from the start.

Or not, but it definitely didn't work when I tried... (or was it the LSP complaining for no reason? idk) I'm very confused

jtaala (approved):
Looks good, merge when ready.

jtaala:
> Looks good, merge when ready.

Ah, oops, I'll just merge this in if you're good with it?

Thesola10:
Just squashed, go ahead

utils.js
jtaala: @Thesola10, is this still an issue?  happy to revert that change if it's causing issues (e.g. use the import).  There is a order to modules being enabled that can cause issues).

Let me know.

Thesola10: @jtaala Works now, hence why I marked this convo as resolved!

i'm just not used to JS' way of doing things I guess

jtaala: Okay, I'll merge this then.  Thanks @Thesola10.

PR #815, opened 2024-04-07
Add EDGE focus mode

valpackett:
This mode is especially aimed at heavy touchpad swipe gesture users.

With the keyboard shortcuts and default focus mode, we already mostly just get one edge of a window always snapped to an edge of the screen.

With swipe gestures however it is way too easy to end up at some arbitrary scroll position, as the default snap behavior (based only on windows getting of bounds) is not enough to ensure snapped edges.

Now with the edge snapping it's a lot better:

[paperwm-edge.webm](https://github.com/paperwm/PaperWM/assets/208340/e1c32ce1-cdaa-423e-99dd-cecbf0fe30f5)


jtaala:
@valpackett thanks! - I can see the preference of some users for having this.

Only question I have is if anyone thinks this would this better be a user preference option settable in settings in the `gestures` section, e.g. a simple boolean option like `snap windows to edge`?

Other than that - I'm actually happy with this!  Although let's also add it to the settable user options to make this the default for workspaces if user's want to e.g.:

image

Given the prevalence of this option, I might add it to the `General` settings tab in user preferences.

_Note: this also works well with the `Switch between Window Focus Modes (e.g. default, center)` keybind.  Might also be worth add a `switch (backwards)...` heybind for this (although I can add this later)._

jtaala:
@valpackett, p.s. I just added this option to settings preference.

jtaala:
@valpackett - just made a few tweaks (including promoting this setting to `General` page, and action a layout when it's changed). :+1:

valpackett:
Oops, I missed the default thing being present in settings. 6am me misread the readme thinking *it* was manually-edit-gsettings-only, when in reality that was about hiding the focus icon :sweat_smile: 

> if anyone thinks this would this better be a user preference option settable in settings in the gestures section, e.g. a simple boolean option like snap windows to edge?

I personally would still like to be able to press a top panel button to go back to the free scroll mode…

tiling.js
```diff
@@ -4546,6 +4556,26 @@ export function setFocusMode(mode, space) {
...
+            if (winMidpoint <= workAreaMidpoint) {
+                space.unfocusXPosition = 0;
+            } else {
+                space.unfocusXPosition = workArea.width;
+            }
+            centerWindowHorizontally(selectedWin);
+        }
+        break;
+    default:
+        // for other modes run a `layout` call to action the mode
```
valpackett, line 4578: oh, it was that simple to not have to repeat code there

jtaala: :+1: I've pushed another one for some more cleanup (forgot to remove manual if check which the above code replaced).

jtaala:
> Oops, I missed the default thing being present in settings. 6am me misread the readme thinking _it_ was manually-edit-gsettings-only, when in reality that was about hiding the focus icon 😅
> 
> > if anyone thinks this would this better be a user preference option settable in settings in the gestures section, e.g. a simple boolean option like snap windows to edge?
> 
> I personally would still like to be able to press a top panel button to go back to the free scroll mode…

Yeah, me too.  Also, in terms of ordering the options for selection when clicking the button, I'm tempted to change the ordering to: `Default` --> `Edge` --> `Center` since it seems to make more sense (in that `Edge` is kind of similar to `Default`).  But I think users may be already used to the _toggle/click once to get to center_.

Thoughts? @Lythenas?

jtaala (approved):
Great work @valpackett!

Made a few tweaks and committed.  I'm going to approve this and will wait a day or so for any other comments.

Planning to merge/release this soon!

jtaala:
P.S. we've trying a thing atm where we put up a donation link from contributors of PRs etc. (if they want) - see https://github.com/paperwm/PaperWM/discussions/723#discussioncomment-8290949.

Feel free to message me or comment here if you would like to have a donation link added. (can also leave a comment on https://jaytaala.com - I won't make any comment there re this public btw) .

valpackett:
huh, rolling donation links between latest contributors is an interesting idea, haven't heard that before. If you want one from me: https://www.patreon.com/valpackett :)

jtaala:
P.S. I've just merged a fix into your branch which fixes #823.

PR #822, opened 2024-04-08
Improve margin settings (and add explicit setting to set the `selected` window coloured border width)

### LINKED ISSUE #817: Not sure if bug - inconsistency between bottom margin and top margin calculation
**Describe the bug**
In the paperWM settings there are 2 fields - bottom margin which adds a bottom margin, and top margin which adds a top margin.

This works fine - _however_, the setting for window gaps, which in a **horizontal stacking window manager** one would assume means just that - gaps _between_ windows, **_also_** adds half its margin to the **top** as well, but _not_ the bottom.

I think this is inconsistent and perhaps wrong, especially if someone wants to use a floating status bar (like I am) and wants a consistent gap between the top and bottom of the bar - this results in too much space between the tops of windows and the status bar.

**To Reproduce**
1. Go to settings
2. Go to top gap
3. Set to 0
4. Go to gap between windows
5. set to any large number
6. There now is a gap at the top of the window, when perhaps there should not be

**Expected behavior**
I would expect:

- Top margin to _only_ add top margin (it does not)
- Bottom to _only_ add bottom margin (it does)
- Window gap to _only_ add gap horizontally between windows (it does not)
- Horizontal gap to _only_ add gaps to the sides of a monitor

Alternatively

- Horizontal margin _only_ adds space _between_ windows
- Window gap adds margin on _all sides_ of a window
- Top margin to _only_ add top margin
- Bottom to _only_ add bottom margin

**Screenshots**
N/A

**System information:**
Not necessary

**Additional context**
I have locally made a 1.5 line code change that fixes the issue in what I believe is the original intention of the settings. It took about 5 seconds and a relogin,

Line 112 - 116 in tiling.js
```
  let setVerticalMargin = () => {
      let vMargin = gsettings.get_int('vertical-margin');
      // let gap = gsettings.get_int('window-gap');
      Settings.prefs.vertical_margin = vMargin;
  };
```

jtaala:
Fixes #817.

See https://github.com/paperwm/PaperWM/issues/817#issuecomment-2042684866 for details.

Allows independent setting of margins and gaps (without the strange behaviour described in #817).

Adds a setting to user preferences which allows setting directly the `selected` window border "thickness":

[Screencast from 2024-04-08 23-51-21.webm](https://github.com/paperwm/PaperWM/assets/30424662/2382e4b2-87fb-437b-b0f1-60edad237af7)

[Screencast from 2024-04-09 08-26-01.webm](https://github.com/paperwm/PaperWM/assets/30424662/b1039dc0-0c7f-4b0d-91e2-2f066a138044)


smichel17 (changes requested):
This approach acts weirdly if you set the border size to nonzero and all margins and gaps to 0.

I have some other thoughts I'll leave in #817 when I find the energy, but wanted to get this in before merge.

jtaala:
> This approach acts weirdly if you set the border size to nonzero and all margins and gaps to 0.
> 
> I have some other thoughts I'll leave in #817 when I find the energy, but wanted to get this in before merge.

Thanks @smichel17.  I'd say it's _predictably_ weird!  But yes, this corner case doesn't quite make sense from a visual pov.

We could set the actual implementation to not let the border size be larger than margins/gaps which should resolve this case.

jtaala:
> We could set the actual implementation to not let the border size be larger than margins/gaps which should resolve this case.

@smichel17, see implementation of this in the latest commit:

[Screencast from 2024-04-09 08-26-01.webm](https://github.com/paperwm/PaperWM/assets/30424662/b1039dc0-0c7f-4b0d-91e2-2f066a138044)

PR #823, opened 2024-04-08
FIX: restore space focusMode on unlock.

### LINKED ISSUE #819: Focus mode resets on screen lock-unlock
**Describe the bug**
Focus mode resets to DEFAULT on screen lock-unlock

**To Reproduce**
1. Change focus mode
2. Lock screen
3. Unlock screen

jtaala:
Fixes #819.

Restores `focusMode` for spaces on unlock.

### DIFF
--- a/tiling.js
+++ b/tiling.js
@@ -343,6 +343,9 @@ export class Space extends Array {
         let workspace = this.workspace;
         let prevSpace = saveState.getPrevSpaceByUUID(this.uuid);
         console.info(`restore by uuid: ${this.uuid}, prevSpace name: ${prevSpace?.name}`);
+
+        // get previous focus mode (if exists)
+        const focusMode = prevSpace?.focusMode;
         this.addAll(prevSpace);
         saveState.prevSpaces.delete(workspace);
         this._populated = true;
@@ -352,7 +355,7 @@ export class Space extends Array {
         this.setSpaceTopbarElementsVisible();
 
         // apply default focus mode
-        setFocusMode(getDefaultFocusMode(), this);
+        setFocusMode(focusMode ?? getDefaultFocusMode(), this);
 
         this.getWindows().forEach(w => {
             animateWindow(w);


tiling.js
```diff
@@ -352,7 +355,7 @@ export class Space extends Array {
         this.setSpaceTopbarElementsVisible();
 
         // apply default focus mode
-        setFocusMode(getDefaultFocusMode(), this);
+        setFocusMode(focusMode ?? getDefaultFocusMode(), this);
```
smichel17, line 358: Is the variable needed for functionality, or would this have the same effect?
```suggestion
        setFocusMode(prevSpace?.focusMode ?? getDefaultFocusMode(), this);
```
I'd find ^that slightly more readable, since I don't need to glance up at an earlier line to see where `focusMode` comes from.

tiling.js
```diff
@@ -352,7 +355,7 @@ export class Space extends Array {
         this.setSpaceTopbarElementsVisible();
 
         // apply default focus mode
```
smichel17, line 357:
```suggestion
        // restore focus mode
```

tiling.js
jtaala: Hey @smichel17,

Yes, this was done due to a concern of a race condition from the `saveState` deletion and setting the focus mode after `this.setSpaceTopbarElementsVisible();`.

Logically I wanted to maintain current/proven/tested order of restoring state, e.g. 

_get/do all our `prevState` things_ ==> _delete/finish prevState usage_ ==> _set space topbar element initial visibility_ ==> _set focus mode_.

Having said that, the concern with a race condition is probably overblown, and thinking about it, we could probably replace line 348 with ` setFocusMode(prevSpace?.focusMode ?? getDefaultFocusMode(), this);`.

Feel free to test and run it for a while and let me know if there's any issues!

_Note: restoring space / window state is quite sensitive and the order of restore has proven quite touchy in the past, so this seemed the safest approach._

tiling.js
jtaala: Have updated this comment in #822.

PR #825, opened 2024-04-13
Feature: move space to other monitors

### LINKED ISSUE #816: Second monitor doesn't have tiling/Crash
**Describe the bug**
When trying to switch a window to my second monitor, using shift+super+ctrl+arrows, it causes a hard crash. The top bar still functions, I can still open quick settings, but everything else just goes black and breaks.
If i do this just after starting gnome starts (I have around 30 seconds), it doesn't crash, but the windows on the second display are floating.

I don't know if this is a missing feature, but I also can't drag a window to another display/workspace, it could be related
**To Reproduce**
1. attempt to move a window using the paperwm keybinds
2. crash, or get a floating window on the second monitor

**Expected behavior**
The window moves to the second monitor, and joins a stack on the second monitor

**System information:**
Distribution: Gentoo
GNOME Shell 45.5
Display server: Wayland
Enabled extensions:
- paperwm@paperwm.github.com
- launch-new-instance@gnome-shell-extensions.gcampax.github.com

**Additional context**
I only have 1 workspace enabled in gnome settings, I have made sure that I disabled all the settings that weren't recommended in the README.

### LINKED ISSUE #824: Option to move a workspace to a monitor instead of swapping
I'd like an easier way to put a workspace on a given monitor.

At the moment I can swap a workspace with the one on the monitor to the left/right/etc. But generally I'm happy with where that workspace is already, otherwise I'd have moved it already. It becomes a bit of a rubik's cube puzzle to figure out which workspace to switch to on the other monitor so that when I swap the two workspaces I'm happy with the result.

I'd love new options to set a shortcuts that would simply move the current workspace to the monitor in a given direction.

jtaala:
This PR resolves #824.

It adds the following:
- keybinds (`Ctrl`+`Alt`+`left`/`right`/`up`/`down`) which moves the current workspace to a different monitor in that direction;
- if a workspace is the last on a monitor, we fallback to swapping that space with the target monitor space;
- informs user with notification if can't move / swap a space;
- FIXES #816 by creating spaces to meet a minimum (i.e. PaperWM breaks if have 2 monitors and only one space...);


Lythenas:
As far as I can tell this works fine.

Are you sure you want to use these keybindings? I don't really use it that often but e.g. in gimp ctrl+alt+arrow moves the selection. I imagine in other graphical programs this does something similar. Not sure if that is a big deal.

jtaala:
> As far as I can tell this works fine.
> 
> Are you sure you want to use these keybindings? I don't really use it that often but e.g. in gimp ctrl+alt+arrow moves the selection. I imagine in other graphical programs this does something similar. Not sure if that is a big deal.

Good pickup.  Yes, that would be confusing for sure (go to move selection in gimp and you move the workspace to another monitor...).  Any suggestions for alternate keybinds?  

I was trying to keep some relation to the swap keybinds (`alt+`super+...).

~~What about `Ctrl`+`Super`+`left`/`right`/`up`/`down`?~~ can't do them since they clash with move window to monitor keybinds...

What about 
- `Ctrl`+`Alt`+`Shift`+`left`/`right`/`up`/`down`?

Dang, running out of keybinds...

Lythenas:
> `Ctrl`+`Alt`+`Shift`+`left`/`right`/`up`/`down`?

That is probably fine.

I would also be fine with not having default keybindings, but then the feature might be hard to discover.

PR #829, opened 2024-04-18
Dynamic window opening direction

### LINKED ISSUE #727: Dynamic window opening direction
I use very often `super+return` mostly for terminal and Firefox windows (since I'm using PaperWM I don't use tabs anymore and spawn and close windows when needed).

Recently, an option to open the new window on the left was added #679. It's cool but too static for my workflow, most of the time opening to right is fine for me.
What I was about to suggest is to add a shortcut, let's say `super+shift+return` to open the new windows on the left.
But maybe this idea can be generalized a bit: 
- it could dynamically change the parameter introduce in #679 
and even solve #725
- or a set of shortcuts that temporaly modify the opening direction 

In any case, a user feedback about the "open direction" could be given like what has been done to show the user the current focus mode (aka an icon in the top bar)

Many thanks

jtaala:
Fixes #727.

This PR adds the ability to change the "direction" that new windows will be opened (e.g. to the right/left of the current window).

It also adds:
- an icon on the Gnome topbar indicating the current mode;
- setting to hide this icon;
- setting to set the modes available for cycling through (e.g. `right`, `left`, `start`, `end`);
- a keybind to cycle these modes (`Shift`+`Super`+`w`);

[Screencast from 2024-04-19 08-06-43.webm](https://github.com/paperwm/PaperWM/assets/30424662/4e4aa415-d047-44cb-b87a-d7e08493ecbd)


jtaala:
P.S. @valpackett, I've changed the icon for the `edge mode` since the original one reminded me of fullscreen mode.

Having said that, if you have strong feelings towards the original icon let me know (and not the new one) let me know!

P.S.S. have a look at https://flathub.org/apps/org.gnome.design.IconLibrary and let me know if any other ones from there look okay/better.

valpackett:
> P.S. @valpackett, I've changed the icon for the `edge mode` since the original one reminded me of fullscreen mode.

The new one looks generally better, the edges it shows are rather thin I guess but that's fine

PR #834, opened 2024-04-22
Update README.md

EyalRo:
had a typo:
`you will can then enable PaperWM`
to
`you can then enable PaperWM`


jtaala:
Thanks @EyalRo.  Yes, doco does needs a good updating.

Issue #837, opened 2024-04-24
Switch windows across monitors and workspaces with the keyboard


ankostis:
(continuation of #330)

### Problem

I'm missing a fast way (single keystroke like <kbd>Alt</kbd>+<kbd>Tab</kbd>) to cycle between the recently-used windows _across all monitors_ (first) and workspaces (good to have).   A case requiring special attention due to its frequent use is to go back quickly (no popup) to the immediately previous window with a singular keystroke.

> Note that i'm NOT talking about switching of "Applications", since gnome-shells already offers this operation with a key-binding _"Switch applications" (by default  bound to <kbd>Super</kbd>+<kbd>Tab</kbd>) that selectively includes apps from current/all workspaces or not depending of the _"Settings | Multitasking | Include apps from all workspaces/current workspace only"_.  Unfortunately this functionality is particularly cumbersome when switching between windows of the same application (eg. browser), also a very common operation.  In that case, many extra cursor keystrokes are needed to select the desired app's window.

### Alternatives i've considered

- PaperWM's _"Switch to previously active window/backwards"_ bindings include windows from current workspace/monitor only.

- PaperWM's _"Switch to the next/previous window"_ & _"Switch to the left/right/... window"_ bindings include windows from current workspace/monitor only, and that does make sense here.  But even if we wanted to retrofit them somehow to include windows from other monitors & workspaces, they would invalidate the recent-activation stack since they activate each window as they traverse them.

- In vanilla gnome-shell, i switch windows across all monitors simply by using a single workspace on all of them.
  This is impossible with PaperWM which, by design, it uses a different workspace per monitor.

- When i'm forced to use multiple workspaces, i install extensions like the _[AATWS](https://extensions.gnome.org/extension/4412/advanced-alttab-window-switcher/)_.
  Unfortunately all such extensions would fail when switching windows across monitors, I guess, due to the way PaperWM focusing works (it follows the mouse): after selecting the window to switch , eg with <kbd>Alt</kbd>+<kbd>Tab</kbd>, PaperWM refocuses back to the window in the monitor where the mouse is located (probably not the newly selected monitor). 

- The _Switcher_ extension suffers from the above focusing issue, and requires, best case, +1 keystroke (the final <kbd>Return</kbd> to activate the selected window).

[...]

PR #839, opened 2024-04-25
Fix warp pointer if focus window on another monitor

jtaala:
Fixes a related issue partly described in #837.

Note that in Gnome 45+, PaperWM `alt`+`tab` does indeed switch through all windows across spaces and monitors (by default).  However, an issue does occur that if we switch to (and focus) a window on another monitor via `alt`+`tab` etc., when we move the mouse focus will be pulled to the monitor that the mouse is on.

This PR fixes this by warping the mouse pointer to the new monitor if needed (along with a mouse pointer ripple effect).

This PR also now links the behaviour for showing `all` windows or `only windows from current space` in PaperWM to the Gnome Multitasking setting:

image

### DIFF
--- a/utils.js
+++ b/utils.js
@@ -12,7 +12,7 @@ import * as Config from 'resource:///org/gnome/shell/misc/config.js';
 
 import { Lib } from './imports.js';
 
-const Display = getGlobal().display;
+const Display = global.display;
 export let version = Config.PACKAGE_VERSION.split('.').map(Number);
 
 let warpRipple;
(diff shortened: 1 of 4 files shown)


utils.js
```diff
@@ -12,20 +12,30 @@ import * as Config from 'resource:///org/gnome/shell/misc/config.js';
...
 
+/**
+ * Convenience function to get global.  Avoids linter flagging
+ * references to global as no-undef
+ * @returns global
+ */
+export function getGlobal() {
+    // eslint-disable-next-line no-undef
+    return global;
+}
```
Lythenas: I don't think this is needed just to fix a lint warning. For eslint you should be able to fix the warning by adding `global` to the list of globals configured in `.eslintrc.yml`.

But also I haven't gotten the `.eslint.yml` to work for me. I have only ever worked with js configs for eslint.

Lythenas (commented):
Other than the eslint comment, the changes seem fine. But I haven't been able to test them yet.

utils.js
jtaala: Well that's much easier!  I got sick of individually ignoring (with eslint ignore lines) every `global` reference... so thought to do it one place.

Will update the lint config.

jtaala: For reference, I'm just using https://open-vsx.org/extension/dbaeumer/vscode-eslint within `Code` for development.

I just installed it, did a `npm install -g eslint`, closed/opened `Code` and then added the PaperWM folder to the workspace.

Issue #847, opened 2024-04-29
Select window or workspace above or below


gustavcedersjo:
I have a hard time to remember a lot of different key bindings, so I try to make them few, but useful, and with a regular pattern to more easily remember. For example, I use Super+(direction) to change focus. However, if I want to switch focus to the workspace below, a need a different shortcut for that.

I would really like to reuse the same keybinding for selecting windows and workspaces. That way, I don't have to remember which combination of modifiers I have for switching workspaces. My suggestion is to introduce new keybindings for selecting the window or the workspace above and below, respectively. If selecting upwards, and there is a window above, then that window is selected, otherwise, the workspace above is selected.

I'm on the gnome-44 branch and made a quick implementation there. I can make a pull request if anyone wants to try it.

Lythenas:
By workspace above do you mean you have a monitor above or do you mean switching to another workspace on the current monitor?

Just to mention this: Currently the `Super+Up/Down` would select a window above or below in the current column. Although I never used columns.

gustavcedersjo:
Maybe you could have a similar system for monitors as well, but I was thinking about switching to another workspace. The workspaces are aligned vertically, but monitors are not always set up like that.

Lythenas:
We do have keybindings for switching to a monitor in all directions, including up and down. Although I've never used them. So if you want to switch to the workspace that is shown on the monitor above the current one you can use that keybinding. But that does not sound like what you want.

So you basically want `Super+Down` to act like `Super+PageDown`.

I think that would be ok to add. But probably as a setting or another set of keybindings instead of changing the default.

Unfortunately it is not 100% obvious what the "correct" default behavior should be. E.g. if you have to monitors on top of each other I would expect navigating down to select the other monitor instead of switching workspaces on the current monitor. But it should be fine with a setting or separate keybindings.

gustavcedersjo:
Yes, I want `Super+Down` to act like `Super+PageDown` if there is no window below to switch to. But, of course, as a new command, that I can bind `Super+Down` to.

I think combining "Activate window below/above" and "Switch to workspace above/below" works quite well. Doing something similar in the left/right direction would be trickier, since that would require scrolling through all windows before reaching the end.

Lythenas:
Actually it is probably about the same. Because we do have columns where you can stack windows on top of each other in the workspace. Although I have never used that.

gustavcedersjo:
[...]

PR #850, opened 2024-05-05
Add commands switch-(up/down)-or-else-workspace

gustavcedersjo:
Add commands to switch to the window above (or below), but fall back to switching to the workspace above (or below) if there was no window above (or below).

Keyboard shortcuts are added as well, but without default keybindings.

Issue #847


Lythenas:
The code looks fine. I will try it out tomorrow if I get time.

Also I think the github action is broken...

schemas/org.gnome.shell.extensions.paperwm.gschema.xml
```diff
@@ -238,6 +238,14 @@
         <default><![CDATA[[]]]></default>
         <summary>Switch to the below window (no monitor boundary)</summary>
     </key>
+    <key type="as" name="switch-up-or-else-workspace">
+        <default><![CDATA[['']]]></default>
```
Lythenas: @jtaala should this be `<![CDATA[['']]]>`or `<![CDATA[[]]]>`? It looks like we used it inconsistently.

jtaala: Good pickup @Lythenas - I don't think it matters here much(?), but for consistency and intent, it should be the latter for no keybind applied.

jtaala:
Thanks @gustavcedersjo!  Hopefully will have time to look at this one tonight.

jtaala:
Hey @gustavcedersjo - nice work.  I like it.

Is everyone happy with the description of these shortcuts?:

`Switch to the above window (no workspace boundary)`
`Switch to the below window (no workspace boundary)`

it definitely fits the decription for `Switch to the above window (no monitor boundary)` etc.  Any simpler way to describe these? (let me know if anyone has any bright ideas there).

Otherwise, am approving this and will merge tomorrow evening.

Jay.

jtaala (approved):
Looks good.  Great work @gustavcedersjo.

Lythenas:
> `Switch to the above window (no workspace boundary)`
> `Switch to the below window (no workspace boundary)`

Maybe `Switch to window or workspace above/below`? Since we are switching to another workspace and not just switching a window if there is no window above/below.

gustavcedersjo:
When I first made the patch, I called it something along the lines of your suggestion, @Lythenas, but I changed it when I saw the other settings. The `switch-gobal-*` could also be called something like `Switch to window or monitor above/below`.

jtaala:
How's this?

image

jtaala:
P.S.@gustavcedersjo,  we've trying a thing atm where we put up a donation link from contributors of PRs etc. (if they want) - see https://github.com/paperwm/PaperWM/discussions/723#discussioncomment-8290949.

Feel free to message me or comment here if you would like to have a donation link added. (can also leave a comment on https://confluence.jaytaala.com/display/AJT/ - I won't make any comment there re this public btw) .

PR #856, opened 2024-05-09
Add keybindings for activating windows at other position (second, third, fourth...)

### LINKED ISSUE #853: Keybindings for window at position X
Hi all,
On the back of [this suggestion](https://github.com/paperwm/PaperWM/discussions/842) of being able to load certain apps in the same window position, I thought it might also be handy to have the ability to map window positions to the keyboard as well.
Currently Super+Home and Super+End jump to the start and end of the window list which is really handy.

I'm thinking that since I don't use my function keys (F-keys top of keyboard) for anything other than controlling media it would be handy to map F1 to position-1, F2 to position-2 and F3 to position-3 etc etc. (or what ever the user prefers).

This is something I think I heard Primeagen (ThePrimeTime on youtube) mention and I came to understand how it's important to his workflow. Having certain apps which he can always access with the exact same key combo.

If it is possible to implement "App P always loads in Position X of current workspace", then we are able to map a key to jump to that position for fast navigation, F1 will take me to Position X where I always have App P loaded, maybe that's a web browser or your text editor/IDE of choice for example.

I imagine that the above is easier to implement rather than "mapping key-combo to jump to App P where ever it is in this workspace" as you might have multiple windows open for app P.

In terms of "always loading App P on position X", any additional windows will probably need to be loaded in the next "non-reserved" spot in the workspace maybe.

Hope the idea sounds sane, and interesting to implement (wish I knew how to code in JS). :)

Many thanks.

_Originally posted by @RobLW in https://github.com/paperwm/PaperWM/discussions/849_

jtaala:
Fixes #853.

This PR adds keybind options for activating windows at different positions.  Note, by default these keybinds are unassigned.

image


Lythenas (approved):
Works fine.

jtaala:
Thanks @Lythenas.

jtaala:
Will release tonight.

PR #859, opened 2024-05-14
FIX: consume / expel functions (for vertically stacked windows) respects open window directions

jtaala:
This PR fixes / implements the respecting of the side that we `consume` and `expel` windows on when using vertically stacked windows.

See this comment: https://github.com/paperwm/PaperWM/pull/679#issuecomment-1806796391

That is, if we opening windows on the `left` (or `start`) then we also `consume` and `expel` windows to the left.

This PR also:
- adds a new keybind (`shift`+`super`+`o`) which expels the currently active window (instead of always only expelling the bottom window);

[Screencast from 2024-05-14 20-04-06.webm](https://github.com/paperwm/PaperWM/assets/30424662/3bbf4f23-6a08-4b8c-8990-38cad5717f4e)


jtaala:
Maybe `super`+`p` isn't a great default keybind (since it's used for changing monitors I believe by default in Gnome)?

Issue #858, opened 2024-05-12
Vertical layout for portrait mode


nicolandu:
I have a primary monitor in landscape mode and a second monitor in portrait mode. Consuming/expelling windows to get full-width windows stacked above one another on my secondary is a painstaking process, and new windows keep opening to the right rather than below.

I would suggest a per-monitor setting of the "ribbon direction" (top-bottom vs side-to-side), to be able to have a sideways layout on a landscape monitor and a vertical layout on a portrait monitor.

I have tried looking up ways to change this setting, such as changing the window opening direction (#829), but it seems to revolve around a horizontal layout.

For me, the lack of a per-monitor orientation setting is a showstopper, and I'll seriously have to reconsider my workflow. PaperWM seems to have a lot of potential for customization, but it doesn't seem to play nice with vertical monitors.

jtaala:
Hey @nicolandu,

Sounds like `i3` may suit your workflow better?  PaperWM definitely isn't for everyone and their preferred workflows.

In any case, I'm starting to look at related functionality (see #725, #852).  First step is functionality to "open" new windows in a vertical stack (which is not implemented atm).  Once we have that, we could then add this behaviour (vertical stacking modes) to a new dynamic "openWindow" directions.  Last step would be making dynamic opening direction "per monitor" (like we do with "focusModes").

It's on the books/plan to be added.

Cheers,

Jay.

nicolandu:
I appreciate your support, and I really like PaperWM! I'll have a look at i3, but for now, horizontal mode should do the job. Would you consider appropriate to close the issue?

jtaala:
No probs.  Keep the issue open for sure - I'm currently working on having `consume` and `expel` windows from a vertical stack honour the current "open window" direction (e.g. left/right).  I'm pretty much done that and will add a PR soon.

Planning on tackling the "open window in a vertical stack" mode next/soon.

rbutera:
> No probs. Keep the issue open for sure - I'm currently working on having `consume` and `expel` windows from a vertical stack honour the current "open window" direction (e.g. left/right). I'm pretty much done that and will add a PR soon.
> 
> Planning on tackling the "open window in a vertical stack" mode next/soon.

Hey I'm considering moving to PaperWM but I am struggling to understand something... If I change the direction that new windows will open in will it apply to all monitors or just the monitor currently in focus?

[...]

PR #862, opened 2024-05-20
Feature: add modes to create windows in vertical stack

### LINKED ISSUE #725: Shortcut to open new window vertically
My most frequently used keyboard shortcut (`super+return`) launches a new terminal window. Most of the time, the default behavior of adding the new window to the right of the currently focused window does what I want but sometimes I want to open the new terminal window _below_ the currently focused window. To do this, I have to execute four keyboard shortcuts:
1. Open terminal (`super+return`)
2. Switch to left window (`super+h` - I remapped)
3. Absorb newly created window (`super+i`)
4. Switch to absorbed window (`super+j` - remapped)

I used to use i3 and there is a way to modally switch the direction of the next split _prior_ to making the split. So, an idea is to do something like:
1. Switch split direction (`super+v`)
2. Open terminal (`super+return`)

Of course, this would mean having some sort of indication of where the next split would be created would be useful.

jtaala:
This PR improves the situation described in #858, and resolves #725;

More specifically, it adds new "create window" modes:
- `down`: create new windows in vertical stack (below current active window);
- `up`: create new windows in the vertical stack (above current active window);

image

This PR also adds new keybinds (unset by default) to directly set the position to open windows:
image


schemas/org.gnome.shell.extensions.paperwm.gschema.xml
```diff
@@ -171,6 +171,30 @@
...
+      <summary>Create new windows in vertical stack (up)</summary>
+    </key>
+    <key type="as" name="open-window-position-start">
+      <default><![CDATA[[]]]></default>
+      <summary>Create new windows at start position</summary>
+    </key>
+    <key type="as" name="open-window-position-end">
+      <default><![CDATA[[]]]></default>
+      <summary>Create new windows at end position</summary>
+    </key>
```
Thesola10, line 197: Maybe this should be an enum? idk how GNOME does it but `org.gnome.mutter.experimental-features` has the schema I mean

tiling.js
```diff
@@ -3868,15 +3874,35 @@ export function insertWindow(metaWindow, { existing }) {
...
+                stackSlurpTimeout = GLib.timeout_add(
+                    GLib.PRIORITY_DEFAULT,
+                    100,
+                    () => {
+                        slurp(active, SlurpInsertPosition.ABOVE);
+                        return false; // on return false destroys timeout
+                    });
+                break;
+            }
+        });
```
Thesola10, line 3905: Any way to make the animation cleaner? Right now it opens to the side and immediately snaps down as if the keybind was used

Thesola10 (approved):
:+1:

schemas/org.gnome.shell.extensions.paperwm.gschema.xml
Thesola10: In fact reusing the widget that `dconf` uses for that kind of input would be better to tidy up the settings view

jtaala: These are for keybinds... so they can't be enums.

jtaala: Note that the implementation is using an enum:

https://github.com/paperwm/PaperWM/blob/ba27c1bcaf71230911b5183c080b1267132a4dc1/tiling.js#L41

tiling.js
jtaala: Not currently - the whole slurping and layouting concept is built on the assumption that windows need to be correctly tiled before slurping.

PR #872, opened 2024-06-06
swipe: fix focus behavior

valpackett:
The unrelated direction variable was used as the direction argument there, and the intended direction was passed as an unused third argument.

---

With this fix, which window gets the focus when swiping at least makes some sense, it kinda didn't before. I'd like to also add a mode where it's the window under the cursor that ends up focused…


jtaala:
Nice find!

jtaala (approved):
Good find @valpackett.

PR #877, opened 2024-06-13
FIX for #709: Apps not using full height in workspaces without top bar

### LINKED ISSUE #709: Apps not using full height in workspaces without top bar
I hide the top bar on all workspaces except one, and when switching between workspaces, apps in workspaces where the top bar is hidden get resized as if the top bar is there so they are not taking the maximum height.

Apps in workspaces without the top bar should not take the bar height into consideration. 

**Additional context**

https://github.com/paperwm/PaperWM/assets/97781843/fb3e56f4-5479-4b5f-a6e1-957fc7e110d4

Edit:
Even if the bar is hidden in all workspaces, it still happens because the bar is shown while switching workspaces, I tried disabling the bar on the desktop with the Just perfection extension and apps still have the full height after switching between workspaces, So maybe an option to not show the top bar while switching workspaces would fix the issue?

jtaala:
FIXES #709.

This PR refactors and rewrites the layout logic to more explicitly address workArea height when a space has been set to `Hide top bar`:

image

This PR also:
- adds a new option for workspaces to _Hide Window Position Bar_ (overrides showing window position bar just for a specific space(s);
- updates some default margin settings to align with changes made in #822;
- improves various code modules with formatting and alignment with PaperWM linter;
- provides and implements safer (checked) substitutes for `actor` `add_child`, `remove_child` (for example, will check if a child exists before calling `remove_child`... cleans up unneeded errors in logs);

Finally, this PR also adds some convenience keybinds to toggle show/hide of the topbar and window position bar:

image


jtaala:
~~Converted to draft.  Still got a few margin issues and the topbar showing when moving a window between monitors.~~

Think I've resolved issues for this now.

PR #883, opened 2024-06-26
Implement mouse scrolling on topbar to  scroll viewport

### LINKED ISSUE #206: Scroll windows with mouse wheel
How about a new feature of scrolling Windows(the same which I assume is now happening when scrolling with three fingers on the touchpad on Wayland) but when scrolling with the mouse wheel in the top bar?  
I think it will greatly improve the usability of PaperWM when using only mouse.

jtaala:
Implements #206.

Allows switching windows with mouse scrolling on the gnome topbar.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -120,7 +120,7 @@ rules:
     - error
     - allow:
       - '!!'
-  no-invalid-this: error
+  # no-invalid-this: error
   no-iterator: error
   no-label-var: error
   no-lonely-if: error
--- a/gestures.js
+++ b/gestures.js
@@ -195,7 +195,7 @@ export function gestureWorkspaceFingers() {
    connected from each space.background and bound to the space.
  */
 let start, dxs = [], dts = [];
-export function horizontalScroll(space, actor, event) {
+export function horizontalScroll(space, _actor, event) {
     if (event.type() !== Clutter.EventType.TOUCHPAD_SWIPE) {
         return Clutter.EVENT_PROPAGATE;
     }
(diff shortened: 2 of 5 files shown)


jtaala:
~~Note: it currently doesn't work well with the Gnome pill icon (events getting intercepted and it gets a bit confused).~~

I believe I've now fixed this.

jtaala:
See #881.

grota:
tested and it works, thank you!

jtaala:
@Thesola10, can I get you to test this one out on a touchscreen?  ~~We're making the topbar reactive in this PR - don't think it would interfere with the touchscreen stuff but we better check.~~

Now using `Main.panel` (which is already reactive).  I have move the touchscreen signals from `captured-event` to `touch-event` though (for efficiency as `captured-event` captures every single event and is heavy.  Let me know if you see any issues on the latest commit here.

Thesola10:
`touch-event` used to cause issues with the scroll event where for some reason the selected window frame would update, but not the window itself. It seems to have been resolved

Seems however like the scroll handling is conflicting with vertical workspaces extensions such as V-Shell, just a heads-up

(that would have been an approving review :p)

jtaala:
> Seems however like the scroll handling is conflicting with vertical workspaces extensions such as V-Shell, just a heads-up

Thanks @Thesola10.  Hmm - I suppose we can make the topbar scroll handling an option (so users can disable it - if it messes with v-shell or other extensions).

jtaala:
hmm... it might be that we're `return Clutter.EVENT_STOP;` to our handling on the topbar doesn't bubble the events (once we handle them).

What are you seeing in V-Shell that's busted? (just so I can debug and check if the above fixes it).

Thesola10:
it seems both PaperWM and V-Shell try to handle the scroll wheel event, sometimes the PaperWM selection is moving but mostly V-Shell wins.

Then again, maybe it's a Dash to Panel thing, but nothing is "broken" per se.

jtaala:
hmm - seems v-shell is working fine for me.  Will have a quick looksy at dash-to-panel.

jtaala:
eh, either way it might be worth adding an option to disable the topbar scrolling.

PR #885, opened 2024-06-27
Added Nix build files for bleeding-edge NixOS

Thesola10:
This allows NixOS users to directly use PaperWM source as their extensions package, which tracks changes faster than even Nixpkgs ever could.

It's also a very low-maintenance patch, maybe requiring someone run `nix flake update nixpkgs` once in a blue moon to bump the dependency locks.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,27 @@
+{ pkgs, stdenv, glib, ... }:
+
+let
+  uuid = "paperwm@paperwm.github.com";
+in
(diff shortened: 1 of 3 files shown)


jtaala (approved):
Cool - I don't understand the files (as I'm not a nix user...yet?).

Anyways, approving this as can't see any harm in adding these files.

Thesola10:
> (as I'm not a nix user...yet?).

soon...

PR #886, opened 2024-06-27
Improve topbar mouse scrolling feature to work better with other extensions

jtaala:
@Thesola10 mentioned some potential conflicts with the new mouse scrolling (on topbar) feature (#883) and other extensions (e.g. V-Shell and/or dash-to-panel).

Small change in how PaperWM: propagate scroll-event signal after switching windows (in case other extensions require the event).

Not sure if it will make a difference though.

### DIFF
--- a/topbar.js
+++ b/topbar.js
@@ -164,13 +164,13 @@ export function disable() {
 export function topBarScrollAction(event) {
     // if topbar workspaceMenu (indicator) has pointer, exit
     if (menu && menu.has_pointer) {
-        return Clutter.EVENT_STOP;
+        return Clutter.EVENT_PROPAGATE;
     }
 
     // same check for gnome pill
     const pill = Main.panel?.statusArea?.activities;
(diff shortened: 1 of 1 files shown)


jtaala:
@Thesola10.  When you get a chance can you test/see if this branch changes anything with V-Shell and/or dash-to-panel?

jtaala:
Not sure if it will - since if those extentions catch the event (and don't propagate) then PaperWM might not get the event.  At least with these changes PaperWM is propagating the signal for others.

jtaala:
P.S. I couldn't reproduce any issues with V-Shell and dash-to-panel - but that might be just the settings I used in those extensions.

Thesola10:
Can confirm it works now!

Thesola10 (approved):

jtaala:
yay!

jtaala:
Dang - will do another release now.

Issue #730, opened 2023-12-07
Add a keyboard shortcut for "Swipe tilling windows"


bilboquet:
Hello

Most of the time, my main screen is arranged like this:
![image](https://github.com/paperwm/PaperWM/assets/8612342/339d039d-1acc-4e2a-81a2-69f9900f669b)
- A browser on the left (or on the right) to display some documentation/information
- My main application in "the center" (more or less)
- A terminal on one side to monitor logs, etc

Using a touchpad, it's possible to make windows flow from left to right and right to left with 3 fingers (default configuration).
That's the touchpad gestures: Swipe tilling windows (swipe left/right).
That's useful because, without changing my main task (so focus still on the window in the center) I can have more feedback from the left or from the right.
I'm often found myself simulating this behavior with the keyboard only by quickly focussing the window on one side then coming back to the center one.
The default is that it's not always very precise.

So I wish I have keyboard shortcuts that act like the swipe gesture on the touchpad.
It's probably better to leave this kind of shortcut disabled by default.
4 shortcuts would be ideal:
- swipe left smoothly
- swipe right smoothly
- swipe left roughly (same as above but by "jump")
- swipe right roughly

Please let me know what's your thought.

jtaala:
I can see the utility here - could be useful on X11 too (which doesn't support gnome mouse gestures).

KETHERCORTEX:
Is there a way to invoke this functionality by launching a command (in terminal or with GNOME Alt+F2 launcher)?

jtaala:
> Is there a way to invoke this functionality by launching a command (in terminal or with GNOME Alt+F2 launcher)?

Hey @KETHERCORTEX,

No, I don't believe (AFAIK) there's anything like this for gnome extensions - other that simulating touchpad swipes.

This would be nice though, thinking something like `i3-msg`.

Lythenas:
I think I mentioned this in another issue, but we could provide a dbus interface for our "actions". That could be used from outside.

jtaala:
Hey @Lythenas - yes, this is a good idea.  In PaperWM we use a dbus hook to get enriched monitor info.

Can you add an issue? (can reference this issue for example).

Thanks @Lythenas.

Lythenas:
@jtaala I created an issue: #738

jtaala:
Hey @bilboquet,

Please see PR #917.  Are you able to test this please?

```
git fetch --all
git checkout improve-730-swipe-drifting
./install.sh
```

[...]

PR #914, opened 2024-07-30
Drift tiling viewport left/right when mouse pointer is at monitor edge (while dragging a window)

### LINKED ISSUE #765: Add scroll hotspots to window drag-and-drop
This is especially noticeable with touch screens, where there is no alternate way of moving the scroll view while dragging a window. By PaperWM's very nature, repositioning a window across the scroll view is hard without using the keyboard, as window movement is restricted to the spaces currently visible on screen.

Moving a tiled window close to the edge of the screen should progressively start scrolling the workspace in that direction, like file drag-and-drop operations in Nautilus for instance.

### LINKED ISSUE #910: Scrolling when reordering windows
When I am trying to reorder my windows using a touchpad, I cannot scroll the view. This makes it so you have to either use the keyboard or go into the overview and swap the positions there. 

When pulling down on the window (and with that moving it), you should still be able to scroll the view with three fingers (one will be clicking, since you are moving the window, and the other two could scroll the view).

The alternative could be the view automatically scrolling when you reach the edge of the screen when moving the window. Or a preview could appear at the edge of the screen, meaning that the window will automatically tile after the window that you are currently hovering.

jtaala:
Fixes #910 and #765.

This PR allows scrolling the tiling viewport (left/right) when dragging a window and mouse touches the monitor left/right edge:

[Screencast from 2024-07-30 22-55-52.webm](https://github.com/user-attachments/assets/a2a452c2-d91b-479f-8545-30db32531f7f)

I've also added a setting to control the "drift" speed:
image


jtaala:
Functionality can be ported to resolve #730.

Thesola10 (approved):
Nice! Works just fine with touch gestures as well, a good bump for touch usability

PR #917, opened 2024-08-01
Implement keyboard-based "swipe tiling windows" (tiling viewport drifting)

### LINKED ISSUE #730: Add a keyboard shortcut for "Swipe tilling windows"

jtaala:
This PR resolves #730.

This PR provides two new keybinds which effectively "swipe" the tiling viewport left/right (similar to trackpad behaviour, but keyboard-based):

Default keybinds are `ctrl`+`shift`+`left`/`right`.

Below I'm using this functionality using just the keyboard (not trackpad):

https://github.com/user-attachments/assets/b5e42f3c-99e8-4598-846d-c784155a24bb

You can change the drift speed in advanced settings:
image


Thesola10 (approved):
looks good, thanks!

Issue #911, opened 2024-07-30
Add animation for centering window horizontally on scratch layer


zhaob1n:
**Is your feature request related to a problem? Please describe.**
I generally like the animation all over paperwm. But there seems no animation for centering window horizontally on scratch layer.

**Describe the solution you'd like**
I'd love it if there are animations for that action.

jtaala:
This would be a very low priority for me, but would accept a PR if you would like to work on this.

jtaala:
Completed by #920.

Issue #912, opened 2024-07-30
Add "center window vertically" or "center window" action for windows on scratch layer


zhaob1n:
**Is your feature request related to a problem? Please describe.**
For windows on scratch layer, center window horizontally works, but I think there should be a method to center window vertically and also center window both horizontally and vertically.

usecases: there could be many windows on scratch layers without tiling, these two functions are useful for window management on scratch layer.
**Describe the solution you'd like**
implement center window vertically and also center window both horizontally and vertically.

**Describe alternatives you've considered**
currently no, I think this is straightforward.

glupi-borna:
The only thing that comes to mind here is that these bindings should be specific to scratch windows, because tiled windows can not be centered vertically, so the bindings would probably be called "center scratch window vertically" and "center scratch window". @jtaala, am I correct in my thinking here?

jtaala:
@glupi-borna, yes you are correct.

I wouldn't use this though, but if anyone here thinks would be useful for their workflows definitely feel free to create a PR for this and I'd be happy to look at it.

Cheers.

dlealmt:
There is already an option `org.gnome.mutter.center-new-windows` which automatically centers windows, both vertically and horizontally, but it seems like PaperWM doesn't respect that. Would it make sense to respect that as an alternative to the keybinding to center windows?

jtaala:
Could someone please check this PR out and test?

```
git clone https://github.com/paperwm/PaperWM.git
cd PaperWM
git checkout improve-center-window-keybinds
./install.sh
```

then logout / login.

New keybind is `super`+`v` to center scratch windows vertically (`super`+`c`+ centers both horizontally/vertically).  Let me know how it feels.

Jay.

zhaob1n:
It seems great!  
Btw I'm not sure if there should be an individual option for "horizontally only". We might as well add it for symmetry in case someone need it.

PR #920, opened 2024-08-05
Ease scratch windows when centering horizontally

glupi-borna:
I was poking around the code and found the `easeScratch` function, which looked to me like it could be used to implement #911. I'm not sure if this is the right function to call because it's used only in one other place (in `makeScratch`), but it seems to work fine at first glance -- though I don't know if there are edge cases to consider.

### DIFF
--- a/scratch.js
+++ b/scratch.js
@@ -42,10 +42,11 @@ export function disable() {
    other windows/clones (clones if the space animates)
  */
 export function easeScratch(metaWindow, targetX, targetY, params = {}) {
-    let f = metaWindow.get_frame_rect();
-    let b = metaWindow.get_buffer_rect();
-    let dx = f.x - b.x;
-    let dy = f.y - b.y;
+    const complete = params?.onComplete ?? function() {};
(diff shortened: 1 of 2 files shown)


jtaala:
Nice one @glupi-borna.  

You're on the right track!  Although that method/approach is a little old and over-complicated (but doing what we want).

Let's update those methods in your branch and get you to test.

Jay.

glupi-borna:
What an absolute ninja! :D
I'll give it a whirl ASAP!

jtaala:
So, I was thinking of having it also center vertically as well (which kind of makes sense for scratch windows).

jtaala:
> So, I was thinking of having it also center vertically as well (which kind of makes sense for scratch windows).

Eh, maybe just do this with #912 (e.g. add an extra shortcut).

glupi-borna:
Yup, I was actually investigating #912 when I noticed `easeSearch`. I can't believe you already got it done! You're super fast.

PR #924, opened 2024-08-06
Add keybind for centering scratch window vertically (center horiztonally/vertically by default for scratch window)

### LINKED ISSUE #912: Add "center window vertically" or "center window" action for windows on scratch layer

jtaala:
Fixes #912.

image


glupi-borna (approved):
This seems to work nicely! The only thing I'm noticing is that the 'center-horizontally' action now centers scratch windows on both axes. I'm not sure if that is intentional or not, but I'm not opposed to it working that way at all.

EDIT: Nvm, just saw your comment where you explained that it is intentional. Good stuff, one less keybind to remember! :)

jtaala:
> This seems to work nicely! The only thing I'm noticing is that the 'center-horizontally' action now centers scratch windows on both axes. I'm not sure if that is intentional or not, but I'm not opposed to it working that way at all.

I might change this though.  It feels good but isn't quite "center horizontally" as the keybind suggests.  I'll add another keybind specifically for `center window` (including non-tiled).  Will probably leave it unmapped by default.

PR #935, opened 2024-08-25
Update README.md to explain where settings are

pjsxw:
As a new user, it took a bit of googling to figure out where to find GNOME extension settings. This PR adds a hint to the README for that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ Most keybindings start with the <kbd>Super</kbd> modifier (by default), which is
 
 Most keybindings will grab the keyboard while <kbd>Super</kbd> is held down, only switching focus when <kbd>Super</kbd> is released. <kbd>Escape</kbd> will abort the navigation taking you back to the previously active window.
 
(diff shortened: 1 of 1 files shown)


Lythenas (commented):
Thanks for the addition.

I would maybe say "Extensions app" instead of just extensions to be even more clear.

Btw. I usually access the settings by right clicking the workspace name on the top left and clicking "Workspace preference".

AndrewKvalheim:
It’s never actually called “Extensions app” anywhere though. Maybe make it a link?

> [Extensions](https://apps.gnome.org/Extensions/) → PaperWM → ⋮ → Settings

Lythenas:
yeah, maybe `"Extensions" app` is better. It should just be clear where the _entrypoint_ of that chain is. (an app instead of some menu in e.g. the gnome settings). Maybe I'm overthinking this :smile: 

Adding the link is fine, although I would expect that this is installed by default.

jtaala:
Hey @pjsxw, looks good.  Were you going to add a link for the `Extensions` (is a good idea).

Add the link and we can approve/merge in.

pjsxw:
Yeah, adding a link makes it clearer what is being referred too. Have updated

PR #950, opened 2024-09-24
Add winprop to select space for new window

Lythenas:
Fixes #949

This adds a new winprop that allows selecting the space a new window should be inserted.

Screenshot From 2024-09-24 12-00-30

I added a dropdown to select the workspace. Not sure if that may cause problems with dynamic workspaces etc. Maybe the simple textbox was better. If so, we can simply revert the last commit.

Note: Sometimes it feels a bit confusing because you don't see anything happening if a window opens on another workspace. E.g. on firefox if you drag a tab out of the window to create a new window it is simply gone. But I think that is kind of expected for this. Maybe adding another option to automatically switch to the workspace and focus the window would be good.

PR #956, opened 2024-10-03
Dynamic workspace behaviour: add behaviour option that works better with #950

jtaala:
This PR adds an option for dynamic workspaces that works much better with #950 and dynamic workspaces:

#950 now allows defining a space (via winprops) to always open applications on.  With dynamic workspaces, PaperWM will create workspaces if the set workspace does not exist (which is likely with dynamic workspaces).  This can cause an issue with workspace names as default dynamic workspaces will remove any empty workspaces.

I've added a PaperWM setting which handles this more consistently and intuitively:

image


Lythenas:
This seems fine to me, but you should merge it after #950 or rebase this PR.

PR #967, opened 2024-11-10
feat: auto width

Hoffmano:
# Auto Width

The purpose of this PR is to optimize monitor space by automatically managing window widths. Cropped or partially visible windows can detract from the user experience and provide little value, so this module will automatically set the width of windows based on a specified number of columns.

This change is designed to save time and remove the need for manual resizing, enhancing usability and screen utilization.

# To Do

- [x] Auto width based on quantity of columns

[Starter discussion](https://github.com/paperwm/PaperWM/discussions/965#discussion-7436917)


jtaala:
It looks like you've branched off a very (very) old version of PaperWM that wouldn't be compatible with Gnome 45+.

It's also using now deprecated approaches (such as `dynamic_function` refs etc.).

As such, this one wouldn't be accepted or merged in (i.e. you wouldn't be able to merge this branch in into later PaperWM versions).

Any reason you're using Gnome 42 branch? (i.e. Gnome 42 released in 2022 and well passed end-of-life.  Gnome 44 is also EOL now as well).

Hoffmano:
@jtaala 

I'm a Pop!_OS user, which currently uses GNOME 42.

If you and the other collaborators think it would be beneficial for PaperWM, I'd be open to working on adapting it for other GNOME versions in a virtual machine.

For now, I'm just focusing on what’s useful to me.

image

Hoffmano:
My first demo


https://github.com/user-attachments/assets/18dcf9ff-a1dc-4380-b7e3-f39c27401a1f

dudumaroja:
Im new paperwm user. i was looking for some auto adjust windows funcions in option.. and found none..

what i really miss is when you you have two windows on screen and when with our mouse you grab the middle bar and drag it to resize both windows at same sime

and maybe a new sortgut like shift+super+r the active window and the next window auto ajust to predefineld useful states like  75/25, 50/50 and 25/75

PR #984, opened 2025-01-09
prefer horizontal_margin to minimum_margin

### LINKED ISSUE #975: Maximized window width should respect horizontal margin
**Describe the bug**
Now `Super`+`=` can increase the window width until `screen_width - 2*horizontal_margin`, and `Super`+`F` sets the width to `screen-width - 2*minimum_margin`, where `minimum_margin` is initialized as `Math.min(15, prefs.horizontal_margin)`.
And if I switch to the maximized window which is off screen quickly (without showing the minimap), its width will be decreased to a smaller value, otherwise the width remains unchanged. This only happens when an external monitor is connected.

Plus, when I swipe the view quickly, the first or last window will "snap" to the screen edges, with a gap of `minimum_margin` but I think this should be `horizontal_margin`. Also when I switch the focus mode from EDGE to DEFAULT it will snap to screen edges with a gap of `window_gap` instead of `horizontal_margin`.

**To Reproduce**
1. Set *horizontal margin* to 64.
2. Maximize a window by `Super`+`F` (its width would be `screen_width - 2*minimum_margin`, while I believe it should be `screen_width - 2*margin`).

**Expected behavior**
*Horizontal margin* and *vertical margin* define the maximum view region in the screen so I believe the active window should never exceed it.

**System information:**

```
Distribution: Arch Linux
GNOME Shell: 47.2
Display server: Wayland
PaperWM version: 47.1.0
Enabled extensions:
- Bluetooth-Battery-Meter@maniacx.github.com
- lockkeys@vaina.lt
- AlphabeticalAppGrid@stuarthayhurst
- paperwm@paperwm.github.com
- moveclock@kuvaus.org
- custom-osd@neuromorph
- kimpanel@kde.org
- runcat@kolesnikov.se
- blur-my-shell@aunetx
- gnome-ui-tune@itstime.tech
```

lost-melody:
prefer horizontal_margin for:

1. toggling windows maximize horizontally
2. snapping to left/right edges in swipe gestures

fixes #975.

### DIFF
--- a/tiling.js
+++ b/tiling.js
@@ -4305,10 +4305,10 @@ export function ensuredX(meta_window, space) {
     } else if (frame.width > workArea.width * 0.9 - 2 * (Settings.prefs.horizontal_margin + Settings.prefs.window_gap)) {
         // Consider the window to be wide and center it
         x = min + Math.round((workArea.width - frame.width) / 2);
-    } else if (x + frame.width > max) {
+    } else if (x + Settings.prefs.horizontal_margin + frame.width > max) {
         // Align to the right prefs.horizontal_margin
         x = max - Settings.prefs.horizontal_margin - frame.width;
-    } else if (x < min) {
+    } else if (x < min + Settings.prefs.horizontal_margin) {
         // Align to the left prefs.horizontal_margin
         x = min + Settings.prefs.horizontal_margin;
     } else if (x + frame.width === max) {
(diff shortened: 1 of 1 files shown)


Thesola10:
image

Your PR is on top, current `develop` on bottom. I can definitely see an improvement -- makes horizontal margin more useful. That would mean that users who want more real-estate (old behavior) should either lower horizontal margin (default value makes this change unnoticeable) or disable PaperWM maximization to favor fully maximized windows.

I'm rebasing this on latest `develop` so others can try it out in the [builtin VM](https://github.com/paperwm/PaperWM/wiki/Using-the-test-VM) and give feedback

Thesola10 (approved):

Issue #991, opened 2025-01-24
crash "Could not grab modal"


joeyh:
**Describe the bug**

Paperwm crashes with "Could not grab modal" and I have to restart gnome shell in order to switch between windows.

Here's the top lines from journalctl --user after the crash:

```
Jan 24 13:25:43 darkstar alacritty[3796845]: ../../../../modules/im/ximcp/imDefLkup.c,419: The application disposed a key event with 133011 serial.
Jan 24 13:25:41 darkstar gnome-shell[1814058]: JS ERROR: Error: Could not grab modal
                                               ActionDispatcher@file:///home/joey/.local/share/gnome-shell/extensions/paperwm@paperwm.github.com/navigator.js:88:19
                                               getActionDispatcher@file:///home/joey/.local/share/gnome-shell/extensions/paperwm@paperwm.github.com/navigator.js:521:18
                                               preview_navigate@file:///home/joey/.local/share/gnome-shell/extensions/paperwm@paperwm.github.com/navigator.js:548:39
                                               asKeyHandler/<@file:///home/joey/.local/share/gnome-shell/extensions/paperwm@paperwm.github.com/keybindings.js:368:51
                                               @resource:///org/gnome/shell/ui/init.js:21:20
Jan 24 13:25:41 darkstar gnome-shell[1814058]: Failed to grab modal
```

**To Reproduce**

I don't have a reliable way to cause this crash, but I have seen it repeatedly when all I have open is some alacritty windows, probably fullscreen, and am switching back and forth between them using paperwm hotkeys.

**System information:**
Distribution: Debian GNU/Linux
GNOME Shell: 47.3
Display server: Xorg
PaperWM version: 47.1.0
Enabled extensions:
- netspeedsimplified@prateekmedia.extension
- executor@raujonas.github.io
- just-perfection-desktop@just-perfection
- paperwm@paperwm.github.com
- battery-usage-wattmeter@halfmexicanhalfamazing.gmail.com
- drive-menu@gnome-shell-extensions.gcampax.github.com

icyflame:
@joeyh Thanks for this issue report. I have run into the same issue multiple times on Ubuntu 22.04 and Gnome 44 with this extension enabled. 

I don't know much about Gnome Shell extensions, but looking through the code, this looks like it is caused by a function calling `pushModal` => failing to get a grab (?) over the keyboard => not restoring the state to the way that things were _before_ `pushModal` and instead failing immediately.

[...]

PR #1031, opened 2025-05-09
Fail gracefully instead of crashing if grab fails with the error "Could not grab modal"

icyflame:
I have been using PaperWM on a daily basis for the past few months and it is great. One of the problems that I kept running into repeatedly was https://github.com/paperwm/PaperWM/issues/991. This issue was reported on Gnome 47. I ran into this problem on Gnome 42.9.

I attempted a naive fix which simply tries to clean-up all the state instead of crashing. This works well; I have been using a version of this fix, applied on top of the gnome-44 branch, for the past month and I have not suffered any `Could not grab modal` crashes. I have backported this change to the `gnome-44` branch as well: https://github.com/paperwm/PaperWM/pull/1032.

It would be great if I could get some guidance about whether this fix looks logical, and whether there might be a way to reproduce this code path somehow. We need to somehow get into `preview_navigate` (which happens when one uses a keyboard shortcut to switch windows, I believe?) but get there in a state where the keyboard is already grabbed by something else, thus causing PaperWM to fail to grab the keyboard. 

Meanwhile, I will continue to use PaperWM daily on two machines (albeit with old Gnome versions: 3.38 and 42.9) I will attempt to get PaperWM up and running on a more recent version and check to see if I run into this crash on the default branch; if I do, then, I will also see whether this fix solves the crash.

I would really like to fix this issue within PaperWM (upstream) because I have not run into any other bug at all and this is a great project!

### DIFF
--- a/navigator.js
+++ b/navigator.js
@@ -82,10 +82,21 @@ class ActionDispatcher {
 
         // grab = stage.grab(this.actor)
         grab = Main.pushModal(this.actor);
+		// Assume that grab succeeds and store that state in the object
+		this.success = true;
         // We expect at least a keyboard grab here
         if ((grab.get_seat_state() & Clutter.GrabState.KEYBOARD) === 0) {
             console.error("Failed to grab modal");
-            throw new Error('Could not grab modal');
+			// Release current grab and let the user try again
+			try {
+				this.success = false;
+				if (grab) {
+					Main.popModal(grab);
+					grab = null;
+				}
+			} catch (e) {
+				console.error("Failed to release grab");
+			}
         }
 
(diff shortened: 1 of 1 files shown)


jtaala:
See comments on #1032.

Thanks @icyflame - I haven't seen this issue on later Gnome versions (e.g. current 47 etc.).  Although, that may be due to how I use PaperWM etc.

Let me know if succeed in seeing this issue in latest gnome and PaperWM.  If so, I'm happy to run this branch for a few days just to make sure it doesn't introduce any other issues (and if not, can merge).

Cheers.

PR #1052, opened 2025-08-08
Nested colum

### LINKED ISSUE #1035: Need Nested Column Within Row
Often i need to split my screen like this

Image

50% to code, 50% to monitor device I'm currently working on. In this setup i need maximum horizontal space for wireshark, that is ok. But for two serial monitor below, I actually need maximum vertical space but i prefer not having them on separate column as it would require me to scroll right and i will lose view for vscodium. Or when i need to focus on vscodium i won't be able to see my terminal, if they occupies separate column.

Why paperWM does not implement "nested" column i.e. a row may contain column on their own. I think many other tiling window manager usually implement these things.

Like i said before, having terminal and wireshark on their own columns make them too thin to fit into single view.

oisanjaya:
New feature: Nested column inside rows.


Thesola10:
I'll mark your PR as draft for now, mark it as ready once everything is ready to be merged. Appreciate the contribution!

PR #1055, opened 2025-08-17
fix: add delay before activateWithFocus call

SanTeeg0:
**Bug**
Having Winprop that:
1. Should insert window to specific workspace.
2. Should focus window as its created.
causes Gnome Shell crash when currently active workspace is different from workspace in which window was inserted.

**Fix**
This PR fixes it, buy adding delay before trying to activate window.

### DIFF
--- a/tiling.js
+++ b/tiling.js
@@ -4218,7 +4218,9 @@ Opening "${metaWindow?.title}" on current space.`);
         else {
             delete metaWindow.focusOnOpen;
             console.debug("#winprops", "focusing space of inserted window");
-            spaces.spaceOfWindow(metaWindow)?.activateWithFocus(metaWindow, false, true);
+            Utils.later_add(Meta.LaterType.IDLE, () => {
+                spaces.spaceOfWindow(metaWindow)?.activateWithFocus(metaWindow, false, true);
+            });
         }
     }
 


tiling.js
```diff
@@ -4218,7 +4218,9 @@ Opening "${metaWindow?.title}" on current space.`);
         else {
             delete metaWindow.focusOnOpen;
             console.debug("#winprops", "focusing space of inserted window");
-            spaces.spaceOfWindow(metaWindow)?.activateWithFocus(metaWindow, false, true);
+            Utils.later_add(Meta.LaterType.IDLE, () => {
```
smichel17, line 4221: Just out of curiosity, did you try a shorter delay, like `BEFORE_REDRAW`?

([Docs](https://gjs-docs.gnome.org/meta12~12/meta.latertype#default-idle) link for others' curiosity / future reference)

smichel17 (approved):
Thank you! I have not yet tested this myself but the code looks good and is only called under the conditions you described, so it should not have any unintended side effects → safe to merge.

I'll give others (who are more involved in maintenance) a few days to look at this, then merge if nobody else has comments.

tiling.js
SanTeeg0: No, i actually did not. Will try and hit back with results.

SanTeeg0: So i tried, and it seems like all variants of `Meta.LaterType` fix initial issue.

Thesola10 (approved):
Did summary testing, lgtm

SanTeeg0:
Is there anything i have to do on my end? Or everything is fine, and it already will be merged whenever possible?

smichel17:
Just me to not forget about it :laughing:

smichel17:
@SanTeeg0 Thanks for the PRs. I've added you (and @openendings) to the repo so you can merge your own PRs now (after approval; important branches have branch protection rules), and can approve other people's PRs so they can merge.

I've also removed or lowered the access level of folks who made small contributions and haven't been active recently — @bilboquet, @gombosg, @jonseppanen, and @YaLTeR. This is just housekeeping (giving out access easily means we need to clean up to keep the list manageable); if you become active again we can add back your access :)

PR #1061, opened 2025-08-28
use default gsettings schemas when installed system-wide

niol:


SanTeeg0 (changes requested):
Thank you for your contribution!
I think it would be better to keep the default behavior of `make install`(install the extension locally for user), as it's what will be used most often anyway. This will also ensure some backward compatibility.
Other than that, everything looks good to me.

niol:
I understand. What do you want as target name for system-wide install?
Please note `make install` is the standard way of installing software on Unixes system-wide.

SanTeeg0:
> Please note `make install` is the standard way of installing software on Unixes system-wide.

That's true, but mainly for packages, tools, etc. Extensions, themes, and so on are mostly user-specific. Also, PaperWM only had one way for users to install it locally, which was with `make install`, so changing this behavior could cause more confusion.

> I understand. What do you want as target name for system-wide install?

Something like `make install-system` could be sufficient.

SanTeeg0 (changes requested):
Looks great! 
But currently, the only option is to install the extension in the system using `make install-system`, but I think it would be great if there was a way to uninstall it. Especially using the same `make uninstall` command (to remove both the local and system versions, if any).

Sorry for my oversight, I could have noticed this earlier.

Makefile
```diff
@@ -8,18 +8,20 @@ SOURCE      := $$PWD
 EXT_ID      := paperwm@paperwm.github.com
 EXT_DIR     := $(XDG_DATA_HOME)/gnome-shell/extensions
 TARGET      := $(EXT_DIR)/$(EXT_ID)
+SYSTARGET   := $(DESTDIR)/usr/share/gnome-shell/extensions/$(EXT_ID)
```
Thesola10: This and the `install`/`install-system` dichotomy are really unnecessary -- a system-wide builder will update `EXT_DIR` to point to `/usr/share/gnome-shell/extensions/...` (see how our `default.nix` does it)

At worst, I'd add a `GSCHEMA_DIR` and have the `install` target try to install the schemas there if nonempty (or set it to `$(TARGET)/schemas` by default). This way complexity remains at a minimum and the Makefile can easily be made to install gschemas system-wide as desired.

Thesola10 (changes requested):

Makefile
Thesola10: Under this system, what you want to achieve would be done by running the following command:

```sh
make EXT_DIR=pkg/usr/share/gnome-shell/extensions GSCHEMA_DIR=pkg/usr/share/glib-2.0/schemas install
```

niol: The [common interface](https://www.gnu.org/prep/standards/html_node/DESTDIR.html) is what I need:
```
$ make install DESTDIR=pkg
```
I'll propose something conforming to this common interface with conditions as to where to install gschemas.

Thesola10: A good start would be to just rename `EXT_DIR` into `DESTDIR` then

niol: I'm not sure how the script should choose between making a symlink or copying files around.

niol:
> Looks great! But currently, the only option is to install the extension in the system using `make install-system`, but I think it would be great if there was a way to uninstall it. Especially using the same `make uninstall` command (to remove both the local and system versions, if any).

Then I'll need `$prefix` so that I'll try to work out something.

niol:
The `uninstall` target is currently broken due to `/bin/sh` syntax errors when run on my machine.

Thesola10 (approved):

PR #1071, opened 2025-09-19
Add codespell support (config, workflow to detect/not fix) and make it fix some typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,7 @@
+[codespell]
+# Ref: https://github.com/codespell-project/codespell#using-a-config-file
+skip = .git*,*.svg,*.lock,*.css,.codespellrc
+check-hidden = true
+# ignore-regex =
+# initialY: variable name for initial Y coordinate
+ignore-words-list = easer,positionOf,initialY
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ package-lock.json
 
 # generated disk image for test VM
 nixos.qcow2
+.claude/
--- a/grab.js
+++ b/grab.js
@@ -203,7 +203,7 @@ export class MoveGrab {
         this.selectDndZone(space, sx, sy);
     }
 
-    /** x,y in scroll cooridinates */
+    /** x,y in scroll coordinates */
     selectDndZone(space, x, y, initial = false) {
         const gap = Settings.prefs.window_gap;
         const halfGap = gap / 2;
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,25 @@
+# Codespell configuration is within .codespellrc
+---
+name: Codespell
+
+on:
+  push:
+    branches: [release]
+  pull_request:
+    branches: [release]
+
+permissions:
+  contents: read
+
+jobs:
+  codespell:
+    name: Check for spelling errors
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Annotate locations with typos
+        uses: codespell-project/codespell-problem-matcher@v1
+      - name: Codespell
+        uses: codespell-project/actions-codespell@v2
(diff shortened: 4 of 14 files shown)


grab.js
```diff
@@ -416,7 +416,7 @@ export class MoveGrab {
         clone.y = y - dy;
 
         const threshold = 300;
-        dy = Math.min(threshold, Math.abs(clone.y - this.initialY));
+        dy = Math.min(threshold, Math.abs(clone.y - this.initially));
```
smichel17: This one is wrong. Can it be smarter about camelCase and/or only linting comments?

smichel17 (changes requested):
The rest looks good

grab.js
yarikoptic: oh, missed that. We can add regex to exclude all camelCases, e.g. `\b[a-z][A-Za-z]+\b' (Python regex), do you want me to do that? or we could exclude some specific ones like this to minimize possible false negatives.

yarikoptic:
should I rebase into `develop`?

PS FWIW, in some projects we eventually switched everywhere default branches to be the ones to seek PRs into to avoid hassle of needing to ask to rebase etc.  Should may be `develop` to become the default branch to be cloned / sent PRs to?

smichel17:
Yes, all PRs should go to `develop`.

I wasn't involved with the decision to set up the branches how they are now (See [CONTRIBUTING.md](https://github.com/paperwm/PaperWM/blob/release/CONTRIBUTING.md) for history (it's short)). But `release` is the default to facilitate people who are obtaining paperwm from github instead of getting it through extensions.gnome.org (at one point during the transition to a community project, paperwm was not on extensions.gnome.org).

We could probably do a better job of handling the install process (I'd guide people to use the `gnome-*` branches instead of `release`. Personally I use a git worktree to install my running version, rather than the symlink that the install script creates. (Related: #1061)

grab.js
okias: I think `initalY` is not very well readable, and if there is no strong reason to keep it that way, I would suggest renaming the variable.

yarikoptic:
re-based etc, might be ready for the prime time! ;-)

.gitignore
```diff
@@ -11,3 +11,4 @@ package-lock.json
 
 # generated disk image for test VM
 nixos.qcow2
+.claude/
```
Thesola10, line 14: sneaky ;)

.codespellrc
```diff
@@ -0,0 +1,7 @@
+[codespell]
+# Ref: https://github.com/codespell-project/codespell#using-a-config-file
+skip = .git*,*.svg,*.lock,*.css,.codespellrc
+check-hidden = true
+# ignore-regex =
+# initialY: variable name for initial Y coordinate
+ignore-words-list = easer,positionOf,initialY
```
Thesola10, line 7: good idea, but https://github.com/codespell-project/codespell/issues/196 still not being implemented gives me pause as to the quality of codespell itself...

Thesola10 (approved):
Looks ready for prime time

grab.js
Thesola10: readability notwithstanding, `initialY` was added to exceptions

.gitignore
yarikoptic: want me to filter it out?

.codespellrc
yarikoptic: oh, if that was the only concern... all software is constant WiP / trove of skeletons etc... only visidata (or replace with paperwm if you like) is perfect! ;)

yarikoptic: FWIW -- I used codespell across HUNDREDS of projects with good return rate... here is some overview of state of my endavors on that end if you like to see more 'victims' of typos-less world of mine: https://github.com/yarikoptic/improveit-dashboard/blob/master/READMEs/yarikoptic.md

PR #1074, opened 2025-09-22
Add shellcheck support and some fixes to make shell scripts "cleaner"

yarikoptic:
More info is at https://www.shellcheck.net about shellcheck

Note that the last commit was done mostly by `claude code` but upon my review should be kosher.

In view of potential desire for moving some logic from Makefile (see #1073 ) into external scripts, might especially be nice to ensure that shell scripts code is more robust

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -1,0 +1,26 @@
+---
+name: Shellcheck
+
+on:
+  push:
+    branches: [release]
+  pull_request:
+    branches: [release]
+
+permissions:
+  contents: read
+
+jobs:
+  shellcheck:
+    name: Check shell scripts
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Install dependencies
+        run: |
+          sudo apt update && sudo apt install -y shellcheck
+      - name: shellcheck
+        run: |
+          git grep -l '^#\( *shellcheck \|!\(/bin/\|/usr/bin/env \)\(sh\|bash\|dash\|ksh\)\)' | xargs shellcheck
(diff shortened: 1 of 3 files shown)


shell.sh
```diff
@@ -35,13 +35,13 @@ case $1 in
         echo "Running X11 Gnome Shell"
         Xephyr $NEW_DISPLAY &
         DISPLAY=$NEW_DISPLAY
-        args=--x11
+        args=(--x11)
```
smichel17, line 38: We're not really using the array, so I think we might as well make this `x11_arg="--x11"` here and `gnome-shell "$x11_arg"` later

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,26 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Install dependencies
+        run: |
+          sudo apt update && sudo apt install -y shellcheck
+      - name: shellcheck
+        run: |
+          git grep -l '^#\( *shellcheck \|!\(/bin/\|/usr/bin/env \)\(sh\|bash\|dash\|ksh\)\)' | xargs shellcheck
```
smichel17, line 26: What does `*shellcheck` do?

shell.sh
yarikoptic: well, then better be named also `x11_arg` to be factually correct? since if `args` then would assume multiple and thus treated accordingly ;-)

.github/workflows/shellcheck.yml
yarikoptic: https://www.shellcheck.net/ says in the title "finds bugs in your shell scripts." ;)  
Quite often it is so, and most often could be observed if you run any script under path with a space in some folder name . (that is why in our datalad tool testing we even come up with some most obscure path filesystem can tollerate to ensure we always correctly quote etc)

smichel17: Ah, I should have clarified. I know shellcheck (it's great, I use it for all my shell scripts). I was asking about the string "`*shellcheck`" in the `git grep` regex.

PR #775, opened 2024-02-07
Added switch-global-* actions

dawsers:
This pull request is to add four new actions that complement `switch-right`, `switch-left`, `switch-up` and `switch-down` with *global* versions that change window regardless of monitor boundaries. This means the action will not respect the current monitor boundaries and will select a window on the adjacent monitor if the previous selection was at a boundary. For example, `switch-global-right` when the selected window is the rightmost of a *space* will move to the monitor on the right and select the first window. While within monitor boundaries, the behavior is exactly the same as `switch-*`.

1. It doesn't change other actions, you can always keep on changing monitors with `<Super><Shift><right/left/up/down>`, which preserves the last selected window on each *space* when moving back and forth, while `switch-global-*` actions don't (they move to the *"closest"* window).
2. There are no default key bindings for these actions. For example, I have replaced the ones for `switch-*` with the actions `switch-global-*`, but by default, the change doesn't do anything.

The pull request is backwards compatible, transparent, and needs the user to enable the behavior knowingly; it doesn't add any key bindings by default.

PR #1076, opened 2025-09-22
Global switch windows - move to **visually** closest window

smichel17:
## Overview

I use paperWM with multiple monitors, but have always struggled with the hotkeys for movement. I've memorized them out of necessity, but…

<details><summary>…let's be honest, they make no sense.</summary>
<p>

- `<Super>` with `Left` and `Right` arrows, or `comma`/`period`, are the hotkeys for switching windows. Ok.
- Add either `Shift` or `Ctrl` to comma/period to drag the window instead. Great.
- For arrows, `Ctrl` works the same way, but shift suddenly moves the focus between monitors. ???
- Okay so to move the current window to a different monitor, now I use `<Super><Ctrl><Shift>Left`. That's a lot of keys but I guess once I understand that `<Shift>Left` is moving monitors then adding control grabs the window.
- `<Ctrl><Alt>Left` moves the entire workspace. Why does a less common, larger action, require fewer modifiers than moving just one window? Why doesn't it use `<Meta>` like all other PaperWM "Window manager" hotkeys?

I can go on, but I think my point is made.

</p>
</details>

Recently I saw the new actions added in https://github.com/paperwm/PaperWM/pull/775, and it was a serious :bulb: :exploding_head: revelation for me. _Obviously_ this is how PaperWM is _supposed_ to work. If you are switching left and you reach the end of one monitor, the same hotkey just _keeps going left_.

Well, kinda. There were a few things missing, which I have added in three separate PRs:

1. #1076 « :pushpin: YOU ARE HERE
2. #1081
3. #1082

## This PR

Intuitively, I expected that "keep going left" would focus the window that I already see active when looking at the monitor to the left. If I'm only using the keyboard it works great, but if I have the windows like this:

Two workspaces; each letter is a window; uppercase means focused.
```
[A b c d] [e f g h]
```
And each window is 50% size, so they're currently visible on two monitors like this:
```
[A b] [e f]
```
If I use my mouse to select E
```
[a b] [E f]
```
And then use the keyboard shortcut to move left, I get whiplash as the left monitor whisks everything away to focus D.
```
[c D] [e f]
```
This PR changes it to just activate the closest _visible_ window on the monitor.
```
[a B] [e f]
```

@dawsers Can you play around with this and tell me what you think of it?

### DIFF
--- a/tiling.js
+++ b/tiling.js
@@ -1219,6 +1219,10 @@ export class Space extends Array {
                 return;
             }
 
+            // Ensure if we change workspaces and then monitors,
+            // that the new workspace stays active on the starting monitor.
+            space.activateWithFocus(space.selectedWindow, false, true);
+
             let newMonitor = Main.layoutManager.monitors[i];
             space = spaces.monitors.get(newMonitor);
             if (dir === Meta.DisplayDirection.LEFT) {
--- a/tiling.js
+++ b/tiling.js
@@ -1190,6 +1190,14 @@ export class Space extends Array {
     switchGlobalUp() { this.switchGlobal(Meta.MotionDirection.UP); }
     switchGlobalDown() { this.switchGlobal(Meta.MotionDirection.DOWN); }
     switchGlobal(direction) {
+        const motionToDisplayDirection = {
+            [Meta.MotionDirection.LEFT]: Meta.DisplayDirection.LEFT,
+            [Meta.MotionDirection.RIGHT]: Meta.DisplayDirection.RIGHT,
+            [Meta.MotionDirection.UP]: Meta.DisplayDirection.UP,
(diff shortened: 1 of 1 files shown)


smichel17:
> smichel17 wants to merge 7 commits into develop from sm/universal-move


This branch name is foreshadowing… after polishing a bit, I think we should make this the default hotkey for movement.

dawsers:
> 
> @dawsers Can you play around with this and tell me what you think of it?

Sorry, I am no longer using PaperWM, I ended up writing my own thing, [scroll](https://github.com/dawsers/scroll).

I don't remember how it works, but if what you say is what happens, then you are totally right, the more intuitive behavior would be to focus on the closest window on the other monitor that can be seen.

smichel17:
No worries. Thank you for the inspiration!

PR #1077, opened 2025-09-23
Fix compatibility with GNOME 49

valpackett:
Maximization API rename broke PaperWM:
https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/4415

### DIFF
--- a/tiling.js
+++ b/tiling.js
@@ -3424,7 +3424,7 @@ export function isMaximized(metaWindow) {
     return metaWindow.get_maximized() === Meta.MaximizeFlags.BOTH;
 }
 
-export function isMaximizedHorizontal(metaWindow) {
+function isMaximizedHorizontal(metaWindow) {
     if (!metaWindow) {
         return false;
     }
--- a/tiling.js
+++ b/tiling.js
@@ -571,8 +571,7 @@ export class Space extends Array {
 
             let f = mw.get_frame_rect();
 
-            let resizable = !mw.fullscreen &&
-                mw.get_maximized() !== Meta.MaximizeFlags.BOTH;
+            let resizable = !mw.fullscreen && !isMaximized(mw);
 
             if (mw.preferredWidth) {
                 let prop = mw.preferredWidth;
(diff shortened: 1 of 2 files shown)


victornoel:
FYI I tested this locally and it seems to work ok.

Thesola10 (approved):

tiling.js
```diff
@@ -3417,6 +3415,33 @@ export function isScratch(metaWindow) {
     return Scratch.isScratchWindow(metaWindow);
 }
 
+export function isMaximized(metaWindow) {
```
smichel17, line 3418: :information_source: (Just a comment, nothing to change here)

:+1: I think these wrappers around gnome-shell functionality are great. The less one needs to know about gnome-shell internals in order to contribute high-level functionality to PaperWM, the better.

In the future, we should talk about organizing all the stuff that's currently in `tiling.js`.

- Currently I don't see any clear logic for when to create a top level function, a method on `Space`, or a method on `Spaces`.
- There's not a clear separation of space-local operations and global ones (that work based on the current focused monitor/workspace/window).
- We don't have any public/private separation, which means that renaming anything might break someone's customizations (e.g. stuff in `examples/`).

Today though, I don't have a good enough understanding of _all the stuff that exists_ to know how to organize it, so here is fine for now. Since they're not exported, we can always move them later.

tiling.js
```diff
@@ -3417,6 +3415,33 @@ export function isScratch(metaWindow) {
...
+    if (!metaWindow) {
+        return false;
+    }
+    if (typeof metaWindow.is_maximized === 'function') // GNOME >= 49
+        return metaWindow.is_maximized();
+    return metaWindow.get_maximized() === Meta.MaximizeFlags.BOTH;
+}
+
+export function isMaximizedHorizontal(metaWindow) {
```
smichel17:
```suggestion
function isMaximizedHorizontal(metaWindow) {
```

smichel17 (changes requested):
Thank you!

smichel17 (approved):

capfredf:
I am confused here. why does this PR still require approvals given two reviewers have already approved the changes?

Thesola10:
@capfredf Probably because the topmost commit is by smichel17 and GitHub doesn't count the last committer's approval

PR #1081, opened 2025-09-26
Global move -- allow switching to empty monitors

smichel17:
This is part of a series of PRs. See the first one for more context.

1. #1076
2. #1081 « :pushpin: YOU ARE HERE
3. #1082

Updates the "select window or monitor to the <direction>" to move to/from empty monitors as well as monitors with windows.

### DIFF
--- a/tiling.js
+++ b/tiling.js
@@ -1197,22 +1197,9 @@ export class Space extends Array {
             [Meta.MotionDirection.DOWN]: Meta.DisplayDirection.DOWN,
         };
         const dir = motionToDisplayDirection[direction]
-
         let space = this;
-        let index = space.selectedIndex();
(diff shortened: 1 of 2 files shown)


tiling.js
```diff
@@ -1221,26 +1208,53 @@ export class Space extends Array {
...
+            break;
+        case Meta.MotionDirection.LEFT:
+            index--;
+        }
+        if (index < 0 || index >= space.length) {
+            switchMonitor();
+            return;
         }
 
         let column = space[index];
```
smichel17, line 1260: TODO: apply to switching vertically, too. So far I was just focused on horizontally because that's what I use; I don't have any monitors laid out vertically.

PR #1082, opened 2025-09-26
Add global move selected window hotkey

smichel17:
This is part of a series of PRs. See the first one for more context.

1. #1076
2. #1081
3. #1082 « :pushpin: YOU ARE HERE

This is backwards compatible. However, I think we should eventually make this the default (after I have used it for a long period of time and the animations are improved). So, I updated the text to make this one described as just "move window to the left" and clarified that the current default stays on the same monitor.


tiling.js
```diff
@@ -1095,6 +1095,45 @@ export class Space extends Array {
...
+            targetRow--;
+            break;
+        }
+        if (targetIndex < 0 || targetIndex >= this.length) {
+            spaces.switchMonitor(dir, true, true, metaWindow);
+            return;
+        }
+        let column = this[index];
+        if (targetRow < 0 || targetRow >= column.length) {
+            // TODO: Move to workspace above/below
```
smichel17: I need to give more thought to how switching up/down should work when there is a monitor above.

smichel17 (commented):
Neither this nor the other PRs are my cleanest code… there is a fair amount of duplicate index-finding now that we could probably clean up. But I wasn't sure how much the implementation would diverge, so I opted for copy-pasting the relevant pieces of logic instead of reusing functions, to give freedom to change whatever is needed.

PR #1095, opened 2025-10-14
Fix crash when drag detaching Chrome tabs

### LINKED ISSUE #1069: PaperWM Layout Breaks And Causes Unexpected Behavior
PaperWM breaks the layout, now if i open anything (included all the system apps)  it looks buggy and won't fit the screen it is out of the screen and things like border on focused window also doesn't visible and windows are overlapping each other like in the scratch layer 

**Steps to reproduce the behavior:**
1. Go to Chrome,Brave or Chromium
2. Create Multiple Tabs
3. Try to move any tab out to the new window

https://github.com/user-attachments/assets/c28f12ee-2d90-46da-a142-67cf295e54e2

**Note**
This won't happen on firefox or vivaldi browser .

https://github.com/user-attachments/assets/e40989e0-d260-42af-bebc-3d1ae9d94840

**System information:**

Distribution: Fedora Linux 42 (Workstation Edition)
GNOME Shell: 48.4
Display server: Wayland
PaperWM version: 48.0.3
Enabled extensions:
- color-picker@tuberry
- clipboard-indicator@tudmotu.com
- extension-list@tu.berry
- status-area-horizontal-spacing@mathematical.coffee.gmail.com
- caffeine@patapon.info
- paperwm@paperwm.github.com
- Vitals@CoreCoding.com
- top-bar-organizer@julian.gse.jsts.xyz
- Bluetooth-Battery-Meter@maniacx.github.com
- places-menu@gnome-shell-extensions.gcampax.github.com
- gsconnect@andyholmes.github.io

chromakode:
Fixes #1069

When detaching tabs from Chrome by dragging them out of the current window, PaperWM calls `layoutGrabColumn` with a null `selectedWindow`, causing the layout recalculation to crash.

This PR changes it to pass `inGrab.window` directly for the `grabWindow` parameter, which seems to match the if predicate better too.

I also fixed an incidental crash when formatting the error in `layoutGrabColumn` when null gets passed in.

Finally, when testing this out, I noticed that `layoutGrabColumn` positions the detached window with a somewhat arbitrary seeming y position (seems to be based on where it was created on detach):

https://github.com/user-attachments/assets/78d8893f-cd63-406e-8633-f61d44afd552

Adding `inGrab.dnd` to the predicate uses the normal layout function and prevents this behavior:

https://github.com/user-attachments/assets/6bfa0c30-4c0d-4ce5-a255-27b59ea0067c

I don't yet fully understand the cases where `layoutGrabColumn` is important, but it seems like the gap preservation only matters when we're in the fully engaged "DnD grab" where the window is pulled out of the layout and being moved to a new tiled position.

There's still a little funkiness I've seen occasionally in testing, but this change prevents the stateful layout corruption described in the original issue.

### DIFF
--- a/tiling.js
+++ b/tiling.js
@@ -744,7 +744,7 @@ export class Space extends Array {
             if (inGrab && column.includes(inGrab.window) && !allocator) {
                 [resultingWidth, relayout] =
                     this.layoutGrabColumn(column, x, y0, targetWidth, availableHeight, time,
-                        selectedInColumn);
+                        inGrab.window);
             } else {
                 allocator = allocator || allocateDefault;
                 let targetHeights = allocator(column, availableHeight, selectedInColumn);
--- a/tiling.js
+++ b/tiling.js
@@ -741,7 +741,7 @@ export class Space extends Array {
 
             let resultingWidth, relayout;
             let allocator = allocators && allocators[i];
-            if (inGrab && column.includes(inGrab.window) && !allocator) {
+            if (inGrab && inGrab.dnd && column.includes(inGrab.window) && !allocator) {
                 [resultingWidth, relayout] =
                     this.layoutGrabColumn(column, x, y0, targetWidth, availableHeight, time,
                         inGrab.window);
--- a/tiling.js
+++ b/tiling.js
@@ -541,7 +541,7 @@ export class Space extends Array {
 
         const k = column.indexOf(grabWindow);
         if (k < 0) {
-            throw new Error(`Anchor doesn't exist in column ${grabWindow.title}`);
+            throw new Error(`Anchor doesn't exist in column ${grabWindow?.title}`);
         }
 
         const gap = Settings.prefs.window_gap;


chromakode:
When a Chrome tab is being detached in this way, `inGrab.begin` / `inGrab.beginDnD` never get called. I explored calling them, because ideally the new detached window you're grabbing would enter the "DnD grab" where you can select a new tiled position. However, there seemed to be more going on here, because upon release the detached window doesn't end up in the tiled position selected.

Perhaps Chrome is doing something more here and recreating the window upon grab end.

I'd love to get to the bottom of this but wanted to keep the scope of this PR small.

yarikoptic:
@chromakode in the context of 

- #1098

do you think there is some way to "script" triggering such a bug e.g. with https://github.com/jordansissel/xdotool or alike?

chromakode:
> @chromakode in the context of
> 
> * [Integration testing under "fake" GNOME instance on github CI #1098](https://github.com/paperwm/PaperWM/issues/1098)
> 
> do you think there is some way to "script" triggering such a bug e.g. with https://github.com/jordansissel/xdotool or alike?

A simple approach could be to capture snapshots of the mutter events received in PaperWM and have a way to replay them on the core code. That'd make smaller unit tests that run faster (and are potentially less flaky than setting up a whole Wayland environment) at the cost of being potentially less resilient if Mutter or Wayland change their event format. That kind of test would cover this bug well.

yarikoptic:
sounds interesting but would it make it wayland specific if I got it right?

even if not a full way to do scripting of tests, I think it might be useful to get documentation (and/or helpers) on how to capture/replay so user issues could be attempted to be reproduced.

Thesola10 (approved):

conceptualspace:
any idea when this will be included in a release?

PR #1097, opened 2025-10-15
Reduce minimum fixed workspaces to the number of monitors

### LINKED ISSUE #1087: PaperWM requires 2 workspaces despite only 1 physical monitor connected
**Describe the bug**
PaperWM displays the error message "PaperWM requires a minimum of 2 workspaces for your monitor configuration" even when only 1 physical monitor is connected and I try to set fixed workspaces to 1. PaperWM appears to be detecting a virtual or phantom monitor that the user cannot see or control.

**To Reproduce**
1. Connect only 1 physical monitor (verified with `xrandr --listactivemonitors` showing only one monitor)
2. Open GNOME Settings → Multitasking
3. Attempt to set 'Number of Workspaces' to 'Fixed Number' with value of 1
4. See error: "PaperWM requires a minimum of 2 workspaces for your monitor configuration"

**Expected behavior**
 PaperWM should allow setting the number of fixed workspaces to 1.

**Screenshots**
Image
Image

**System information:**
```
Distribution: NixOS 25.05 (Warbler)
GNOME Shell: 48.2
Display server: Wayland
PaperWM version: 48.0.2
Enabled extensions:
- AlphabeticalAppGrid@stuarthayhurst
- paperwm@paperwm.github.com
- transparent-top-bar@ftpix.com
```

**Additional context**
Related to #1057.

zabenn:
Fixes #1087.

Removes the requirement for an empty workspace at the end when using a fixed number of workspaces. Each monitor gets exactly one workspace minimum instead of having an extra unused one. Tested on my machine and is working fine.

### DIFF
--- a/patches.js
+++ b/patches.js
@@ -583,7 +583,7 @@ export function _checkWorkspaces() {
     let workspaceManager = global.workspace_manager;
     let i;
     let emptyWorkspaces = [];
-    let minimum = Main.layoutManager.monitors.length + 1;
+    let minimum = Main.layoutManager.monitors.length;
 
     if (!Meta.prefs_get_dynamic_workspaces()) {
         // if less spaces than minimum, create!
@@ -640,8 +640,7 @@ export function _checkWorkspaces() {
     }
 
     /**
-     * Set minimum workspaces to be max of num_monitors+1.
-     * This ensures that we have at least one workspace at the end.
+     * Set minimum workspaces to be max of num_monitors.
      */
     // Make sure we have a minimum number of spaces
     for (i = 0; i < minimum; i++) {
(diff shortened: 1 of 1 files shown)


smichel17 (changes requested):
This works when you have fixed workspaces, but breaks things like "move window to workspace below" when you are using dynamic workspaces. So you'd need to make this conditional based on that gnome setting.

zabenn:
> This works when you have fixed workspaces, but breaks things like "move window to workspace below" when you are using dynamic workspaces. So you'd need to make this conditional based on that gnome setting.

Should be fixed

zabenn:
@smichel17 thoughts?

Thesola10 (approved):
Behaves as expected.

zabenn:
It seems like @smichel17 has been away for a while and won't be able to update his review status anytime soon, what should we do with the PR?

PR #1125, opened 2026-01-07
Switch windows with Super+Mousewheel (Example)

LDAP:
As described here https://github.com/paperwm/PaperWM/issues/206#issuecomment-3719761093 this is an example how to get Super+Mousewheel to work for window switching. 

These are my personal patches which also invert all scroll directions (topbar, alttab, ...), so probably no good candidate for merging until toggles for the features are added.


ecrevisseMiroir:
Any idea when this is coming? Seems like the cherry on top

Issue #1137, opened 2026-02-03
3-finger overview gesture gets stuck on gnome-shell 49.3


Fringale:
**Describe the bug**
Since upgrading to gnome-shell 49.3 on my system, the three-finger vertical touchpad gesture used to enter and exit the Activity Overview gets stuck in an intermediate state.

When performing the gesture, swiping up to enter the overview or down to return to the desktop no longer snaps to its final state: the animation stops just before completion and remains indefinitely in a “gesture in progress” state.

While stuck:
* Mouse clicks no longer register.
* Keyboard input continues to work.
* The only way found so far to exit this state is to perform three-finger horizontal swipes, which force the gesture to complete.

The other three-finger vertical gesture (the PaperWM-specific one to switch workspaces) is not impacted by this issue.

No relevant logs appear in `journalctl` while the shell is in this stuck state.

**To Reproduce**
1. Make sure to be on gnome-shell 49.3
2. Perform a three-finger vertical swipe to enter the overview
3. Try to end the gesture by swiping down to get back to the desktop

**Expected behavior**
The gesture snaps back to the desktop.

**Actual behavior**
The gesture reaches near completion but never snaps, leaving the shell stuck in a partially transitioned state. Further vertical swiping is possible but do not change the outcome.

**System information:**
```
Distribution: Aeon (openSUSE Tumbleweed derivative)
GNOME Shell: 49.3
Display server: Wayland
PaperWM version: 49.0.0
Enabled extensions:
- paperwm@paperwm.github.com
```

**Additional context**
This is probably related to the recent Gnome Shell gesture handling changes introduced in
https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/3997.

The issue is reproducible both with and without the `patches.js` change introduced in https://github.com/paperwm/PaperWM/pull/1106.

t3dium:
Also encountering this bug, the 3 finger overview gesture causes gnome shell to hang & be unresponsive.

Using alpine, with gnome 49.3.

> The only way found so far to exit this state is to perform three-finger horizontal swipes, which force the gesture to complete.

Thanks for mentioning this by the way, good temporary fix for the time being for anyone else who accidentally triggers overview.

YounesJY:
👋 Another user confirming this issue

I can also confirm this bug on GNOME 49.3/49.4 with PaperWM. Interestingly, I have two separate Linux installations on the same hardware, which allows me to compare:

✅ Working correctly:

[...]

PR #1138, opened 2026-02-17
Fix gesture handoff and propagation to overview

TecN01R:
Proposed fix for [Issue 1137 ](https://github.com/paperwm/PaperWM/issues/1137)

Add a handoffToOverview flag and use it to avoid swallowing touchpad END/UPDATE events when GNOME overview should handle them, preventing the overview from getting stuck. Ensure shouldPropagate branches return events so GNOME can receive them. Tighten horizontal gesture handling by ignoring updates/ENDs if the current direction is not horizontal and clear direction on end. Use optional chaining for navigator.finish() to avoid errors, and return the result from the captured-event handler to propagate the gesture outcome.

Tested on fresh Fedora GNOME 49.4 install with no other extensions. Tested with base extension settings with both 3-finger and 4-finger settings.

### DIFF
--- a/gestures.js
+++ b/gestures.js
@@ -15,6 +15,7 @@ const DIRECTIONS = {
 };
 
 let vy, time, vState, navigator, direction, signals;
+let handoffToOverview = false;
 // 1 is natural scrolling, -1 is unnatural
 let natural = 1;
 export let gliding = false; // exported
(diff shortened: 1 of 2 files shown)


Thesola10 (approved):

Fringale:
Thanks a lot for this! On my system (Aeon, GNOME Shell 49.4) I confirm it fixes https://github.com/paperwm/PaperWM/issues/1137.

PR #1141, opened 2026-03-13
Introducing NixOS end-to-end test framework

### LINKED ISSUE #91: Add testing
We want to add tests (perhaps CI if it's possible too).

One way to do testing is setting `XDG_CONFIG_HOME`, populating it with a `user.js` which then would run the tests.

A simple test for something like fullscreen reactivity (which I broke twice recently):
- create a window, wait for it to appear
- set fullscreen
- check that actor.visible == true and clone.visible == false
- signal completion to the test runner, with eg. `emit('done', [ok, result])`

A more complex example would be restarts on x11:
- create some windows
- save some state
- restart
- check that we've restarted
- signal completion

### LINKED ISSUE #1098: Integration testing under "fake" GNOME instance on github CI

I see many problems/fixes do not come with tests and there is no way to script triggering them. E.g. recent

- https://github.com/paperwm/PaperWM/pull/1095

And I think it could be very handy to have some "play script" which performs some tricky actions on an active instance thus codifying triggers for prior crashes etc, or at least smoke tests on a running instance.

FWIW -- we do something like that for our https://github.com/ReproNim/reprostim/blob/master/.github/workflows/ci-cd.yml where we drive under xvfb and take screenshots to potentially later identify and fix issues.

with quick google I found

- https://www.reddit.com/r/gnome/comments/t5lxkm/enable_continuous_integration_tests_for_your/

with references there is which could be handy on how to establish testing under gnome session using podman.  Some tools like xdotool could potentially be used to drive the session but there might be even better ways.

May be some other WMs already have something like that, and a more complete solution could be borrowed?

Overall, just thought to drop a note since failed to find a similar issue so far.

Thesola10:
The time has finally come! This system should allow interactive tests by simulating the mouse and keyboard and comparing against screenshots (and maybe also ensure GNOME hasn't crashed).

Also fixes #91 and #1098 

### To-do list:
- [x] Wire Behave into NixOS test system
- [x] Play back input devices using `libinput`
- [x] Run GJS code within GNOME Shell
- [x] Wait until PaperWM has finished loading
- [x] Take screenshots using OpenCV
- [x] Produce test report artifact with screenshots attached
- [ ] Extract useful information such as window positions from OpenCV
- [ ] Write a complete, meaningful test
- [ ] Stretch goal: obtain [coverage data](https://github.com/GNOME/gjs/blob/master/gjs/coverage.cpp)


Thesola10:
I'd love to introduce code coverage as well as [GJS seems to support it](https://github.com/GNOME/gjs/blob/master/gjs/coverage.cpp) but with GNOME being welded together I can't even begin to figure out how to use it right now.

Thesola10:
I think I'll merge this once the test functions all work and start implementing the tests themselves in a separate PR.

PR #1142, opened 2026-03-19
Fix #1139 by assuming Wayland by default

### LINKED ISSUE #1139: GNOME 50 support
GNOME 50 has introduced an incompatibility that prevents the extension from running.

This is the error:

    TypeError: (intermediate value).is_wayland_compositor is not a function

Also, shell-version in [metadata.json](https://github.com/paperwm/PaperWM/blob/443a2789f082194cccfd73b035811846879a85ca/metadata.json#L7) does not yet include "50".

Thesola10:
Fixes #1139 by abstracting `is_wayland_compositor()`. On GNOME 50+, Shell will only ever run on Wayland, so function missing should mean we're running on Wayland.

### DIFF
--- a/gestures.js
+++ b/gestures.js
@@ -24,7 +24,7 @@ let touchpadSettings;
 export function enable(extension) {
     signals = new Utils.Signals();
     // Touchpad swipes only works in Wayland
-    if (!Meta.is_wayland_compositor())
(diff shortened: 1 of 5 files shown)


Thesola10:
@skewty Can you give this a try to see if it fixes things on your end?

PR #1146, opened 2026-04-01
respect picture-options

Thimoteus:
picture-options (zoom, centered, etc) changes for desktop background were being ignored due to a hardcoded ZOOM setting


Thesola10 (approved):
Looks good! Will be merging as part of 50.0

PR #1149, opened 2026-04-08
Fix for GNOME 50s API changes

JaniSoto:
This PR addresses the breaking API changes introduced in GNOME 50 that cause PaperWM to crash during keybindings and window movement.

## Changes
- **Native GNOME 50 Cursor Support:** Migrated cursor management in `grab.js` from the legacy `Meta.Cursor` enums to the modern `Clutter.CursorType` system via `global.stage.set_cursor_type`. This ensures mouse icons (GRABBING, DEFAULT) function correctly on GNOME 50.
- **GrabState Polyfill:** Implemented a hoisted `var GrabState` fallback across `grab.js`, `navigator.js`, and `tiling.js`. This prevents `ReferenceError` crashes during the extension's circular import phase while providing the necessary constants for internal tiling logic.
- **API Safety Checks:** Added a type check for `get_seat_state` in `navigator.js` to prevent `TypeError` on GNOME 50's unified input grabs.

## Testing
Tested on GNOME 50 (CachyOS). 
- Super + Arrow keys (navigation) are fully functional.
- Mouse cursors update correctly when grabbing/moving windows.
- No initialization errors on extension load/reload.

Fixes #1147

PR #1150, opened 2026-04-09
Merge #1149 into develop

Thesola10:
Merges GNOME 50 testing into the development branch for preliminary release


stefangweichinger:
Ah, that's why the branch disappeared while I changed from desktop to laptop ;-)
thanks all

PR #1151, opened 2026-04-10
Fix touch screen drag freezing input

### LINKED ISSUE #1127: Drag using touchscreen freezes the environment.
### Describe the bug
When I drag a window in PaperWM using a finger on a touchscreen, the environment freezes, so keys and mouse input and touch inputs are ignored, until Escape is pressed.

### To Reproduce
1. Have windows open.
2. Drag a window down by the titlebar with a finger on the touchscreen until the blue silhouette showing where it would be dropped appears.
3. Still dragging the window, drag it back up and drop it.
4. Try to control the window using mouse or keyboard or touchscreen, and see that nothing happens.
5. Press Escape, and see that everything works normally.

### Expected behavior
I expect that the environment would not freeze after dragging.

### Screenshots
https://github.com/user-attachments/assets/ca4f4195-37c8-4eae-8440-955aabb881c8

It is not seen in the recording, but the point is that, after the drag, I click and it does not work, and I press Escape, and it works again.

### System information
Distribution: Arch Linux
GNOME Shell: 49.2
Display server: Wayland
PaperWM version: 49.0.0
Enabled extensions:
- paperwm@paperwm.github.com
- drive-menu@gnome-shell-extensions.gcampax.github.com
- system-monitor@gnome-shell-extensions.gcampax.github.com
- caffeine@patapon.info

JaniSoto:
Fixes #1127. 

**Cause:**
The code was successfully intercepting `TOUCH_END` and running `this.end()`. Workaround of faking a virtual mouse click to drop the grab only works for mouse/pointer grabs. Mutter ignores the fake mouse click if the grab was started by a touch sequence, leaving the grab stuck open.

**Fix:**
Since hitting `Esc` acts as a global kill-switch for Mutter grabs, this PR adds a `virtualKeyboard` device. Alongside the fake mouse click, it now fires a virtual `Escape` keypress during cleanup. This successfully breaks the touch grab. 

Tested on CachyOS / GNOME 50 and it works perfectly!

### DIFF
--- a/grab.js
+++ b/grab.js
@@ -586,7 +586,7 @@ export class MoveGrab {
         Utils.later_add(Meta.LaterType.IDLE, () => {
             if (!global.display.end_grab_op && this.wasTiled) {
                 let time = Clutter.get_current_event_time();
-                
+
                 // For Mouse users: Fakes a mouse click to break Mutter's pointer grab
                 let [x, y] = global.get_pointer();
                 getVirtualPointer().notify_absolute_motion(time, x, y);
--- a/grab.js
+++ b/grab.js
@@ -38,22 +38,11 @@ export function disable() {
 }
 
 /**
- * Returns a virtual pointer (i.e. mouse) device that can be used to
- * "clickout" of a drag operation when `grab_end_op` is unavailable
+ * Returns a virtual keyboard that can be used to
+ * "escape" of a drag operation when `grab_end_op` is unavailable
  * (i.e. as of Gnome 44 where `grab_end_op` was removed).
  * @returns Clutter.VirtualInputDevice
 */
-let virtualPointer;
-export function getVirtualPointer() {
-    if (!virtualPointer) {
-        virtualPointer = Clutter.get_default_backend()
-            .get_default_seat()
-            .create_virtual_device(Clutter.InputDeviceType.POINTER_DEVICE);
-    }
-
-    return virtualPointer;
-}
-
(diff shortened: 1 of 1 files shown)


grab.js
```diff
@@ -575,16 +585,17 @@ export class MoveGrab {
...
-                    Clutter.get_current_event_time(),
-                    x, y);
-
-                getVirtualPointer().notify_button(Clutter.get_current_event_time(),
-                    Clutter.BUTTON_PRIMARY, Clutter.ButtonState.PRESSED);
-                getVirtualPointer().notify_button(Clutter.get_current_event_time(),
-                    Clutter.BUTTON_PRIMARY, Clutter.ButtonState.RELEASED);
+                getVirtualPointer().notify_absolute_motion(time, x, y);
+                getVirtualPointer().notify_button(time, Clutter.BUTTON_PRIMARY, Clutter.ButtonState.PRESSED);
+                getVirtualPointer().notify_button(time, Clutter.BUTTON_PRIMARY, Clutter.ButtonState.RELEASED);
```
Thesola10: I wonder if the mouse click is necessary if we use Escape now. Ideally we'd keep the focus fix to a single action

JaniSoto: You're right.. I've removed the virtual mouse now in the latest commit. Both mouse and touch drags still work perfectly fine!

Thesola10 (approved):

PR #1166, opened 2026-06-10
gestures: don't disable swipe trackers mid-gesture (GNOME 49/50 overview hang)

### LINKED ISSUE #1165: Overview gesture hangs / shell freezes after a preceding 3-finger gesture on GNOME 49/50 (SwipeTracker enabled=false → _interrupt → invalid overview transition)
**Describe the bug**

On GNOME 49/50, doing a 3-finger horizontal scroll and then *immediately* a 3-finger swipe up to the overview hangs the overview mid-animation. Pressing <kbd>Super</kbd> at that point freezes the shell completely.

The shell logs a burst of overview state-machine errors when it happens:

```
JS ERROR: Error: Invalid overview shown transition from HIDDEN to HIDING
JS ERROR: Error: Invalid overview shown transition from SHOWING to SHOWING
JS ERROR: Error: Invalid overview shown transition from HIDING to HIDING
```

**Environment**

- PaperWM 50.0.1 (v148)
- GNOME Shell 50.2, Wayland
- Fedora Silverblue 44

**To Reproduce**

1. Have a couple of windows tiled in a space.
2. 3-finger swipe left/right to scroll windows.
3. *Immediately* 3-finger swipe up to the overview.
4. Overview sticks part-way; pressing <kbd>Super</kbd> then hangs the shell.

**Root cause**

GNOME 49 rewrote `SwipeTracker` (`js/ui/swipeTracker.js`). The touchpad path is now a `TouchpadSwipeGesture` connected via `actor.connectObject('event::touchpad', …)` plus a `Clutter.PanGesture` action, and the `enabled` setter does this:

```js
set enabled(enabled) {
    if (this._enabled === enabled)
        return;
    this._enabled = enabled;
    if (!enabled && this._state === State.SCROLLING)
        this._interrupt();   // emits a synthetic 'end' with cancelProgress
    this.notify('enabled');
}

_interrupt() {
    this.emit('end', 0, this._cancelProgress);
    this._panGesture.cancel();
    this._reset();
}
```

So on GNOME 49+, calling `swipeTrackersEnable(false)` **while the overview tracker is mid-gesture** (`State.SCROLLING`) no longer just flips a flag — it fires a synthetic `end`, which drives the overview into an illegal state transition (`HIDDEN→HIDING`, `SHOWING→SHOWING`, …). The animation never completes and the shell wedges.

PaperWM disables trackers from several places that can fire while an overview hand-off gesture is live. The one that triggers it on every up-swipe is the `Main.overview 'hidden'` handler in `gestures.js`, which runs during the hand-off and disables the still-scrolling tracker. The 300 ms `pillSwipeTimer` in `patches.js` can do the same when a previous gesture's timer lands inside the next gesture.

**Evidence**

Instrumenting the disable sites confirmed the `overview 'hidden'` handler fires *during* the hand-off (`inGesture=true`) on every reproduction. Deferring/guarding those mid-gesture disables eliminates the `Invalid overview …` errors entirely and the hang no longer occurs.

**Proposed fix**

Never disable a swipe tracker while a PaperWM gesture is in progress:

- Track an `inGesture` flag (BEGIN → END/CANCEL).
- Defer the `overview 'hidden'` disable to idle and skip it if a gesture is still active.
- Skip the `pillSwipeTimer` disable while a gesture is active.

This keeps the follow-finger overview hand-off intact. PR follows.

woongzeyi:
## Summary

Fixes the overview hang / shell freeze that happens when a 3-finger up-to-overview gesture immediately follows another 3-finger gesture on GNOME 49/50.

Fixes #1165

## Root cause

GNOME 49 rewrote `SwipeTracker`. Its `enabled` setter now interrupts an in-progress gesture instead of just flipping a flag:

```js
set enabled(enabled) {
    ...
    if (!enabled && this._state === State.SCROLLING)
        this._interrupt();   // emits a synthetic 'end' with cancelProgress
    ...
}
```

So calling `swipeTrackersEnable(false)` **while the overview tracker is mid-gesture** emits a synthetic `end`, which pushes the overview into an illegal state transition:

```
JS ERROR: Error: Invalid overview shown transition from HIDDEN to HIDING
```

The overview animation then never completes and the shell wedges (pressing Super afterwards freezes it).

PaperWM disables trackers from a few spots that can run while an overview hand-off gesture is live. The `Main.overview 'hidden'` handler fires during the hand-off on every up-swipe; the `pillSwipeTimer` (300 ms) can land inside the next gesture.

## Fix

Never disable a tracker while a PaperWM gesture is in progress:

- **gestures.js** — add an `inGesture` flag (set at `BEGIN`, cleared at `END`/`CANCEL`) and an `isInGesture()` accessor. Defer the `overview 'hidden'` tracker-disable to idle and skip it when `inGesture` is still set, so the in-flight overview animation settles (tracker → `State.NONE`) before anything disables it.
- **patches.js** — the `pillSwipeTimer` skips disabling trackers when `Gestures.isInGesture()` is true.

Follow-finger overview hand-off is unchanged.

## Testing

- GNOME Shell 50.2 (Wayland), PaperWM 50.0.1.
- Before: horizontal scroll → immediate up-swipe reliably hung the overview and logged `Invalid overview shown transition …`; Super then froze the shell.
- After: repeated reproductions show **zero** `Invalid overview …` errors and no hang. Verified via `journalctl --user`.

### DIFF
--- a/gestures.js
+++ b/gestures.js
@@ -16,8 +16,15 @@ const DIRECTIONS = {
 
 let vy, time, vState, navigator, direction, signals;
 let handoffToOverview = false;
+// true between BEGIN and END/CANCEL of a PaperWM-managed swipe. Used to defer
+// any tracker disabling that would otherwise interrupt a live overview gesture.
+let inGesture = false;
 // 1 is natural scrolling, -1 is unnatural
 let natural = 1;
+
+export function isInGesture() {
+    return inGesture;
+}
 export let gliding = false; // exported
 
 let touchpadSettings;
@@ -42,9 +49,19 @@ export function enable(extension) {
      * ensure swipe trackers are reset.
      */
     signals.connect(Main.overview, 'hidden', () => {
-        if (gestureEnabled()) {
-            swipeTrackersEnable(false);
+        if (!gestureEnabled()) {
+            return;
         }
+        // Defer to idle so any in-flight overview swipe animation settles
+        // (tracker -> State.NONE) before we disable. On GNOME 50 disabling a
+        // SCROLLING tracker calls _interrupt() -> synthetic 'end' -> illegal
+        // overview state transition -> stuck/hang.
+        GLib.idle_add(GLib.PRIORITY_DEFAULT, () => {
+            if (!inGesture && gestureEnabled()) {
+                swipeTrackersEnable(false);
+            }
+            return GLib.SOURCE_REMOVE;
+        });
     });
 
     /**
@@ -88,6 +105,7 @@ export function enable(extension) {
             natural = touchpadSettings.get_boolean("natural-scroll") ? 1 : -1;
             direction = undefined;
             handoffToOverview = false;
+            inGesture = true;
             navigator = Navigator.getNavigator();
             navigator.connect('destroy', () => {
                 vState = -1;
@@ -147,6 +165,7 @@ export function enable(extension) {
             return Clutter.EVENT_PROPAGATE;
         case Clutter.TouchpadGesturePhase.CANCEL:
         case Clutter.TouchpadGesturePhase.END:
+            inGesture = false;
             // If this finger count should be handled by GNOME, never consume
             // completion here (prevents overview from getting stuck in 4-finger mode).
             if (shouldPropagate(fingers)) {
--- a/patches.js
+++ b/patches.js
@@ -16,7 +16,7 @@ import * as WindowManager from 'resource:///org/gnome/shell/ui/windowManager.js'
 import * as WindowPreview from 'resource:///org/gnome/shell/ui/windowPreview.js';
 import * as Screenshot from 'resource:///org/gnome/shell/ui/screenshot.js';
 
-import { Utils, Tiling, Scratch, Settings, OverviewLayout } from './imports.js';
+import { Utils, Tiling, Scratch, Settings, OverviewLayout, Gestures } from './imports.js';
 
 /**
   Some of Gnome Shell's default behavior is really sub-optimal when using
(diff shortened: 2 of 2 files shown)


patches.js
```diff
@@ -153,6 +153,13 @@ export function setupOverrides() {
             const reset = () => {
                 // gnome windows switch animation time = 250, do that plus a little more
                 pillSwipeTimer = GLib.timeout_add(GLib.PRIORITY_DEFAULT, 300, () => {
+                    // Don't disable mid-gesture: on GNOME 50 disabling a
+                    // SCROLLING tracker fires _interrupt() -> synthetic 'end'
+                    // -> illegal overview transition -> hang.
+                    if (Gestures.isInGesture?.()) {
```
Thesola10, line 159: Why is there a null-check? This is our own module, the symbol should always exist

gestures.js
```diff
@@ -42,9 +49,19 @@ export function enable(extension) {
...
-            swipeTrackersEnable(false);
+        if (!gestureEnabled()) {
+            return;
         }
+        // Defer to idle so any in-flight overview swipe animation settles
+        // (tracker -> State.NONE) before we disable. On GNOME 50 disabling a
+        // SCROLLING tracker calls _interrupt() -> synthetic 'end' -> illegal
+        // overview state transition -> stuck/hang.
+        GLib.idle_add(GLib.PRIORITY_DEFAULT, () => {
+            if (!inGesture && gestureEnabled()) {
```
Thesola10, line 60: Consider using the `isInGesture()` accessor even within the module

Thesola10 (dismissed):
Some nitpicks

Thesola10 (changes requested):
See previous comments

PR #1173, opened 2026-07-17
Make popup windows fully visible on focus and attention-demand

### LINKED ISSUE #1174: Popup/transient windows not repositioned into view — partially off-screen or invisible
**Describe the bug**
When an app opens a popup/transient window (dialog, modal, notification popup), PaperWM never repositions it into view. Tiled windows are scrolled into view on focus via the clone container, but transients are real `MetaWindow`s that mutter positions itself — they're rejected by `add_filter`, so they're not in the clone container and the existing `ensureViewport`/`move_to` scroll machinery can't reach them. 

From the user's perspective, popups are shown but only partially, basically with no content in sight rendering them useless and requiring user to either move window into view or scroll the view over a popup window. Both are cumbersome and distracting.

Two visible flavours:
- **Popup gets focus** (e.g. triggered from the focused window): left clamped at the monitor edge, often partially off-screen.
- **Popup only demands attention** (focus denied by mutter's focus-stealing prevention — the common case for background-spawned popups): not visible at all. PaperWM connects no `window-demands-attention` handler, and gnome-shell's default handler bails on `skip_taskbar` transients (which most are).

**To Reproduce**
1. Run `gjs -m tests/trigger-popup.js [delay_secs]` (repro script in PR #1173), or open a background app whose window you then scroll out of view.
2. A parent window opens and gets tiled.
3. Scroll the parent OUT OF VIEW in PaperWM.
4. After the delay a transient modal dialog pops up.
5. See it land partially off-screen (Case A), or — if you didn't touch the parent after it opened (so focus is denied) — not visible at all (Case B).

**Expected behavior**
The popup is repositioned fully inside the monitor's work area so it's visible, whether or not it received focus. Demands-attention popups should become visible without stealing focus from the window the user is currently in.

**Screenshots**
N/A — the symptom is a window that's partially clipped to the monitor edge (or absent entirely in the demands-attention case).

**System information:**
```
Distribution: openSUSE Tumbleweed
GNOME Shell: 50.3
Display server: Wayland
PaperWM version: 50.0.1
Enabled extensions:
- paperwm@paperwm.github.com
```

**Additional context**
Root cause: `focus_handler` early-returns for transients without repositioning them, and PaperWM connects neither `window-demands-attention` nor `window-marked-urgent` on `global.display`. The popup is a real `MetaWindow` (not a clone in the scrollable container), so the existing scroll machinery physically cannot move it.

Fix: PR #1173 — adds `move_frame`–based repositioning for popup-class windows on both paths (focus, and a new demands-attention handler that makes the popup visible without stealing focus), with the pure positioning logic factored into a shell-free `popuputil.js` for GJS unit tests.

develop7:
Closes #1174

UPD: I've tested the fix and it works for me just fine, am using it daily.

**Transient/popup windows are now repositioned fully on-screen when they appear**, instead of being left wherever mutter clamped them — usually half-off the monitor edge, partially visible.

Popups (transients and dialogs) are real `MetaWindow`s that mutter positions itself. PaperWM never scrolled them — they're rejected by `add_filter` so they're not in the clone container, meaning `ensureViewport`/`move_to` (which move `cloneContainer`) physically can't reach them. Two entry points now reposition them with `move_frame`:

| Trigger | Before | After |
| --- | --- | --- |
| Popup gains focus | `focus_handler` early-returned; popup stayed clamped | repositioned fully into the workArea |
| Popup demands attention but focus is denied (mutter focus-stealing prevention) | nothing — PaperWM connected no `window-demands-attention` handler | repositioned **without stealing focus** (the user's current window keeps focus, by design) |

The demands-attention path is the common one for background-spawned popups (most are `skip_taskbar`, so gnome-shell's default `WindowAttentionHandler` bails on them too — they previously disappeared entirely).

### How it's structured

The pure pieces live in a new shell-free `popuputil.js` so they're unit-testable with standalone GJS (`tiling.js` imports `gi://Meta` + the live shell, so it can't be loaded in plain `gjs`):

- `workAreaToBounds` — reconciles `Space.workArea()` (monitor-relative) to the screen-absolute coords `get_frame_rect`/`move_frame` use. *Folding the monitor origin in here is the load-bearing detail: without it the popup jumps to the wrong monitor on multi-head.*
- `computeClampedPosition` — clamps a frame into bounds; an oversize window pins to the leading edge rather than going negative.

`tiling.js` holds the thin shell-facing adapters: `isPopupClass` (derived from the canonical `add_filter`, minus sticky/scratch), `ensureVisibleInWorkArea` (clamp + `move_frame`), and `positionPopupOnDemand` (the demands-attention handler, no focus steal).

### Caveats

- The focus path handles actual transients only (`isTransient`); the broader `isPopupClass` (transients + non-NORMAL dialogs) is used only on the new demands-attention path. Widening the focus branch too regressed side-effects for standalone non-NORMAL windows (topbar/overlay/selection), caught in review.
- `move_frame` on Wayland may be re-clamped by mutter's transient constraints on a later pass — the pure positioning logic is unit-tested, but the live `move_frame` behaviour needs manual verification in a gnome-shell session (can't be automated without a running compositor).

`tests/test-popup-visibility.js` (17 checks, `gjs -m`) covers the coordinate reconciliation, the clamp arithmetic including the oversize edge case, and a multi-monitor end-to-end.

---

_Generated by [`/do`](https://github.com/srid/agency) on opencode (model `zai-coding-plan/glm-5.2`)._

### DIFF
--- a/imports.js
+++ b/imports.js
@@ -17,3 +17,4 @@ export * as Topbar from './topbar.js';
 export * as Utils from './utils.js';
 export * as Workspace from './workspace.js';
 export * as OverviewLayout from './overviewlayout.js';
+export * as PopupUtil from './popuputil.js';
--- a/popuputil.js
+++ b/popuputil.js
@@ -48,19 +48,3 @@ export function computeClampedPosition(frame, bounds) {
     const y = Math.max(minY, Math.min(frame.y, maxY));
     return { x, y };
 }
-
-/**
- * Whether a window is "popup-class" for visibility purposes: a real, non-tiled
- * surface that mutter positions itself (transients plus non-NORMAL dialog /
- * modal / utility windows), excluding sticky and scratch windows which have
- * their own positioning. Takes plain booleans so it is testable without
- * MetaWindow / Scratch.
- *
- * @param {{isTransient:boolean,isNormalType:boolean,onAllWorkspaces:boolean,isScratch:boolean}} w
- * @returns {boolean}
- */
-export function classifyPopup({ isTransient, isNormalType, onAllWorkspaces, isScratch }) {
-    if (onAllWorkspaces || isScratch)
-        return false;
-    return isTransient || !isNormalType;
-}
--- a/popuputil.js
+++ b/popuputil.js
@@ -1,0 +1,66 @@
+/**
+ * Pure (shell-free) helpers for positioning popup/transient windows.
+ *
+ * Deliberately free of `gi://` and gnome-shell imports so it can be unit-tested
+ * with standalone GJS (tests/test-popup-visibility.js). `tiling.js`'s
+ * `ensureVisibleInWorkArea` is a thin adapter that feeds live MetaWindow /
+ * Space objects into these functions.
+ */
+
+/**
+ * Reconcile a Space's monitor-relative workArea into screen-absolute bounds.
+ *
+ * `Space.workArea()` returns monitor-relative coordinates (it subtracts
+ * `monitor.x` / `monitor.y`), while `MetaWindow.get_frame_rect()` and
+ * `move_frame()` are screen-absolute. Folding the monitor origin back in here
+ * avoids relocating windows to the wrong monitor on multi-head.
+ *
+ * @param {{x:number,y:number}} monitor  the monitor's screen-absolute origin
+ * @param {{x:number,y:number,width:number,height:number}} workArea  monitor-relative
+ * @returns {{x:number,y:number,width:number,height:number}}
+ */
+export function workAreaToBounds(monitor, workArea) {
+    return {
+        x: monitor.x + workArea.x,
+        y: monitor.y + workArea.y,
+        width: workArea.width,
+        height: workArea.height,
+    };
+}
+
+/**
+ * Clamp a frame rectangle into bounds, returning the target {x, y}.
+ *
+ * A window that fits is shifted minimally so it is fully inside bounds. One
+ * that is larger than bounds on an axis is pinned to the leading edge
+ * (minX / minY) rather than allowed to go negative.
+ *
+ * @param {{x:number,y:number,width:number,height:number}} frame
+ * @param {{x:number,y:number,width:number,height:number}} bounds
+ * @returns {{x:number,y:number}}
+ */
+export function computeClampedPosition(frame, bounds) {
+    const minX = bounds.x;
+    const maxX = bounds.x + bounds.width - frame.width;
+    const minY = bounds.y;
+    const maxY = bounds.y + bounds.height - frame.height;
+    const x = Math.max(minX, Math.min(frame.x, maxX));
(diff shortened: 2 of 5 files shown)


develop7:
## [Hickey/Lowy](https://kolu.dev/blog/hickey-lowy/) Analysis

Post-implement structural review (two parallel sub-agents). Both lenses converged on the same three headline findings.

| # | Lens | Finding | Disposition |
|---|------|---------|------------|
| 1 | Hickey+Lowy | `isPopupClass` re-derived the four tiling-eligibility booleans (`transient`, `window_type`, `on_all_workspaces`, `scratch`) that `add_filter` already inspects | **Fixed in this PR** (88ee77a) — delegates to `add_filter` |
| 2 | Hickey+Lowy | `positionTransientOnDemand` was a no-`this` instance method, and its name said "Transient" while the body gates on the wider `isPopupClass` | **Fixed in this PR** (ff6b555) — hoisted to module-level `positionPopupOnDemand` |
| 3 | Hickey+Lowy | `workAreaToBounds` left the canonical inline reconciliation (`centerWindow`, `cycleWindowWidthDirection`) unconverted | **Fixed in this PR** for `cycleWindowWidthDirection` (168f675); ⚠️ **No-op** for `centerWindow` — it straddles two coordinate spaces (monitor-relative for `move_to`, screen-absolute for `easeScratch`), so forcing the single-space helper regresses clarity |
| 4 | Hickey | duplicate non-NORMAL-type test cases (`MODAL_DIALOG` and `UTILITY` collapse to the same `isNormalType:false`) overclaimed distinct coverage | **Fixed in this PR** (b0bdfb0) |
| 5 | Hickey | `ensureVisibleInWorkArea` docstring asserted a monitor-agreement precondition without enforcing it | **Fixed in this PR** (88ee77a doc note, then enforced as a `get_monitor()` guard in 4d98346) |
| 6 | Hickey | "route the focus branch through `positionPopupOnDemand` to unify the gate" | ⚠️ **No-op** — would add a redundant `isPopupClass` check; the focus branch already does a single discriminating check + direct `ensureVisibleInWorkArea` |

### Notes
- The pure/impure split (`popuputil.js` shell-free ↔ `tiling.js` adapter) is a **clean seam** (Lowy) — it tracks a real axis (GJS testability), interface is a stable rect-in/rect-out contract. Not a hidden re-completing concern.
- Workspace relocation of cross-workspace popups is correctly left to the existing `insertWindow` redirect — **not re-complected** here.
- The focus-vs-no-focus policy (Case A keeps focus / clears selection; Case B doesn't steal focus) is intentionally expressed as caller choice.

> Subsequent `/code-police` pass refined one item further: it reverted the focus branch's `isTransient`→`isPopupClass` widening (cfd21ae) because non-transient non-NORMAL windows lost tiled-focus side effects (topbar/overlay/selection) — a regression hickey-lowy's predicate-consistency lens didn't catch. The broader `isPopupClass` now lives only on the new demands-attention path.

Issue #951, opened 2024-09-24
Full screen windows end up on Scratch layer after turning the screens off and on again


KETHERCORTEX:
**Describe the bug**
Full screen applications end up on a scratch layer after the screens get turned off either by GNOME's saving (Settings > Power Saving > Screen Blank) or manually executing `xdg-screensaver activate`. 

**To Reproduce**
1. Open a full screen window that is not on scratch layer
2. Either wait for GNOME's Screen Blank timeout or launch `xdg-screensaver activate`
3. Wait until the screens will turn off
4. Move mouse or press keys to wake the GNOME back up
5. See that full screen windows have Scratch option turned on 

**Expected behavior**
Windows should keep their Scratch option/state unchanged after turning the screen off and on again

**System information:**
```
Distribution: Ubuntu 24.04.1 LTS (Noble Numbat)
GNOME Shell: 46.0
Display server: Wayland
PaperWM version: 47.0.0
Enabled extensions:
- paperwm@paperwm.github.com
- dash-to-dock@micxgx.gmail.com
- NotificationCounter@coolllsk
- switcher@landau.fi
- windowIsReady_Remover@nunofarruca@gmail.com
- ddterm@amezin.github.com
- undecorate@sun.wxg@gmail.com
- window-title-is-back@fthx
- ubuntu-appindicators@ubuntu.com
```

AquariusDue:
It's happening to me too on Ubuntu 24.04 LTS with similar extensions.

Lythenas:
I can also confirm this. I tried to debug this a bit, but unfortunately I didn't really find anything useful. I also can't see any logs that are related to the problem. :(

lost-melody:
So the minimum reproduce is:

1. make a window full-screen (or mark it "Always on Top")
2. disable paperwm while the full-screen window is focused: `gnome disable paperwm@paperwm.github.com`
3. enable paperwm: `gnome enable paperwm@paperwm.github.com`

The reason is that paperwm doesn't really remember windows' scratch state across extension lifetimes, instead it saves the state to `meta_window[float]`, where `float` is initialized on enabled and zeroed on disabled:

- [scratch.js:L159](https://github.com/paperwm/PaperWM/blob/07214b0ed9983806aa827f6b371431f4a5e58bd8/scratch.js#L159)

moves always-on-top windows into scratch layer upon paperwm enabled:

[...]

PR #1187, opened 2026-09-12
Preserve scratch and tiling state across extension reloads

ricardosalta:
PaperWM currently rebuilds scratch membership from `above`/`minimized` when it is re-enabled. That can move a previously tiled window into scratch after locking and unlocking, and a minimized tile can lose the marker that would return it to tiling when restored.

Preserve per-window scratch/tiling intent in `SaveState` before teardown and consume it when rebuilding each workspace. Keep pending minimize-restoration intent until its idle callback completes, and invalidate stale callbacks on cleanup, re-minimize, or an explicit scratch choice. Deliberately scratched windows retain that choice, including when minimized. Session modes and screen-lock behavior are unchanged.

Fixes #951. This follows [lost-melody's lifecycle diagnosis](https://github.com/paperwm/PaperWM/issues/951#issuecomment-2581757579) and [wisdom-in-snow's suggestion to use SaveState](https://github.com/paperwm/PaperWM/issues/951#issuecomment-2789499226).

Validation:

- `tests/run-scratch-tests`: 18 passing tests; 9 fail against unmodified upstream. The suite loads the complete production modules with explicit GNOME mocks.
- Native GNOME 46 automation: an actual fullscreen tile survives disable/enable and exit from fullscreen; immediate restoration before disable succeeds; three additional cycles preserve visible/minimized tiles and deliberate scratch windows.
- The same runtime change backported to GNOME 49 has positive desktop-user feedback after a fresh login.
- New tests pass the repository's ESLint configuration; the runtime files introduce no diagnostics beyond the seven already present in upstream.

Native automation exercises the extension lifecycle in an isolated session. It does not simulate physical display power-off or Telegram itself. Test setup and coverage are documented in `tests/README.md`.

### DIFF
--- a/scratch.js
+++ b/scratch.js
@@ -59,7 +59,13 @@ export function easeScratch(metaWindow, targetX, targetY, params = {}) {
     });
 }
 
-export function makeScratch(metaWindow) {
+export function makeScratch(metaWindow, { preserveTiling = false } = {}) {
+    // Internal refreshes keep temporary scratch intent; an explicit choice
+    // supersedes any deferred return to tiling.
+    if (!preserveTiling) {
+        delete metaWindow._tiled_on_minimize;
+        delete metaWindow._pendingScratchRestore;
+    }
     let fromNonScratch = !metaWindow[float];
     let fromTiling = false;
     // Relevant when called while navigating. Use the position the user actually sees.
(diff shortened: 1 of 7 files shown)


Jalst (approved):
Took a deep dive through the changes and verified the fix locally:

- **State preservation & lifecycle**: Storing the snapshot in `SaveState.prevWindowStates` using a `WeakMap` ensures window state survives extension teardown/re-enabling without risking memory leaks for destroyed windows.
- **Intent separation**: Adding the `{ preserveTiling }` flag to `makeScratch` cleanly distinguishes intentional user floating/scratch decisions from transient compositor minimize/fullscreen states.
- **Race condition avoidance**: Using `_pendingScratchRestore` to guard and invalidate asynchronous compositor idle callbacks prevents stale callbacks from restoring windows that were subsequently re-minimized or made scratch.
- **Test coverage**: Pulled branch locally and ran `./tests/run-scratch-tests`—all 18 unit tests passed cleanly with zero regressions.

Clean, robust fix for #951. LGTM! Recommending merge @Thesola10.

PR #1190, opened 2026-09-20
Updated nix dev shell, add envrc

GooseRooster:
Updates the nix flake and nix shell, adding reproducible dependencies for users of Nix.

Included:

.envrc - direnv integration. With direnv installed,  `direnv allow` the flake is automatically activated on cd'ing into the directory.
vtsls - Typescript/Javascript language server. Allows various editors to get LSP support within the flake for the project
eslint - project linter
glib,zip - make dependencies

The flake also contains a gitignored shell hook `.dev.local.sh` allowing developers to amend the flake with chosen environment variables etc.

The `shell.nix` path has also been updated.


flake.nix
```diff
@@ -10,6 +10,30 @@
     in
     { packages.default = hostPkgs.callPackage ./default.nix {};
 
+      devShells.default = hostPkgs.mkShell {
```
Thesola10:
```suggestion
      devShells.default = hostPkgs.callPackage ./shell.nix {};
```

Do not duplicate code if possible

shell.nix
```diff
@@ -1,6 +1,21 @@
...
+pkgs.mkShell {
+  packages = with pkgs; [
+    vtsls
+    eslint
+    nodejs
+    glib
+    zip
+  ];
 
+  shellHook = ''
```
Thesola10: Refrain from adding personal hooks you're not going to commit. Consider adding shell hooks to the .envrc wherever possible instead

.gitignore
```diff
@@ -11,3 +11,7 @@ package-lock.json
 
 # generated disk image for test VM
 nixos.qcow2
+
+# per-developer devshell hook + direnv cache
+/.dev.local.sh
```
Thesola10: I don't believe this is a standard path. You can add non-committed gitignores in .git/ignore if I recall correctly

Thesola10 (changes requested):
I like the idea. Will test more on my end to see if it holds up

Thesola10:
On second read .dev.local.sh isn't a bad idea, but again it should extend .envrc, or .envrc should remain uncommitted to let everyone customize theirs to taste.

GooseRooster:
Thanks for the review! I agree and will make requested changes. I'll place .envrc into gitignore and it can be the source of dev specific customizations. 

Will also keep the pkg source in shell.nix as requested.

GooseRooster:
Done - .envrc is now gitignored so personal hooks can live there. I actually am a bit surprised I didn't think of this before - its such a simple file usually that it makes sense to keep them there. 

The callPackage tweak is also really neat. Added!

